PR #8, opened 2016-02-23
Fixes for update_external_sources.sh

Dr-Shadow:


karl-lunarg:
Dr-Shadow:  We're in the middle of some repo reorganization and the master-new branch is work in progress.

Eventually we'll want the version of update_external_sources.sh that is in the master branch.  I think that the version in master-new is the result of some merges that need to be reconciled and the master-new version will be replaced with the version currently in master.

Assigning to Jon to resolve. (He's in the middle of getting master-new ready now)

Dr-Shadow:
Ok that's very nice and I didn't see the fact I can't merge this update_external_sources.sh on master. This pull request should be closed once you merge master_new on master then.
Btw I have a question : I could compile the master_new branch with some modifications (dependency for ncurses was not resolved) and I saw that the intel driver is really different from the one provided by mesa on vulkan branch.
Can you explain me the differences ? (the driver from mesa is not working on my platform probably because I have Haswell).
Thank you in advance

karl-lunarg:
We may end up using your patches for the script; Jon was in the middle of working on it.  We didn't expect people to be looking at master-new.

The intel driver in VulkanTools is an old LunarG driver that we're trying to deprecate.  We've provided the source more for peoples' interest and for those who have narrow enough requirements that they still find it useful.  For example, we still use it internally for testing.  We do the minimum to keep it alive for these purposes.

This LunarG driver is completely different from the Intel "OTC" driver that is in Mesa.  The LunarG driver does not support Skylake and is incomplete in many areas.

PR #11, opened 2016-02-29
icd: geometry shaders are broken (gen7)

warcog:
If you attempt to draw more than one primitive with one vkCmdDraw() call, the geometry shader output for primitives after the first is broken.

None of the geometry shader tests draw more than one triangle, but you can observe this behavior by modifying the GeometryShaderHelloWorld test to draw 2 triangles instead of one.

I've made a patch which "fixes" (suboptimal, but it should help someone make a proper fix) the issue.


karl-lunarg:
I picked up part of this - the missing break and the protection for dereferencing buf.

Unfortunately, I didn't see the correspondence between drawing two triangles and the proposed fix.  That is, your fix caused a change in the out_count independent of the number of tris.

Perhaps I missed something.  If you'd like to submit another pull request, I'd be happy to try again.

PR #12, opened 2016-03-08
Fix CMake Xlib DisplayServer name.

DavidYen:
The DisplayServer name for Xlib was incorrectly set to XLib which did not match other references
in the build scripts. This has been fixed which fixes the build for XLib support.


jon-lunarg:
David,

Thanks for the bug fix. This file actually is a partial downstream of a different Github project Vulkan-LoaderAndValidationLayers.  This fix  should be applied there first and then the VulkanTools maintainer will merge  in the patch from LoaderAndValidationLayers into trunk and master branch in this repo.  See here: https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers for where to file your pull request.

karl-lunarg:
I've applied the fix to LoaderAndValidationLayers and will update VulkanTools later today.

PR #14, opened 2016-03-11
nulldrv: Claim support for all features

chrisforbes:
Previously we supported no features. SC will shortly start checking for
capabilities matching enabled device features, so we need to expose
at least the features the tests and demos use.

It's safe to just expose everything, so do.


tobine:
+1

chrisforbes:
Alright, my CLA query is resolved; this should be good to go. ilo wants a similar change (but more selective about the features it enables) but I'm not running on a machine that can run ilo currently.

jon-lunarg:
cherry-picked to master

PR #18, opened 2016-04-01
Revive vktraceviewer

plohrmann:
Revive the GUI for generating, viewing, and replaying vktrace files.
* Adds a dependency on Qt, but if Qt is not found, then a warning is printed and vktraceviewer and related projects are not added to the final solution.


davidlunarg:
Files from this pull request have been merged to master.

PR #27, opened 2016-04-14
vktrace: Correct values to VK_INSTANCE/DEVICE_LAYERS

sashinde:
VK_INSTANCE_LAYERS and VK_DEVICE_LAYERS should be assigned
VK_LAYER_LUNARG_vktrace instead of just Vktrace


karl-lunarg:
Change applied to master.  Thanks!

PR #32, opened 2016-04-26
fix host buffer in vkCmdUpdateBuffer is not correct dumped

ttyio:
host buffer should has size "dataSize" instead of sizeof(uint32_t)


davidlunarg:
Merged to main

PR #38, opened 2016-05-06
Fix Building vktrace on ARM Linux

sashinde:
Below are two fixes:
- Remove hardcoding "corei7" in cmake files
- Fix vktrace_get_time()

### DIFF
--- a/vktrace/src/build_options.cmake
+++ b/vktrace/src/build_options.cmake
@@ -177,7 +177,6 @@ endfunction()
 ### TODO: see if sse is generated with these instructions and clang:
 ## -march=corei7 -msse -mfpmath=sse
 
-set(MARCH_STR "-march=corei7")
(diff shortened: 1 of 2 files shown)


karl-lunarg:
The original pull request introduced a compilation warning about exiting the function without a return statement.  Also, the time init function was not getting called and all this was just plain broken on Linux.  So I reworked much of this get_time code (https://github.com/LunarG/VulkanTools/commit/f7b3819ca005caf4cb88df63cac27ad1e6dda405)

Please confirm that it builds on ARM.

sashinde:
Yes vktrace/vkreplay are building. I used below cmake config to build:

cmake -DBUILD_LAYERS=OFF -DBUILD_VKJSON=OFF -DBUILD_ICD=OFF -DBUILD_TESTS=OFF -DBUILD_DEMOS=OFF

PR #46, opened 2016-05-27
Update VK_LAYER_LUNARG_{basic,multi*,screenshot}

olvaffe:
Update them such that

 - there is a clear separation between layer commands and layer library interface functions
 - GIPA handles device commands as well
 - avoid invalid GIPA calls
 - add/fix/improve introspection functions

### DIFF
--- a/layersvt/basic.cpp
+++ b/layersvt/basic.cpp
@@ -122,16 +122,12 @@ basic_GetPhysicalDeviceFormatProperties(VkPhysicalDevice gpu, VkFormat format, V
     printf("Completed wrapped vkGetPhysicalDeviceFormatProperties() call w/ gpu: %p\n", (void *)gpu);
 }
 
-static const VkLayerProperties globalLayerProps[] = {{
+static const VkLayerProperties basic_LayerProps = {
     "VK_LAYER_LUNARG_basic",
     VK_MAKE_VERSION(1, 0, VK_HEADER_VERSION), // specVersion
     1,              // implementationVersion
     "LunarG Sample Layer",
-}};
-
-static const VkLayerProperties basic_physicaldevice_layers[] = {{
-    "VK_LAYER_LUNARG_basic", VK_MAKE_VERSION(1, 0, VK_HEADER_VERSION), 1, "Sample layer: basic, implements vkLayerBasicEXT",
-}};
+};
 
 static const VkExtensionProperties basic_physicaldevice_extensions[] = {{
     "vkLayerBasicEXT", 1,
@@ -139,12 +135,12 @@ static const VkExtensionProperties basic_physicaldevice_extensions[] = {{
 
 VKAPI_ATTR VkResult VKAPI_CALL
 basic_EnumerateInstanceLayerProperties(uint32_t *pCount, VkLayerProperties *pProperties) {
-    return util_GetLayerProperties(ARRAY_SIZE(globalLayerProps), globalLayerProps, pCount, pProperties);
+    return util_GetLayerProperties(1, &basic_LayerProps, pCount, pProperties);
 }
 
 VKAPI_ATTR VkResult VKAPI_CALL
 basic_EnumerateDeviceLayerProperties(VkPhysicalDevice physicalDevice, uint32_t *pCount, VkLayerProperties *pProperties) {
-    return util_GetLayerProperties(ARRAY_SIZE(basic_physicaldevice_layers), basic_physicaldevice_layers, pCount, pProperties);
+    return util_GetLayerProperties(1, &basic_LayerProps, pCount, pProperties);
 }
 
 VKAPI_ATTR VkResult VKAPI_CALL
--- a/layersvt/basic.cpp
+++ b/layersvt/basic.cpp
@@ -201,6 +201,10 @@ VKAPI_ATTR PFN_vkVoidFunction VKAPI_CALL basic_GetInstanceProcAddr(VkInstance in
     if (!strcmp("vkEnumeratePhysicalDevices", pName))
         return (PFN_vkVoidFunction)basic_EnumeratePhysicalDevices;
 
+    PFN_vkVoidFunction proc = basic_GetDeviceProcAddr(VK_NULL_HANDLE, pName);
+    if (proc)
+        return proc;
+
     if (instance == NULL)
         return NULL;
 
--- a/layersvt/basic.cpp
+++ b/layersvt/basic.cpp
@@ -221,12 +221,11 @@ VKAPI_ATTR PFN_vkVoidFunction VKAPI_CALL basic_GetInstanceProcAddr(VkInstance in
     if (!strcmp("vkEnumeratePhysicalDevices", pName))
         return (PFN_vkVoidFunction)basic_EnumeratePhysicalDevices;
 
+    assert(instance);
+
     PFN_vkVoidFunction proc = basic_GetDeviceProcAddr(VK_NULL_HANDLE, pName);
     if (proc)
         return proc;
-
-    if (instance == NULL)
-        return NULL;
 
     if (instance_dispatch_table(instance)->GetInstanceProcAddr == NULL)
         return NULL;
--- a/layersvt/basic.cpp
+++ b/layersvt/basic.cpp
@@ -26,7 +26,6 @@
 #include "vk_dispatch_table_helper.h"
 #include "vulkan/vk_layer.h"
 #include "vk_layer_table.h"
-#include "vk_layer_extension_utils.h"
 
 static std::unordered_map<dispatch_key, VkInstance> basic_instance_map;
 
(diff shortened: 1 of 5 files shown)


olvaffe:
I am assigning this to myself and will commit in a day or two.

karl-lunarg:
Olv, thanks for cleaning this up.  The screenshot changes seem OK to me.  I have not worked much with the other layers.  Hopefully Jon can look at it from the loader/layer interface point of view.  But +1 from me.

layersvt/screenshot.cpp
```diff
@@ -762,6 +766,9 @@ intercept_core_instance_command(const char *name) {
         { "vkCreateInstance", reinterpret_cast<PFN_vkVoidFunction>(vkCreateInstance) },
         { "vkCreateDevice", reinterpret_cast<PFN_vkVoidFunction>(vkCreateDevice) },
         { "vkEnumeratePhysicalDevices", reinterpret_cast<PFN_vkVoidFunction>(vkEnumeratePhysicalDevices) },
+        { "vkEnumerateInstanceLayerProperties", reinterpret_cast<PFN_vkVoidFunction>(vkEnumerateInstanceLayerProperties) },
+        { "vkEnumerateDeviceLayerProperties", reinterpret_cast<PFN_vkVoidFunction>(vkEnumerateDeviceLayerProperties) },
+        { "vkEnumerateInstanceExtensionProperties", reinterpret_cast<PFN_vkVoidFunction>(vkEnumerateInstanceExtensionProperties) },
         { "vkEnumerateDeviceExtensionProperties", reinterpret_cast<PFN_vkVoidFunction>(vkEnumerateDeviceExtensionProperties) }
```
jon-lunarg: Rather than adding these three lines in the intercept_core_command(), it would be better to add them into the introspection version of vkGetInstanceProcAddr.  Adding them here implies that the layer is intercepting these 3 commands but it is not.

layersvt/screenshot.cpp
```diff
@@ -739,6 +739,10 @@ VK_LAYER_EXPORT VKAPI_ATTR PFN_vkVoidFunction VKAPI_CALL vkGetDeviceProcAddr(VkD
 
 VK_LAYER_EXPORT VKAPI_ATTR PFN_vkVoidFunction VKAPI_CALL vkGetInstanceProcAddr(VkInstance instance, const char *funcName) {
     PFN_vkVoidFunction proc = intercept_core_instance_command(funcName);
```
jon-lunarg: The null instance check from lines 749 and 750 could be moved to after the intercept_core_instance_command().

layersvt/multi.cpp
```diff
@@ -167,6 +167,10 @@ VKAPI_ATTR PFN_vkVoidFunction VKAPI_CALL multi1GetInstanceProcAddr(VkInstance in
     if (!strcmp("vkDestroyInstance", pName))
         return (PFN_vkVoidFunction)multi1DestroyInstance;
 
+    PFN_vkVoidFunction proc = multi1GetDeviceProcAddr(VK_NULL_HANDLE, pName);
+    if (proc)
+        return proc;
+
     if (instance == NULL)
```
jon-lunarg: The instance NULL check could be before line 170.

layersvt/multi.cpp
```diff
@@ -26,11 +26,41 @@
 #include "vk_loader_platform.h"
 #include "vulkan/vk_layer.h"
 #include "vk_layer_table.h"
+#include "vk_layer_extension_utils.h"
```
jon-lunarg: Please copy the functions util_GetExtensionProperties  and util_GetLayerProperties directly in this file rather than making this layer dependent upon the layer utils library.  This is an example layer intended to be copied by layer authors to their own repository. It is best if it is standalone  and doesn't pull in a large set of utilities most of which are not necessarily needed by the layer author.

layersvt/multi.cpp
```diff
@@ -154,6 +221,14 @@ VKAPI_ATTR PFN_vkVoidFunction VKAPI_CALL multi1GetDeviceProcAddr(VkDevice device
...
 VKAPI_ATTR PFN_vkVoidFunction VKAPI_CALL multi1GetInstanceProcAddr(VkInstance instance, const char *pName) {
+    if (!strcmp("vkEnumerateInstanceLayerProperties", pName))
+        return (PFN_vkVoidFunction)multi1EnumerateInstanceLayerProperties;
+    if (!strcmp("vkEnumerateDeviceLayerProperties", pName))
+        return (PFN_vkVoidFunction)multi1EnumerateDeviceLayerProperties;
+    if (!strcmp("vkEnumerateInstanceExtensionProperties", pName))
+        return (PFN_vkVoidFunction)multi1EnumerateInstanceExtensionProperties;
+    if (!strcmp("vkEnumerateDeviceExtensionProperties", pName))
+        return (PFN_vkVoidFunction)multi1EnumerateDeviceExtensionProperties;
     if (!strcmp(pName, "multi1GetInstanceProcAddr") || !strcmp(pName, "vkGetInsatnceProcAddr"))
```
jon-lunarg: I don't think the above (except the EnumerateDeviceExtensionProperties() should be added here since this is the layer intercept function.  You already have these in the VK_LAYER_LUNARG_multi1GetInstanceProcAddr. Why are they needed here?

layersvt/multi.cpp
```diff
@@ -247,7 +322,52 @@ VKAPI_ATTR void VKAPI_CALL multi2DestroyInstance(VkInstance instance, const VkAl
...
 VKAPI_ATTR PFN_vkVoidFunction VKAPI_CALL multi2GetInstanceProcAddr(VkInstance inst, const char *pName) {
+    if (!strcmp("vkEnumerateInstanceLayerProperties", pName))
+        return (PFN_vkVoidFunction)multi2EnumerateInstanceLayerProperties;
+    if (!strcmp("vkEnumerateDeviceLayerProperties", pName))
+        return (PFN_vkVoidFunction)multi2EnumerateDeviceLayerProperties;
+    if (!strcmp("vkEnumerateInstanceExtensionProperties", pName))
+        return (PFN_vkVoidFunction)multi2EnumerateInstanceExtensionProperties;
+    if (!strcmp("vkEnumerateDeviceExtensionProperties", pName))
+        return (PFN_vkVoidFunction)multi2EnumerateDeviceExtensionProperties;
     if (!strcmp("vkCreateInstance", pName))
```
jon-lunarg: Ditto previous comment.

layersvt/basic.cpp
```diff
@@ -30,7 +30,7 @@
 typedef VkResult(VKAPI_PTR *PFN_vkLayerBasicEXT)(VkDevice device);
 static PFN_vkLayerBasicEXT pfn_layer_extension;
 
-VK_LAYER_EXPORT VKAPI_ATTR VkResult VKAPI_CALL vkLayerBasicEXT(VkDevice device) {
+VKAPI_ATTR VkResult VKAPI_CALL basic_LayerBasicEXT(VkDevice device) {
     printf("In vkLayerBasicEXT() call w/ device: %p\n", (void *)device);
```
jon-lunarg: I don't care for this particular rename since this would be the extension command name which would probably be of the form vk<some appropriate command name> rather than basic_.

layersvt/basic.cpp
```diff
@@ -201,6 +201,10 @@ VKAPI_ATTR PFN_vkVoidFunction VKAPI_CALL basic_GetInstanceProcAddr(VkInstance in
     if (!strcmp("vkEnumeratePhysicalDevices", pName))
         return (PFN_vkVoidFunction)basic_EnumeratePhysicalDevices;
 
+    PFN_vkVoidFunction proc = basic_GetDeviceProcAddr(VK_NULL_HANDLE, pName);
+    if (proc)
+        return proc;
+
     if (instance == NULL)
```
jon-lunarg: The NULL instance check could be moved before the basic_GetDeviceProcAddr

layersvt/screenshot.cpp
olvaffe: The intention does is for the layer to intercept these three commands.  It allows all introspection functions to be wrappers to the layer's commands.

layersvt/screenshot.cpp
olvaffe: I will move them and change them to `assert(instance)`.  After `intercept_core_instance_command`, `instance` is always valid.

layersvt/multi.cpp
olvaffe: Will move and change to assert.

layersvt/multi.cpp
olvaffe: Will do.

layersvt/multi.cpp
olvaffe: I want the layer to intercept all four of them, mainly for completeness.  Also, if the loader-layer interface was to be updated to export layers' GIPA, the completeness becomes a requirement.  That is, the spec does not allow us to return NULL when GIPA("vkEnumerateInstanceLayerProperties") is called.

layersvt/basic.cpp
olvaffe: The rename is for consistency: all layer commands are prefixed by basic_ instead of vk.  I do not see how it is a extension command matters?

layersvt/basic.cpp
olvaffe: Done.

jon-lunarg:
Olv, thanks for the changes!  Cherry picked the patches to master branch.

Issue #43, opened 2016-05-14
api_dump does not dump vkCreateDevice parameters/results when vkCreateDevice returns an error


jcoder58:
I have the api_dump layer enabled in both the instance and the VkDeviceCreateInfo enabled layers, however the call is not reported.  Other calls make prior to the vkCreateDevice  call are dumped correctly.   When I fix the issue with vkCreateDevice so that it returns success, it is dumped correctly.

karl-lunarg:
Are saying that api_dump quits working if vkCreateDevice fails?

jcoder58:
Yes.  It quits working and does not dump the vkCreateDevice call at all.

karl-lunarg:
OK, thanks.   (Sorry, sometimes I don't combine the bug's Subject line with the description)  We can see the value in having this work even if a vulkan call fails, to help debugging.  We'll have to look at the implementation to see if there's any reason why api_dump cannot continue on in this case.

MarkMLunarG:
According to the code that I am looking at, vkCreateDevice always dumps information, whether the command fails or not. On the other hand vkCreateInstance says nothing upon failure. Are you sure you mean vkCreateDevice and not vkCreateInstance? I do have a fix available for the later, but don't see a problem with the former

MarkMLunarG:
Please see PR #47

jcoder58:
Just verified that it is VkCreateDevice.  You can duplicate it by passing a bad extension name to vkCreateDevice (I used "yyy").

olvaffe:
Bad extension names are likely rejected by the loader, before any validation layer kicks in.

jon-lunarg:
Yes I would not expect api_dump layer to ever get called  for CreateDevice with a bad extension name.

karl-lunarg:
jcoder58:  When vkCreateDevice is called with an invalid extension, it returns VK_ERROR_EXTENSION_NOT_PRESENT.  Jon is correct in that no further calls are made down the layer/device chain under these conditions, so a layer like api_dump would not be called.

It seems that the VK_ERROR_EXTENSION_NOT_PRESENT is a specific enough return code to direct the programmer to examine the extension list for a problem

Is this sufficient to address your issue?

I think that the most we could do otherwise is to change the loader code to log a message that could list the "yyy" extension as not being found.  Then, you could set VK_LOADER_DEBUG=all to see this diagnostic message.

[...]

PR #47, opened 2016-05-30
VulkanTools: GHIssue #43

MarkMLunarG:
Small tweak to handling failure in vkCreateInstance to allow api_dump information
to still be shown

Change-Id: Ifa77e250deec553cde176a4f67aa4caf719d4190


vk-vtlayer-generate.py
```diff
@@ -1174,16 +1174,17 @@ def generate_intercept(self, proto, qual):
...
                      '    VkResult result = fpCreateInstance(pCreateInfo, pAllocator, pInstance);\n'
-                     '    if (result != VK_SUCCESS)\n'
-                     '        return result;\n'
-                     '    initInstanceTable(*pInstance, fpGetInstanceProcAddr);\n'
-                     '    createInstanceRegisterExtensions(pCreateInfo, *pInstance);\n'
+                     '    if (result == VK_SUCCESS) {\n'
+                     '        initInstanceTable(*pInstance, fpGetInstanceProcAddr);\n'
+                     '        createInstanceRegisterExtensions(pCreateInfo, *pInstance);\n'
+                     '    }\n'
                      '    %s%s%s\n'
```
jon-lunarg: My first comment on this code is by not returning early when result != VK_SUCCESS, then api_dump will try to print out all the parameters.  If the cause of the failure was bad parameters then api_dump may cause a crash before printing anything out.  I think this is okay since the loader and ICD also may crash with bad parameters  to CreateInstance.

By skipping the dispatch table initialization you are ensuring that any subsequent Vulkan call will likely crash. 

Overall LGTM.

karl-lunarg: I think that the entire point of the change was to print out the parameters when there is a CreateInstance failure, which might allow the user to examine the parameters to determine why it failed.  

It looks like the code to print out these parameters would survive, but there's always a chance of dereferencing a non-NULL bad pointer, but at least there's a chance of getting through.

And if CreateInstance is failing, then the failure of any subsequent Vulkan calls should be expected.

courtney-g:
Note: The application will get a real error code if CreateInstance or CreateDevice is given an invalid extension name or layer name:
VK_ERROR_LAYER_NOT_PRESENT
VK_ERROR_EXTENSION_NOT_PRESENT

Though, if a layer dies before returning that ...

karl-lunarg:
Yes, I just made that point over in https://github.com/LunarG/VulkanTools/issues/43.  Note that the original issue was for CreateDevice, and I suggested alternatives there.  The loader cuts off CreateDevice if there is a bad extension, and we could say just look at the VK_ERROR_EXTENSION_NOT_PRESENT return code or maybe have the loader log the name of the failed extension.

So the question here is whether or not to put in this (rather unrelated) change for CreateInstance.

MarkMLunarG:
pushed to master :+1: 
SHA-1: d5228657e3a33667023a59fe0d4041e2c6466ffc

* layers: LX506 fix vkFlushMappedMemoryRanges validation error

Fix LX506: vkFlushMappedMemoryRanges validation error
 with VkMappedMemoryRange::size = VK_WHOLE_SIZE.
Resolution: Change test of size variable to correctly test the
 size passed by application

Change-Id: I6d855482985512b5096696f5c37346dde8a391ac

PR #65, opened 2016-07-19
Various improvements to enable trace/replay of DOOM.

plohrmann:
vktrace:
* Use parallel threads when doing large memory copies.
* Fix tracing of entrypoints which include pQueueFamilyIndicies.

vkreplay:
* Report errors on replay if an object cannot be remapped, and don't replay that call because it will likely result in a driver crash.
* Fix a few memory leaks in vkreplay.

vktraceviewer:
* Enable printing of string parameters
* vkCreate calls now show the first object handle that was created.
* Improving formatting of packet header and parameters when printed in UI.
* Update to version 54 of some of the Qt dlls, which corresponds to Qt version 5.6 instead of version 5.3 (this update may have been required to build in VS2015, or it was the latest version available when we started on the project).

### DIFF
--- a/vktrace/src/vktrace_common/optimization_function.cpp
+++ b/vktrace/src/vktrace_common/optimization_function.cpp
@@ -1,55 +1,64 @@
+/*
+* Copyright (C) 2016 Advanced Micro Device, Inc.
+*/
+
 #include <stdio.h>
 #include <string.h>
-#include <ppl.h>
 #include <stdlib.h>
 
+#ifdef WIN32
+#include <ppl.h>
 using namespace concurrency;
-using namespace std;
-
-
 #define PARALLEL_INVOKE_NUM   10
 #define SIZE_LIMIT_TO_USE_OPTIMIZATION 67100000
+#endif
+
+using namespace std;
+
 extern "C" void *opt_memcpy(void * destination, const void * source, size_t size)
 {
-    void *pRet=NULL;
-		if (size < SIZE_LIMIT_TO_USE_OPTIMIZATION)
-		{
-			//memcpy(ptr_address, pBuffer, (size_t)size);
-			pRet=memcpy(destination, source, (size_t)size);
-		}
-		else
-		{
-		    pRet=destination;
-			void* ptr_parallel[PARALLEL_INVOKE_NUM];
-			void* pBuffer_parallel[PARALLEL_INVOKE_NUM];
-            size_t size_parallel[PARALLEL_INVOKE_NUM];
+    void *pRet = NULL;
+#ifndef WIN32
+    pRet = memcpy(destination, source, (size_t)size);
+#else
+    if (size < SIZE_LIMIT_TO_USE_OPTIMIZATION)
+    {
+        pRet = memcpy(destination, source, (size_t)size);
+    }
+    else
+    {
+        pRet = destination;
+        void* ptr_parallel[PARALLEL_INVOKE_NUM];
+        void* pBuffer_parallel[PARALLEL_INVOKE_NUM];
+        size_t size_parallel[PARALLEL_INVOKE_NUM];
 
-			int stepsize = size / PARALLEL_INVOKE_NUM;
-			for (int i = 0; i < PARALLEL_INVOKE_NUM; i++)
-			{
-				ptr_parallel[i] = (char *)destination + i*stepsize;
-				pBuffer_parallel[i] = (char *)source + i*stepsize;
-				if ((i + 1) == PARALLEL_INVOKE_NUM)
-				{
-					size_parallel[i] = size - (PARALLEL_INVOKE_NUM-1)*stepsize;
-				}
-				else
-				{
-                   size_parallel[i] = stepsize;
-				}
-			}
-			parallel_invoke(
-				[&ptr_parallel, &pBuffer_parallel, &size_parallel] { memcpy(ptr_parallel[9], pBuffer_parallel[9], (size_t)size_parallel[9]); },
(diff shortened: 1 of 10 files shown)


vinjn:
How do you track the persistent mapped buffers?
DOOM use it a lot.

plohrmann:
We have another change in-flight that fixes the persistently mapped buffers issue, however the solution is Windows-only. A cross-platform solution is on the horizon though.

vktrace/src/vktrace_common/optimization_function.cpp
```diff
@@ -0,0 +1,55 @@
...
+#include <ppl.h>
+#include <stdlib.h>
+
+using namespace concurrency;
+using namespace std;
+
+
+#define PARALLEL_INVOKE_NUM   10
+#define SIZE_LIMIT_TO_USE_OPTIMIZATION 67100000
+extern "C" void *opt_memcpy(void * destination, const void * source, size_t size)
```
jon-lunarg: How was this number (67100000) determined?  Based on DOOM perf profiling?

vktrace/src/vktrace_common/optimization_function.cpp
```diff
@@ -0,0 +1,55 @@
+#include <stdio.h>
+#include <string.h>
+#include <ppl.h>
+#include <stdlib.h>
```
jon-lunarg: ppl.h is not found on Linux. Any suggestions? Probably this code needs to be rewritten using C++11 STL constructs.

vktrace/src/vktrace_extensions/vktracevulkan/vkreplay/vkreplay_vkreplay.cpp
```diff
@@ -1566,8 +1566,22 @@ VkResult vkReplay::manually_replay_vkWaitForFences(packet_vkWaitForFences* pPack
     {
         (*(pFence + i)) = m_objMapper.remap_fences((*(pPacket->pFences + i)));
     }
-    replayResult = m_vkFuncs.real_vkWaitForFences(remappedDevice, pPacket->fenceCount, pFence, pPacket->waitAll, pPacket->timeout);
-    VKTRACE_DELETE(pFence);
+	if (pPacket->result == VK_SUCCESS)
+	{
+		replayResult = m_vkFuncs.real_vkWaitForFences(remappedDevice, pPacket->fenceCount, pFence, pPacket->waitAll, UINT64_MAX);// mean as long as possible
+	}
```
jon-lunarg: What is the motivation for waiting for as long as possible here?  Shouldn't one just wait for the same timeout as was originally done during tracing?

vktrace/src/vktrace_replay/vkreplay_main.cpp
```diff
@@ -109,7 +109,11 @@ int main_loop(Sequencer &seq, vktrace_trace_packet_replay_library *replayerArray
                         if (res != VKTRACE_REPLAY_SUCCESS)
                         {
                            vktrace_LogError("Failed to replay packet_id %d.",packet->packet_id);
-                           return -1;
+						   static BOOL QuitOnAnyError=FALSE;
```
jon-lunarg: whitespace off on line 112

vktrace/src/vktrace_common/optimization_function.cpp
jon-lunarg: Shouldn't the number of threads to invoke be determined at runtime based on the given platform.

jon-lunarg:
I tested this patch series with Dota2 trace and replay and it seems to work just fine.  I don't see any noticable tracing speed improvement though from the threaded memcpy.

vktrace/src/vktrace_common/optimization_function.cpp
plohrmann: Parallel_Invoke should handle the threading side for us based on the specific system; in this case it will use up to 10 threads if they are available otherwise it will partition the work among the necessary threads.

vktrace/src/vktrace_common/optimization_function.cpp
plohrmann: Will update to make this be included only on Windows. We will work on cross-platform solution.

vktrace/src/vktrace_extensions/vktracevulkan/vkreplay/vkreplay_vkreplay.cpp
plohrmann: We had a few cases where the timeout from the application wasn't long enough during replay, and replay would stop because of different return return values (received TIMEOUT when SUCCESS was expected). I guess we could continually loop and keep the app's original timeout. At least then the additional wait is inside the vkreplay code instead of waiting inside the driver. I'll review this and see if we can simplify these conditionals.

plohrmann: After reviewing a bit more, I think the existing solution is the best way to most closely replicate the original applications behavior. If the original app only called vkWaitForFences() once and got a success, this will best ensure that we can get success by calling vkWaitForFences() only once. 

If the original app got a TIMEOUT, then we'll replay the call once with the minimum amount of time (0) in hopes of getting a TIMEOUT. Note that the only way to truly guarantee a timeout is to skip the call completely and just return a TIMEOUT, but then we're not replaying a call that the application made so it's probably not a great solution. In an actual app, we'd expect them to make this call again, so they'll likely see SUCCESS on an upcoming call.

If the call returned an error, then it really doesn't matter what we set the timeout value to, because we're just taking a chance anyway.

Based on the above logic, I think this multi-conditional approach is reasonable.

ardaLunarG:
All merged except multi-thread memcopy patch.

PR #66, opened 2016-07-19
support print out frame counter in api_dump layer

ttyio:
sample output syntax:
     t{0} f{2476} vkBeginCommandBuffer .....


karl-lunarg:
Applied with https://github.com/LunarG/VulkanTools/commit/10ecc52763c4637f6047141469eb46a822808c9c

PR #70, opened 2016-07-27
Android support for api_dump

cnorthrop:
These changes get api_dump working on Android.  It routes the dump to stdout.  I added a test that is disabled, it is a little too verbose.  I'll try to come back to that at some point so we can ensure api_dump works on all platforms.


karl-lunarg:
Committed to master.

PR #71, opened 2016-08-01
loader: support discovering manifests in XDG locations on Linux

johndrinkwater:
Some users set the environment variable XDG_DATA_HOME to relocate
~/.local/share/ to a more friendly location and so it is not wise to
look solely at that location.

This change adds basic support for that env var without changing the
behaviour for typical users. If XDG_DATA_HOME is defined, the loader
checks XDG_DATA_HOME/vulkan/..., if HOME is defined the loader will
check HOME/.local/share/vulkan/..., and with neither it skips looking
for manifests in the user dir tree.

loader_get_manifest_files() now takes a relative location inside the
config path eg `/vulkan/implicit_layer.d` and appends that onto the
preferred location. As this was only used for the Linux side of the
loader, I thought this was ok.

vk_loader_platform.h Linux defines got minor adjustments to support
this change, with an added bonus of being more readable.

Have tested locally with debug/release builds, and with the
applications vulkaninfo, Dota2, Talos Principle, and the Tools test
suite.
https://gist.github.com/johndrinkwater/4e7ac2c72db5d52c9d2045c9b2a714df for log output from them.

### DIFF
--- a/loader/vk_loader_platform.h
+++ b/loader/vk_loader_platform.h
@@ -50,42 +50,23 @@
 #define PATH_SEPERATOR ':'
 #define DIRECTORY_SYMBOL '/'
 
-#define VULKAN_ICDCONF_DIR                                                     \
-    "/"                                                                        \
-    "vulkan"                                                                   \
-    "/"                                                                        \
-    "icd.d"
-#define VULKAN_ICD_DIR                                                         \
-    "/"                                                                        \
-    "vulkan"                                                                   \
-    "/"                                                                        \
-    "icd"
-#define VULKAN_ELAYERCONF_DIR                                                  \
-    "/"                                                                        \
-    "vulkan"                                                                   \
-    "/"                                                                        \
-    "explicit_layer.d"
-#define VULKAN_ILAYERCONF_DIR                                                  \
-    "/"                                                                        \
-    "vulkan"                                                                   \
-    "/"                                                                        \
-    "implicit_layer.d"
-#define VULKAN_LAYER_DIR                                                       \
-    "/"                                                                        \
-    "vulkan"                                                                   \
-    "/"                                                                        \
-    "layer"
+#define VULKAN_DIR            "/vulkan/"
+#define VULKAN_ICDCONF_DIR    "icd.d"
+#define VULKAN_ICD_DIR        "icd"
+#define VULKAN_ELAYERCONF_DIR "explicit_layer.d"
+#define VULKAN_ILAYERCONF_DIR "implicit_layer.d"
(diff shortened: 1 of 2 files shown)


karl-lunarg:
Not the right repo, but MarkY can handle it from here.

loader/vk_loader_platform.h
```diff
@@ -94,25 +75,25 @@
...
+    "/usr/" DATADIR VULKAN_DIR VULKAN_ILAYERCONF_DIR
 #define DEFAULT_VK_LAYERS_PATH ""
 #if !defined(LAYERS_SOURCE_PATH)
 #define LAYERS_SOURCE_PATH NULL
 #endif
 #define LAYERS_PATH_ENV "VK_LAYER_PATH"
-#define HOME_VK_DRIVERS_INFO "/.local/share" VULKAN_ICDCONF_DIR
-#define HOME_VK_ELAYERS_INFO "/.local/share" VULKAN_ELAYERCONF_DIR
-#define HOME_VK_ILAYERS_INFO "/.local/share" VULKAN_ILAYERCONF_DIR
+#define HOME_VK_DRIVERS_INFO VULKAN_DIR VULKAN_ICDCONF_DIR
```
MarkY-LunarG: This seems like a change to me that will break non XDG locations.  You've gone from /.local/share/vulkan/icd.d to /vulkan/icd.d/.  That no longer seems like a "home" folder.

johndrinkwater: See the changes in loader.c, wherever the user has defined XDG_DATA_HOME we need to append /vulkan/icd.d/, and for the fallback case it appends that to HOME + .local/share.
Otherwise there need to be further duplicated defines of VULKAN_DIR VULKAN_[ICD|ELAYER|ILAYER]CONF_DIR

johndrinkwater: Thinking HOME_ should be removed from the name now to avoid confusion.

MarkY-LunarG: Not a big deal.  I see what you're talking about, thanks for the clarification.  The process for this is to go through the Khronos LVL repository  [https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers] which is what Karl was talking about.  Then it will eventually be branched into this.  The changes look good from an initial look.  I'll run it through our internal tests and let you know.

johndrinkwater: Ah, thank you.

MarkY-LunarG:
Committed to the LoaderAndValidationLayers repo with commit 227a2c8eae5448646ae7ca14c62945c3ab3f4ccb.  That change will be pulled into this branch on the next trunk merge.

Issue #72, opened 2016-08-06
Linux BUILD_WSI_XLIB_SUPPORT defaults on, cannot turn it off


davidhubbard:
Interesting series of bugs. This first part is not a VulkanTools bug:
```
$ cmake -H. -Bbuild -DCMAKE_BUILD_TYPE=Debug -DBUILD_WSI_XLIB_SUPPORT=On
$ grep XLIB build/CMakeCache.txt 
BUILD_WSI_XLIB_SUPPORT:BOOL=On
$ cd build && make
```
The error I am getting is an interaction between XLib and Qt5:
```
[ 93%] Building CXX object vktrace/src/vktrace_viewer/CMakeFiles/vktraceviewer.dir/main.cpp.o
In file included from /usr/include/X11/Xlib.h:44:0,
                 from VulkanTools/include/vulkan/vk_platform.h:113,
                 from VulkanTools/include/vulkan/vulkan.h:31,
                 from VulkanTools/include/vulkan/vk_icd.h:26,
                 from VulkanTools/vktrace/src/vktrace_common/vktrace_multiplatform.h:28,
                 from VulkanTools/vktrace/src/vktrace_common/vktrace_trace_packet_utils.h:24,
                 from VulkanTools/vktrace/src/vktrace_viewer/vktraceviewer.h:30,
                 from VulkanTools/vktrace/src/vktrace_viewer/main.cpp:24:
/usr/include/qt5/QtCore/qurl.h:139:9: error: expected identifier before numeric constant
         None = 0x0,
         ^
/usr/include/qt5/QtCore/qurl.h:139:9: error: expected ‘}’ before numeric constant
/usr/include/qt5/QtCore/qurl.h:139:9: error: expected unqualified-id before numeric constant
```
The Xlib.h header is maybe [#included in the wrong place](http://lists.qt-project.org/pipermail/development/2013-March/010511.html). I'm not sure.

Ok, so punt on that since I don't plan to use XLib anyway.

```
$ rm -rf build; cmake -H. -Bbuild -DCMAKE_BUILD_TYPE=Debug -DBUILD_WSI_XLIB_SUPPORT=Off
$ grep XLIB build/CMakeCache.txt 
BUILD_WSI_XLIB_SUPPORT:BOOL=Off
$ cd build && make
```

Now this is interesting:
```
[  3%] Building C object icd/nulldrv/CMakeFiles/VK_nulldrv.dir/nulldrv_gpa.c.o
[...]

PR #73, opened 2016-08-06
GH72 Fix XLIB=Off build

davidhubbard:
Here is a potential fix for the XLIB=Off issue (#72).

The comment in vktrace_generate.py, "TODO: need a better way to indicate which extensions should be mapped to which Get*ProcAddr" leads me to believe I could expand this pull request to also do that.

But right now this is just the minimum required to get the XLIB=Off build to compile.


davidhubbard:
I am going to redo this pull reuest, as it now shows ~80 commits (there are only actually 3).

PR #76, opened 2016-08-10
Android screenshot layer

cnorthrop:
Some minor changes to screenshot.cpp file if anyone would like to take a look.  Shouldn't affect desktop behavior, but it is outside the build-android folder so wanted to give folks the chance to provide feedback.


cnorthrop:
cherry-picked to master

PR #87, opened 2016-09-08
PageGuard Optimization

mizhen:
Optimization by using page-guard for speed up capture 

  Background:

     The speed is extremely slow when use vktrace to capture DOOM4. It took over half a day and 900G of trace for a capture from beginning to the game menu.
     The reason that caused such slow capture is DOOM updates a big mapped memory(over 67M) frequently, vktrace copies this memory block to harddrive when DOOM calls vkFlushmappedMemory to update it every time.
     Here we use page guard to record which page of big memory block has been changed and only save those changed pages, it make the capture time reduce to round 15 minutes, the trace file size is round 40G, 
     The Playback time for these trace file is round 7 minutes(on Win10/AMDFury/32GRam/I5 system).

  Page guard process:

     1. The target app call map buffer and get a mapped memory pointer(for example pMemReal).
     
     2. The capturer then allocates a duplicated memory of at least the same size(for example pMemCopy). Copies all data from pMemReal to pMemCopy
     
     3. The capturer adds page guard to pMemCopy for every page.Then returns the pMemCopy(not pMemReal) to the target app.

     4. When the target app uses pMemCopy to write, page guard will be triggered and page guard handler will records which page has been changed.
     
     5. When the target app uses pMemCopy to read data, page guard will be triggered and page guard handler records which page the target app wants to read. And only copy that page from real mapped memory to copied memory so after page guard exception, target app will read right data.
     
     6. When the target app calls CPU->GPU synchronization(vkFlushMappedMemoryRanges and submits queue), the capturer saves all changed pages and also copies back these changed pages to pMemReal from pMemCopy, resets all page guard which triggered by write.
     
     7. When the target app calls GPU->CPU synchronization(vkInvalidateMappedMemoryRanges, also include vkQueueSubmit which is before that synchronization happen), clear all read array and resets all page guard which triggered by read.
     
     8. When the target app calls to unmap the memory, the capturer saves all changed pages to HD and also copies back these changed pages to pMemReal from pMemCopy. Finally removes all page guards and frees pMemCopy.

  Known limitations:

     1. for a page which is in mapped memory, if target app first read it before write, then do GPU->CPU synchronization like vkFlushMappedMemoryRanges trigger flush copied mapped memory to (real)mapped memory, that page will not be recorded as changed page.

     2. one page accessed by a thread and the access just happen when another thread already finish copying data to real mapped memory for that page but haven't reset page guard of that page, that page will not be recorded as changed page.


davidlunarg:
I have reviewed this pull request, but there's lots of code, it's difficult to review every single line of code for accuracy. We have run numerous tests on vktrace/replay and have found no problems in the behavior of vktrace/replay.

I did identify several higher level issues:

vkrace_lib_pageguard.cpp/76:
   Shouldn't the pageguard always be enabled?

vkrace_lib_pageguard.cpp/85:
   Pageguard range size should be set by default to a reasonable value, and then possibly set to
   a different value via command line option to vktrace.

What GPUs has pageguard implementation been tested on?

Add Author to header of each file.

vktrace_lib_pageguard.cp/93:
   The line:
      ref_target_range_size() = rangesize;
   is quite confusing. It's technically correct, but it would be much easier to read
   if a static variable was used instead of a function.

vktrace_lib_trace.cpp
vkreplay_vkreplay.cpp
vkreplay_main.cpp
vktrace_generate.py:
    Change line endings to Unix line endings. 

vktrace_pageguard_memorycopy.h/27:
   USE_PAGEGUARD_SPEEDUP is enabled iff WIN32 is true.  Could WIN32 be used instead
   of USE_PAGEGUARD_SPEEDUP?

vktrace_pageguard_memorycopy.cpp/27:
   How was SIZE_LIMIT_TO_USE_OPTIMIZATION determined?

vktrace_pageguard_memorycopy.cpp/85:
   How was PARALLEL_INVOKE_NUM determined?  I'll suggest it be set to one less than the number
   of CPUs.

Replace all instances of memcpy with opt_memcpy?

ardaLunarG:
Patch merged to mainline but we keep this open as it includes David's comments needs to be addressed.

ardaLunarG:
Pushed two patches
86dec7a
7c996b7
To mainline.
Other two patches, merge and line endings has a lot of conflicts with main stream so left out.
I will fix line endings later on main stream

PR #97, opened 2016-09-30
build: add dependencies to enable parallel build

stroyan:
This gets parallel windows builds working.

Fix parallel build of layersvt
 Change layer dependencies to prevent collisions of scripts creating headers.
 Use dependency on one custom target for each group of headers built by a custom command.
 The layers json targets need to depend on the layers targets.
 They expect to copy into a directory created by those targets.
 Make dependency of vk_safe_struct.cpp indirect through generate_vk_layer_helpers.
 That will wait for a single build of the generated files.
vktrace_layer json copy depends on vktrace layer.
vktrace_replay depends on vulkan loader project.

### DIFF
--- a/layersvt/CMakeLists.txt
+++ b/layersvt/CMakeLists.txt
@@ -43,6 +43,7 @@ if (WIN32)
                     COMMAND copy ${src_json} ${dst_json}
                     VERBATIM
                     )
+                add_dependencies(${config_file}-json ${config_file})
             endforeach(config_file)
         else()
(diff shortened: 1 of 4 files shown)


karl-lunarg:
Adding Arda for review purposes.

ardaLunarG:
Approved! Looks good.

karl-lunarg:
I think Arda ran this on our test farm, correct?

ardaLunarG:
Yes, that is right, compile went fine.

stroyan:
Cherry-picked to master

Issue #42, opened 2016-05-13
miss alignment for some arm cpu vfp load instructions


jyangsh:
Some arm vfp load instructions, e.g. vldr, requires the address of the source to align to 4 bytes.

But sometimes, the address decomposed from the packet is not 4 bytes alignment, e.g., the 4 float type members in the VkPipelineRasterizationStateCreateInfo structure, that is causing the vldr instruction fail.

karl-lunarg:
Is this for vktrace/vkreplay?

jyangsh:
Yes, the issue is for vktrace and vkreplay. Sorry for missing this info.:)

davidlunarg:
I have a proposed fix, which I have verified using assert(). Could you try the david_arm_align branch in VulkanTools and let me know if this addresses the problem?

ZLixing:
Hi,  I have tried the branch on arm platform but failed with the assertion (assert((size&0x3) == 0)).

davidlunarg:
@ZLixing : Thanks for trying it. I made a couple of more changes since I sent you the proposed fix, and it is now on master. Could you try it? And if it fails on an assert, could you send me a stack trace?

ZLixing:
@davidlunarg I merged the newest master to my project and no problem happens, thanks for help

davidlunarg:
Glad it works for you. Thanks for testing it, I don't have an ARM system readily available.

PR #98, opened 2016-10-04
Android arm32

cnorthrop:
This change to align struct members gets 32-bit tracing unblocked on Android.


ardaLunarG:
Agree we need to bump the version number.
Cody is it possible to keep old structures and add new ones with 32bit or ARM ifdefs around?

Tony-LunarG:
I knew backwards compatibility would break at some point and have been prepared to retrace.  Are these all the alignment fixes we're going to need or are there more to come?  (Yes, I realize it's impossible to answer that question).

cnorthrop:
Arda - I believe the structures have to match between x86 and ARM, so we couldn't wrap based on arch.  We could probably add new structs that are only used by 32-bit builds, but I'd much rather just get the correct fix in and buy Tony a beer.

Tony - I just grepped for uint64_t and "long long" under vktrace and eyeballed the results, didn't see any other struct members.   If you want, I can try out more content to see if any other problems shake out before going through a full retrace cycle.

Having a manageable retrace strategy is one of the challenges of supporting this project.  It is worth thinking about with every change.

ardaLunarG:
Cody with this changes, we collect trace 32-bit and we have to replay on 32-bit vkreplay right?
Or 64-bit vkreplay would work?

stroyan:
It may be easier to create a filter that converts from an old format to a new format rather than retracing.
That will depend on how many traces you need to recreate.

Tony-LunarG:
Or use the new tracer to trace the replay of the trace with the old replayer

ardaLunarG:
More fundamentally this change is adding an aligned 8 attribute to match up x86 vs ARM only.
If there is a different architecture (mips, sparc etc etc) we have to keep adding this and it will be hard to manage. We have to go to direction using "__packed__" to get rid of cross platform differences and use filter inside based on the running platform as suggested.

karl-lunarg:
Doesn't the problem go far deeper than the two structs Cody fixed up?  For example, the packet_vkWaitForFences struct contains a uint64_t with a bool32 before it.  The offset of the uint64 could easily be different, based on the architecture.   There are many more cases like this one, where I think you would want to add an alignment attribute.

Also, setting an alignment attribute on an individual structure member may guarantee that it always has an aligned offset, but it doesn't guarantee that that offset will be the same across architectures.  Suppose you had:

int16 x1;
int16 x2;
int16 x3;
int16 x4;
int64 y __aligned(8);

If one architecture pads int16 to 32 bits, then the offset of y would be 128.
If another architecture does not pad int16 to 32 bits, then the offset of y would be 64.
Both are aligned(8), but different.

For the two structs in question, this won't matter, but others would need to be examined for the possibility.

I hate to say it, but the thorough way of solving this is to define an architecture-independent specification for the trace file and use something other than C structs to access the elements.  This specification would define structs as lists of elements like CARD8, CARD16, CARD32, CARD64, etc, which are "cardinal" basic data types that are packed together in the trace file.  For each member in a struct, there would be a cardinal type and its offset from the beginning of the struct, as stored in the trace file.   The code that write/reads the trace file would have to insert/extract each member individually by using the offset to locate the data in the trace file as it copies it from/to local memory.  This would be a crazy amount of work though.  This is a lot like the X11 protocol definition, for those who remember such things. 

The code that does the accessing could be made with architecture-specific macros so that natural machine load and stores are used most of the time.  Else, memcpy().  (Maybe look at Xlib/Xserver protocol code for ideas?)

As for Arda's suggestion, wouldn't using packed guarantee unaligned accesses on ARM?  I guess I don't understand the "filter inside" part of this, unless it meant using special code to copy int64 data only on the ARM.  If that problem can be solved, using packed seems like an easier way to get there, as long as packed worked the same across architectures.

stroyan:
Very few of the struct packings have been an issue so far because for simple trace and replay the structs are all being created and then read by code compiled on the same architecture.  The only problem that showed up was with the vktrace_trace_file_header struct that was written by x86 and read by arm32.
The problem will be much larger if the x86 executable starts to read structs written by arm32.
That would happen for a vktrace viewer on x86 paired with an arm trace.
Using a packed struct could lead to problems with architectures that have alignment requirements.
Issue #42 already showed one of those issues with structs that were not adequately aligned.
If the contents of structs themselves are tightly packed it could require that replay would copy the data to new locations that are properly aligned for the architecture.  That would entail a lot of code and slower execution.

stroyan:
If someone wanted to write and read vktrace files in a portable way, then google protocol buffers would probably be the shortest path to accomplish that.

https://developers.google.com/protocol-buffers/docs/overview

ardaLunarG:
Indeed as Vulkan.h structures ensures right alignment would be provided. Thus only file_header might trigger issue here. 

Contents of structs are should be kept in tightly packed (in the trace file) and these need to be copied to new location which properly aligned whichever arch replay works on. This is how it should work. 
There is some overhead here but there is no workaround on this. 
Right now, ARM vs X86 but in the future other archs might require different alignments and maintain that code would bring even more overhead.

cnorthrop:
I updated the branch with a fix reduced to just the known cross platform requirement, vktrace_trace_file_header.   I've tested that this works for Android and builds on Windows.

Are there any 32-bit traces out there?  If not, this change doesn't really affect 64-bit traces.  The alignment for x64 and arm64 was already 8 bytes.  Maybe no need for a version bump.

This series doesn't come close to solving the broader problems, which I agree protocol buffers may be a great match for.

ardaLunarG:
Thanks for the change Cody,

32-bit Windows compile broken, showing up in vktraceviewer. We are analyzing the compile error now.

ardaLunarG:
Here is the error for reference:

c:\users\lunarg\.jenkins\workspace\gerrittools\bits\32\build_mode\debug\label\master\vktrace\src\vktrace_viewer\vktraceviewer.h(112): error C2719: 'fileInfo': formal parameter with requested alignment of 8 won't be aligned (compiling source file C:\Users\LunarG\.jenkins\workspace\GerritTools\BITS\32\BUILD_MODE\Debug\label\master\build32\vktrace\src\vktrace_viewer\moc_vktraceviewer.cpp)

ardaLunarG:
David find the issue, we are working on the fix.

jyangsh:
Great! It works. Though there are also some uint64_t fields in vktrace_trace_packet_header structure, they seem not cause some alignment issues.

jyangsh:
Met another issue. The cube demo cannot be traced twice. The cube demo hangs after kill it and re-run it again with the trace layer is enabled.

jyangsh:
The _pBody_ of _vktrace_trace_packet_header_ seems also need align to 8 bytes. It looks x86 aligns it to 4 bytes, but arm32 aligns it to 8 bytes.

**Output from vktrace32 on x86 host:**
vktrace info: vktrace_trace_packet_header size = 68, uintptr_t size of = 4
vktrace info: offsetof(vktrace_trace_packet_header, pBody) = 64

**Output from vkreplay on arm32 android system:**
10-12 04:18:10.273  2529  2595 V vkreplay: vktrace_trace_packet_header size = 72, uintptr_t size of = 4
10-12 04:18:10.273  2529  2595 V vkreplay: offsetof(vktrace_trace_packet_header, pBody) = 64

ardaLunarG:
Merged

ardaLunarG:
jyangsh if you experience issues please feel free to file a bug.

jyangsh:
@ardaLunarG, Sure, I will double check it.

PR #101, opened 2016-10-18
toolchain: Get LunarGLASS building again

cnorthrop:
Not sure which file changed, but the second "git checkout" in the LunarGLASS path was complaining about overwriting files.  I don't think we need two checkouts, so I merged them to one forced checkout at a specific version.  This gets the toolchain building again.  vktracereplay.sh passes on Nvidia Linux.


lenny-lunarg:
This looks good to me

PR #103, opened 2016-10-25
API Dump Rework

lenny-lunarg:
This branch reworks the API dump to use the xml vulkan spec when generating the API dump. It also adds some functions that weren't getting dumped properly, modifies the output to look more readable, and adds several user settings to the layer. Going forward, we want to create an output other than text, but that hasn't been implemented yet in this pull request.

@cnorthrop, I've verified that this branch builds for android, but I haven't tested that it actually runs fine. If you could run it through your tests whenever you have the chance, I'd appreciate it.

I've attached a sample API dump I generated from cube with the default settings, for whoever wants to see the output without building anything: [sample_api_dump.txt](https://github.com/LunarG/VulkanTools/files/551995/sample_api_dump.txt)


layersvt/api_dump.h
```diff
@@ -1,149 +1,344 @@
...
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- *
- * Author: Courtney Goeltzenleuchter <courtney@LunarG.com>
- * Author: Jon Ashburn <jon@lunarg.com>
- *
- */
+
```
mark-lunarg: Lenny, we'll need a copyright notice in this guy too.

vt_genvk.py
```diff
@@ -0,0 +1,236 @@
...
+
+    # Defaults for generating re-inclusion protection wrappers (or not)
+    protectFile = protect
+    protectFeature = protect
+    protectProto = protect
+
+    #
+    # VulkanTools Generators
+    
+    # Options for api dump layer
```
mark-lunarg: I didn't check, but hopefully this is identical to the vulkan-docs genvk.py except for removing the other genOpts and adding this new one to make merging genvk.py changes easier.

api_dump_generator.py
```diff
@@ -0,0 +1,1281 @@
+
```
mark-lunarg: Copyright (and your name) in this guy too.

Tony-LunarG:
I'm seeing problems dumping the secondary_command_buffer sample, because WriteDescriptorSet thinks that both pImageInfo and pBufferInfo are valid, but when descriptorType is UNIFORM_BUFFER, only pBufferInfo is valid and pImageInfo should be ignored.  There are several calls around descriptor sets where the type determines which pointers are valid.

cnorthrop:
Latest branch builds for me on Android.

There is an api_dump "test" in layer_validation_tests.cpp, although it is disabled by default.  Search for API_DUMP_TESTS.   I don't have time to try it at the moment, but I will soon.  I have a pending fix for api_dump on Windows for Android.

Tony-LunarG:
Latest commit fixed my issue.

lenny-lunarg:
I've fixed the missing copyright notices, plus the VkWriteDescriptorSets issue. The only problem is that this doesn't build for Android when on Windows. However, I also can't build the master branch for Android when on Windows, so I'm just going to merge this branch.

cnorthrop:
I have Windows working locally (before your changes).  Will rebase it and get it uploaded.  Still trying to get Windows desktop to build.  Stumbling on ImageMagick problems.

PR #105, opened 2016-10-29
WIP - vkvalidator: Add new tools application

MarkY-LunarG:
Create an application (called vkvalidator) which will perform
some validation of the Vulkan setup of a machine and spit out
a resulting HTML file.  This file should be used when
submitting Vulkan bugs related to the loader, layers, or
samples.

Eventually, it would be nice to run this tool as part of the
LunarG Vulkan SDK install to validate that everything
is working properly after installation is complete.

To run it, simply run with no parameters and it will write
a resulting file to the current directory, called
'vkvalidator.html'.  If, for whatever reason, it is unable
to write to the current directory, it will write to the
user's home folder.

Change-Id: I37ae7d103cd319e9c92a5de1fdeb65a39ddc6a71


mark-lunarg:
- Can you add comments to the functions/modules in the validator source file?  Also, we put off adding vkjson to the source tree for 8 months because we already had cJSON in the loader.
- This looks like it adds jsonCPP or something -- that will be **two json-libraries too many**.
- Also, this commit should probably be broken up -- adding the new files, etc., and then a final one making the cmake changes to get things working.

karl-lunarg:
I wish this had a different name.  "Validate" has an important alternate meaning in Vulkan.  Some users may have trouble keeping it straight.  Even "vkInstallChecker" would be better and maybe even more accurate.

ardaLunarG:
Wherever string used for "PATH" length are inconsistent.
This should be defined and used consistently.
Ex:
char full_file[512];
char orig_dir[1024];
char path[1035];

Or MAX_PATH should be used defined in vktrace_common.

Additionally throughout the code strings defined in various length.
  char home_path[**256**];
And this used later on as
  0 != GetEnvironmentVariableA("HOMEPATH", home_path, **255**))

If we define these values and use definition instead of real numbers that would save a lot of hassle. There multiple instances of these.
EX:
char home_path[HOME_PATH];
..
0 != GetEnvironmentVariableA("HOMEPATH", home_path, HOME_PATH-1))

ardaLunarG:
It would be much safer if we init this value to 0: 
  uint32_t gpu_count;
 
status =
        vkEnumeratePhysicalDevices(global_items.instance, &gpu_count, NULL);
    if (status) {
        sprintf(generic_string, "ERROR: Failed to query - %d", status);
        PrintTableElement(generic_string);
        failed = true;
    } 
If this fails we print error and set fail flag but later on we call it again.
Should not we exit in this case?

If gpu_count remains is random value it might crash badly.

karl-lunarg:
> This looks like it adds jsonCPP or something -- that will be two json-libraries too many.

How hard would it be to change this tool to use cJSON from ../loader?

MarkY-LunarG:
It would **not be trivial** to change this to use cJSON.  The problem is that JsonCPP incorporates a lot of the parsing internally, while cJSON isn't so clean.

vkvalidator/vkvalidator.vcxproj.user
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
```
MarkY-LunarG: I had intentionally added it for the VK_LAYER_PATH setting.  I could get rid of it as it's no skin off my teeth either way.

mark-lunarg:
My preference has been to remove cJSON from the loader and replace it with VKJSON.  In this case, replacing either or both of these with JsonCPP would be acceptable if we have to use it, but three?  Blick.

karl-lunarg:
VKJSON is a library used to read/write Vulkan objects from/to JSON format; it can't be used as a general JSON parser.  It in fact uses cJSON to do the low-level JSON parsing/writing.  So, I don't see how the loader can replace its use of cJSON with VKJSON when VKJSON itself uses cJSON.  

Further, VKJSON can't read arbitrary JSON files and so it can't read manifest files in the same way the loader reads them with cJSON.  One could add to VKJSON the ability to read/write icd and layer manifest files and have the loader use VKJSON for that.  But you would still have cJSON sitting around, because VKJSON uses it, and it isn't clear that the C loader could use the VKJSON C++ code.

If you really want to get rid of cJSON and adopt JsonCPP instead, you'd probably have to:

* Make sure that the loader can use JsonCPP (C calling C++)
* Pull JsonCPP with update_external_sources in LVL AND VulkanTools
* Migrate VKJSON from cJSON to JsonCPP (probably pretty easy)

For this project, we're talking about pulling JsonCPP into VulkanTools, not LVL.  This keeps the "ugly" extra parser out of LVL, at least util we're ready to do the above migration.  In the meantime, it seems expedient to pull JsonCPP into VT now and get on with finishing this tool.

mark-lunarg:
This all sounds good to me.

vkvalidator/vkvalidator.vcxproj.user
MarkY-LunarG: FYI, the .gitignore does not ignore the .user file.

PR #106, opened 2016-11-03
Hologram problem

mizhen:
there were missing teapots in the replay. sometimes, the Hologram process would crash during trace and the replay process would crash during replay.

XCAP-409

Change-Id: I752a95ed97889526c8e0b2c1214005dcea59b289

### DIFF
--- a/vktrace/src/vktrace_layer/vktrace_lib_pageguard.cpp
+++ b/vktrace/src/vktrace_layer/vktrace_lib_pageguard.cpp
@@ -45,15 +45,6 @@ void pageguardExit()
 }
 #if defined(WIN32) //page guard solution for windows
 
-#define PAGEGUARD_TARGET_RANGE_SIZE_CONTROL
-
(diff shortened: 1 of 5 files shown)


davidlunarg:
In method PageGuardCapture::vkMapMemoryPageGuardHandle there is a statement:
    MapMemoryOffset[memory] = offset;
I don't see a place where this element is erased or otherwise removed from unordered_map. Do we have a memory leak here?

Also, there are three unordered_maps in the class PageGuardCapture, all indexed by VkDeviceMemory. Each time one element is accessed, usually all three elements need to be accessed, resulting in three separate searches. If the three unordered_maps were all put in a single structure, and  only one search would be needed to access all three elements.

davidlunarg:
Arda and I discussed this. We'll merge the pull request, then I'll fix the memory leak. The performance improvement suggestion we can do later.

mizhen:
yes, need  erase this element in vkUnmapMemoryPageGuardHandle like MapMemoryPtr.

for the three maps, MapMemory is only used for mapped memory which already add page guard. 
MemoryPtr, MapMemoryOffset is used for all mapped memory, they can be combined in one to save search time.

ardaLunarG:
merged

PR #113, opened 2016-11-15
Pageguard port to Linux

davidlunarg:
This is a port of the Windows pageguard code to Linux. Rather than using signals/exceptions, it uses /proc/pid/pagemaps and /proc/pid/clear_refs to query and clear the dirty bits of mapped memory. Like Windows, it uses a shadow copies of device memory.


vktrace/src/vktrace_layer/vktrace_lib_trace.cpp
```diff
@@ -175,6 +265,8 @@ VKTRACER_EXPORT VKAPI_ATTR VkResult VKAPI_CALL __HOOKED_vkMapMemory(
     VkFlags flags,
     void** ppData)
 {
+    static void *d1=getenv("D1");
+    while (d1) d1=d1;
```
cnorthrop: This throws an error with clang:
    error: explicitly assigning value of variable of type 'void *' to itself [-Werror,-Wself-assign]
        while (d1) d1=d1;
Do you still need this code?

davidlunarg: Oops, that was debug code. I should have removed it before committing. I'll remove it.

davidlunarg:
Updated this pull request, addressing the issues raised by Cody and Mike. This version reads dirty bits from the pagemap file and clears them using the clear_refs file, but also sets up a signal handler to catch changes to mapped memory that might occur between the read and the clear.

vktrace/src/vktrace_layer/vktrace_lib_trace.cpp
```diff
@@ -136,6 +138,161 @@ void *strip_create_extensions(const void *pNext)
...
+static vktrace_sem_id sighAddrListSem = NULL;
+
+static void segvHandler(int sig, siginfo_t *si, void *ununsed)
+{
+    size_t pageSize = getpagesize();
+
+    // Note: we use sem_wait and mprotect inside a signal handler.
+    // This is not POSIX compliant, but works on Linux.
+
+    // Add the addr that caused the segv to sighAddrList
```
stroyan: It would be safer to check if the faulting address is actually in the list of mapped regions.
Then tracing an application that dies with its own SIGSEGV problem during getMappedDirtyPagesLinux would behave better.
An out of bounds address could be dealt with by either calling a non-NULL previous signal handler returned by sigaction or by flushing buffers and calling VKTRACE_FATAL_ERROR.

vktrace/src/vktrace_layer/vktrace_lib_trace.cpp
```diff
@@ -136,6 +138,161 @@ void *strip_create_extensions(const void *pNext)
...
+    // Iterate through all mapped memory allocations.
+    // Check dirty bits of each page in each mapped memory allocation.
+    for (std::unordered_map< VkDeviceMemory, PageGuardMappedMemory >::iterator it = getPageGuardControlInstance().getMapMemory().begin();
+         it != getPageGuardControlInstance().getMapMemory().end();
+         it++)
+    {
+        pMappedMem = &(it->second);
+        mappedMemory = pMappedMem->getMappedMemory();
+        pEntry=find_mem_info_entry(mappedMemory);
+        addr=pEntry->pData+pEntry->rangeOffset;
```
stroyan: The value in pData+pEntry->rangeOffset is the offset within a buffer given to vkMapMemory.
But it should not be used as an offset from pEntry->pData.
The value in pEntry->pData is the address returned by vkMapMemory with the buffer offset already applied.

vktrace/src/vktrace_layer/vktrace_lib_trace.cpp
```diff
@@ -136,6 +138,161 @@ void *strip_create_extensions(const void *pNext)
...
+        alignedAddrStart = (PBYTE)((uint64_t)addr & ~(pageSize-1));
+        alignedAddrEnd = (PBYTE)(((uint64_t)addr + pEntry->rangeSize + pageSize) & ~(pageSize-1));
```
stroyan: The calculation of alignedAddrEnd should add one less byte before masking.
(PBYTE)(((uint64_t)addr + pEntry->rangeSize + pageSize - 1) & ~(pageSize-1))
The for addr=0, pEntry->rangeSize = 4096, pageSize = 4096, alignedAddrEnd will be 4096 instead of 8192.
That will avoid mprotect giving ENOMEM for going one page past the end of a mapped region.

vktrace/src/vktrace_layer/vktrace_lib_trace.cpp
```diff
@@ -136,6 +138,161 @@ void *strip_create_extensions(const void *pNext)
...
+
+    // Re-enable write permission for all mapped memory
+    for (std::unordered_map< VkDeviceMemory, PageGuardMappedMemory >::iterator it = getPageGuardControlInstance().getMapMemory().begin();
+         it != getPageGuardControlInstance().getMapMemory().end();
+         it++)
+    {
+        pMappedMem = &(it->second);
+        mappedMemory = pMappedMem->getMappedMemory();
+        pEntry=find_mem_info_entry(mappedMemory);
+        addr=pEntry->pData+pEntry->rangeOffset;
```
stroyan: Another pEntry->rangeOffset to remove.

vktrace/src/vktrace_layer/vktrace_lib_trace.cpp
```diff
@@ -136,6 +138,161 @@ void *strip_create_extensions(const void *pNext)
...
+        alignedAddrStart = (PBYTE)((uint64_t)addr & ~(pageSize-1));
+        alignedAddrEnd = (PBYTE)(((uint64_t)addr + pEntry->rangeSize + pageSize) & ~(pageSize-1));
```
stroyan: Another alignedAddrEnd calculation to change.

vktrace/src/vktrace_layer/vktrace_lib_pageguardcapture.cpp
```diff
@@ -25,13 +25,20 @@
...
     EmptyChangedInfoArray.offset = 0;
     EmptyChangedInfoArray.length = 0;
+
+#if defined(PLATFORM_LINUX)
+    // Open the /proc/self/clear_refs file. We'll write to that file
+    // when we want to clear all the page dirty bits in /proc/self/pagemap.
+    clearRefsFd = open("/proc/self/clear_refs", O_WRONLY);
+    if (clearRefsFd < 0)
+        VKTRACE_FATAL_ERROR("Open of /proc/self/clear_refs failed.");
+#endif
```
stroyan: It would be good to call pageRefsDirtyClear during initialization.
Some older kernels (like android uses) implement /proc/self/pagemap, but use a different definition of the pagemap bits until the first writing of "4" to clear_refs.

davidlunarg:
Just pushed changes addressing Mike's most recent feedback. I think there is still an issue somewhere with address calculations - if I trace smoketest for an extended period of time, the output becomes jumpy. I'm investigating this.

vktrace/src/vktrace_layer/vktrace_lib_trace.cpp
```diff
@@ -136,6 +138,166 @@ void *strip_create_extensions(const void *pNext)
...
+
+        if (!vktrace_sem_create(&sighAddrListSem, 1))
+            VKTRACE_FATAL_ERROR("Failed to create sighAddrListSem.");
+
+        sigAction.sa_sigaction = segvHandler;
+        sigfillset(&sigAction.sa_mask);
+        sigAction.sa_flags = SA_SIGINFO;
+        if (0 != sigaction(SIGSEGV, &sigAction, NULL))
+            VKTRACE_FATAL_ERROR("sigaction sys call failed.");
+        getPageGuardControlInstance().pageRefsDirtyClear();
```
davidlunarg: As soon as I submitted this revision, I realized the call to pageRefsDirtyClear should not be done here. It should be moved to the top of  __HOOKED__vkMapMemory.

vktrace/src/vktrace_layer/vktrace_lib_trace.cpp
```diff
@@ -136,6 +138,166 @@ void *strip_create_extensions(const void *pNext)
...
+
+    // Add the addr that caused the segv to sighAddrList.
+    // It may not be in a mapped page we are tracking. If it
+    // isn't, we'll detect it later and generate an error then.
+    vktrace_sem_wait(sighAddrListSem);
+    sighAddrList.emplace_front(si->si_addr);
+    vktrace_sem_post(sighAddrListSem);
+
+    // Change protection of this page to allow the write to proceed
+    mprotect((void *)((uint64_t)si->si_addr & ~(pageSize-1)), pageSize, PROT_READ|PROT_WRITE);
```
stroyan: If this mprotect fails, the signal handler should call VKTRACE_FATAL_ERROR.
If the address cannot be made writable then returning from the signal handler will just retry the faulting instruction.
That will go right back into the signal handler again and again.

davidlunarg: A call to VKTRACE_FATAL_ERROR involves writing to a the trace file and sending a message to stdout, things I would rather not do inside a signal handler. But we're about to exit the process anyway, so I think it should be OK.

cnorthrop:
This isn't working just yet for Android.  I'll get a fix in tomorrow.  It errors out (root or not) with:

    vktrace : Errors: Write to /proc/self/clear_refs failed. Fatal error.

jyangsh:
12-01 07:56:12.514  2016  2031 I vktrace : Errors: Write to /proc/self/clear_refs failed. Fatal error.

+1

Android 6.0 AOSP + Linux kernel version 3.10.0

PR #114, opened 2016-11-18
android: Update vktrace to use abstract socket

cnorthrop:
This allows tracing any Vulkan application, not just those that provide INTERNET permission.

Alter your adb command to use it:

    adb reverse localabstract:vktrace tcp:34201


cnorthrop:
cherry-picked to master

davidlunarg:
+1

PR #119, opened 2016-12-01
Remove ICD from VulkanTools repo

mark-lunarg:
Tried to catch all the references, but might've missed a spot or two.  Please let me know if anybody sees anything out of place or missing.


karl-lunarg:
:+1:

vinjn:
Where can I find the removed ICD?

karl-lunarg:
The repo was tagged with Vulkan-ICD-Snapshot before the ICD was removed.   If you checkout that tag, it should work as well as it did before.  But it will be tied to the Vulkan spec version at that tag and obviously won't support anything added to Vulkan since.

PR #132, opened 2016-12-19
vkreplay: Add -o and --Open as new cmd line args to open trace files.

plohrmann:
Marked -t and --TraceFileName as deprecated, in favor of using -o and --Open.

This makes vkreplay match the args of vktrace and vktraceviewer, and logically makes more sense.


ardaLunarG:
merged

PR #133, opened 2016-12-19
vktrace: Modify generated filenames when in servermode

plohrmann:
Previously the trace files generated when in servermode would automatically get a "0" appended to the file name, which is non-intuitive that the filename specified on the command line is not the same as the
output file. Follow-up trace files would increment this number to "1", "2", etc, but there was actually a bug that would prevent a second app from connecting to the server.

This change modifies this behavior so that the first output trace file will be exactly what the user specified at the command line, and any additional trace files will be generated with a "-1", "-2" appended to the filename.

I've also fixed the bug that was preventing a second app from connecting to the server - the recording thread was not posting a message to the main thread to indicate that the recording was completed.

I also confirmed that (on Windows) the normal mode of launching vktrace continues to work as expected.

Another added feature is that a log message is printed out to inform the user of the trace file name that is generated. None of the vktrace messages get printed when vktrace is run in normal mode from the command line, but it works correctly when launched from Visual Studio. Since this is only when run in normal mode, its not really a concern since the user told vktrace exactly which trace file name to use, and the same name is being output.

XCAP-463

Change-Id: I75c664c28dd0b6c720823d2f281dd20e3bc96cff


ardaLunarG:
merged

PR #135, opened 2016-12-21
vktrace screenshots: add an ability to generate screenshots with specified frame range and interval

mizhen:
This feature provides an ability for vktrace/vkreplay to generate screenshots with specified frame range and interval, and also keep compatible with existing -s option. With this feature, the -s command line option for screenshots can generate screenshots with a list(existing option) or a specified frame range and interval.

 Command line option

--ScreenShot  <start_frame> - <count> - <interval>
--ScreenShot  <frame-1>, <frame-2>, ... <frame-n> 
             or
-s  <start_frame> - <count> - <interval>
-s  <frame-1>, <frame-2>, ... <frame-n>

XCAP-379

Change-Id: Ieeb86b33140b771ada8e9b52866839fbaadde74a

### DIFF
--- a/layersvt/screenshot.cpp
+++ b/layersvt/screenshot.cpp
@@ -156,6 +156,18 @@ static unordered_map<VkPhysicalDevice, PhysDeviceMapStruct *> physDeviceMap;
 // set: list of frames to take screenshots without duplication.
 static set<int> screenshotFrames;
 
+static const int SCREEN_SHOT_FRAMES_INTERVAL_DEFAULT = 100;
+static const int SCREEN_SHOT_FRAMES_UNLIMITED = -1;
+typedef struct {
+    bool valid;
+    int startFrame;// the range begin from (include) this frame.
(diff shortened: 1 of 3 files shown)


davidlunarg:
It would be desirable to allow an option such as -s 10-2, specifying a start frame (10 in this case), and a frame count (2), and use the interval default.

Need to add code to correctly handle the error condition identified at line 1065 of screenshot.cpp.

Reading the code, it looks like it there might be problems if one or more of the startframe, count, or interval are negative, i.e. sscanf will succeed in parsing -s 10--2-1, and the count would then be -2.

Why an interval default of 100? It seems 1 would be more intuitive.

The -s option allows an arg of "all", a count of 0, and an interval default of 100. This can only figured out by reading the code. The command help output should probably document these.

ardaLunarG:
Abandoning as revised on 
https://github.com/LunarG/VulkanTools/pull/141

PR #139, opened 2017-01-04
layers: screenshot: Update doc and rename env var

joey-lunarg:
Renamed _VK_SCREENSHOT to VK_SCREENSHOT_LAYERS. Old name will still work. Some
additional refactoring. Updated README.md

Change-Id: I20ef445b8936a294dc2e12fab5e4b966977a072b

### DIFF
--- a/layersvt/README.md
+++ b/layersvt/README.md
@@ -27,7 +27,8 @@ Note that some layers are code-generated and will therefore exist in the directo
 (build dir)/layers/api_dump.cpp (name=VK_LAYER_LUNARG_api_dump) - print out API calls along with parameter values
 
 ### Capture Screenshots
-layersvt/screenshot.cpp - utility layer used to capture and save screenshots of running applications
+layersvt/screenshot.cpp (name='VK_LAYER_LUNARG_screenshot') - utility layer used to capture and save screenshots of running applications. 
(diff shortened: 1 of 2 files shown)


cnorthrop:
Looking at the diff, you've renamed it to VK_SCREENSHOT_FRAMES, which makes sense.  VK_SCREENSHOT_LAYERS does not, so you'll need to update the commit message.

I'll test this on Android and let you know how it goes.

cnorthrop:
Broken on Android.  env_var_old is only defined with Windows and Linux, but is used unconditionally in QueuePresentKHR.  You can either make the use also conditional, or add an env_var_old to Android section and use the same string.  The latter is cleaner, if redundant.

joey-lunarg:
Thanks for the feedback Cody-
Fixed the message and added the dummy variable for android. Lemme know if this fixes it.

cnorthrop:
Thanks, Joey, this works now.  Although I would prefer the Android env_var_old to be "debug.vulkan.screenshot", just like env_var, since we never supported the desktop var.
How about we just use:

    -const char *env_var_old = "_VK_SCREENSHOT";
     const char* env_var = "debug.vulkan.screenshot";
    +const char *env_var_old = env_var;

joey-lunarg:
Ok makes sense, updated

ardaLunarG:
Hey Cody,
Once you are OK please +1 and I will merge the pull req.

ardaLunarG:
merged

PR #143, opened 2017-01-10
docs: Add a missing 32-bit dependency

cnorthrop:
Ran into this while setting up a different machine for VulkanTools builds.  Cmake was failing for vktrace32 with "pthread not found".  This change installs libpthreads.so for i386.


cnorthrop:
-1

This didn't quite fix the issue yet.

ardaLunarG:
Cody,

sudo apt-get install libc6-dev-i386
and
sudo apt-get install g++-multilib

Does this solve your issue?

cnorthrop:
Ugh, it was just stale CMake cache files.  I was missing a cleanup step with the setup I was testing with.  I've got it fixed but it is unearthing some other build failures that we've been missing.

cnorthrop:
cherry-picked to master

cnorthrop:
I did eventually hit the missing requirement for "g++-multilib".  I just added it.  Thanks!

PR #145, opened 2017-01-12
tests: Add vktrace/replay tests for smoketest

joey-lunarg:
This will test with PMB on, and PMB off with both push constants and
VkFlushMappedMemoryRanges.

Change-Id: I2a5726c65410722994bccca431f4dda42b8a1098


ardaLunarG:
merged

PR #155, opened 2017-01-24
Switch VT to use common codegen dependencies/macros/targets

mark-lunarg:
This catches VulkanTools up with LVL in using the common definitions.  Cody, can you run this through the wringer?  Thanks!


mark-lunarg:
Pushed to Master.  Thanks Cody!

PR #158, opened 2017-01-24
VT generated files rework

mark-lunarg:
Common code-gen'd files now end up in single location, including those in vktrace.  VKtrace code-gen'd files now end up in build directory.  Common generated files are now built from dependencies instead of when cmake is run (vktrace files are still built at cmake time -- this is next).


mark-lunarg:
I'm looking at the failures, it had passed on our farm.

mark-lunarg:
OK, fix force pushed.

Issue #68, opened 2016-07-26
(Feature) Consider an option to "trim" the trace file


ardaLunarG:
Trace files can be very large. There should be a way to collect a trace for only a portion of the entire execution, because the user may be interested only in a few frames.

eero-t:
For example, DOTA2 trace size is >20GB even before game play has started (whereas full DOTA2 game install itself is "just" ~16GB).

plohrmann:
We've been working on this feature and hope to get it pushed publicly soon!

eero-t:
Thanks, sounds great!

Btw. I noticed that current traces compress pretty well, e.g. 20GB DOTA2 trace is only 1GB when Gzipped.  Maybe you could consider adding support for different compression methods to traces?  E.g. LZO (lzop & liblzo) is fast both in compression & decompression (its compression ratio isn't as good as with slower compressors).

Or if you don't want to do compression during tracing, what about having at least vkreplay supporting (separately) compressed trace files?

ardaLunarG:
Fixed in 7bb45cac716

eero-t:
There are few typos in the indicated commit:
- paltform -> platform
- Vulakn -> Vulkan

Btw. It would nice to have also documentation / help output for how one actually enables trimming.

ardaLunarG:
I will address the typos.

For documentation I have filed Github Issue #186 

If it makes to SDK we will add to this to SDK documentation

PR #163, opened 2017-01-27
vktrace: Add Trim Support

plohrmann:
This addresses Feature Request #68.

This is functionality is implemented such that at trace-time you can enable and disable frame tracing on-demand. Currently this only supports being able to generate a single trace file from a single enable/disable tracing pair. Vktrace could be extended in the future to support multiple trims from a single run of the application, but it requires some restructuring to the communication between vktrace and the vktrace_layer and how vktrace recognizes that a new trace file should be created and / or saved. It was beyond the immediate scope of this feature.

Enable this feature by setting environment variable "VKTRACE_TRIM_TRIGGER=<trigger>"
<trigger> can be of the format:
"frames-<first frame>,<num frames>"
"frames-<first frame>-<last frame>"
"hotkey-<key>"

<key> can be F1 through F12, or Tab, ControlLeft, ControlRight on Linux, or TAB, CONTROL on Windows

There will be a pause in the gameplay when trim becomes enabled while it stores a snapshot of all the objects and buffer contents, and another pause when trim is disabled and all the necessary trace packets are written to the trace file.

While the trim is enabled and recording trace packets, memory consumption will climb because the trace packets are stored in memory for post-processing to eliminate some unneeded objects. We have alternatives to this but as they are considered optimizations we decided to leave them until another time. Because of the increasing memory consumption, it is recommended that trims not last longer than 1-2 minutes, but that's highly dependent on your system and the application you are recording.


ardaLunarG:
Thanks for the PR Peter,

Linux compile failed due to missing library related to hotkey I guess.
For ubuntu installing the following fixed the issue and compiled fine:
libxcb-keysyms1 - utility libraries for X C Binding -- keysyms
libxcb-keysyms1-dev - utility libraries for X C Binding -- keysyms, development files
After installing compiled went through.

We might have to add these to readme or cmake.
Anyway, this is just as a note for us to remind this dependency.

I will continue to testing.

vktrace/vktrace_viewer/vktraceviewer_qtracefileloader.cpp
```diff
@@ -217,7 +218,18 @@ bool vktraceviewer_QTraceFileLoader::populate_trace_file_info(vktraceviewer_trac
         pTraceFileInfo->packetCount++;
         fileOffset += packetSize;
 
-        fseek(pTraceFileInfo->pFile, fileOffset, SEEK_SET);
+        seekResult = fseek(pTraceFileInfo->pFile, packetSize - sizeof(uint64_t), SEEK_CUR);
```
ardacoskunses: Are vktraceviewer changes part of the trim feature?
Cant we have trim w/o these changes?

ardaLunarG: Are these vktraceviewer changes related to trim feature?
Cant we have trim support w/o these changes?

plohrmann: We don't need these changes as part of trim, but this change did fix issues that were preventing some trace files from loading in vktraceviewer. I found the issue while testing trim, but it shouldn't be specific to trim.

vktrace/vktrace_layer/vktrace_lib_trim_statetracker.cpp
```diff
@@ -0,0 +1,1772 @@
...
+#define COPY_PACKET(packet) packet = copy_packet(packet)
+
+vktrace_trace_packet_header *copy_packet(vktrace_trace_packet_header *pHeader) {
+    if (pHeader == nullptr) {
+        return nullptr;
+    }
+
+    uint64_t packetSize = pHeader->size;
+    vktrace_trace_packet_header *pCopy =
+        static_cast<vktrace_trace_packet_header *>(malloc(packetSize));
```
ardaLunarG: NULL check would be nice

vktrace/vktrace_layer/vktrace_lib_trim_statetracker.cpp
```diff
@@ -0,0 +1,1772 @@
...
+        }
+        versions.clear();
+    }
+    m_renderPassVersions.clear();
+}
+
+//-------------------------------------------------------------------------
+void StateTracker::add_RenderPassCreateInfo(
+    VkRenderPass renderPass, const VkRenderPassCreateInfo *pCreateInfo) {
+    VkRenderPassCreateInfo *pCopyCreateInfo = new VkRenderPassCreateInfo();
```
ardaLunarG: Would it be safe to add NULL checks to this function all over since multiple allocations involved?

vktrace/vktrace_layer/vktrace_lib_trace.cpp
```diff
@@ -396,18 +423,18 @@ VKTRACER_EXPORT VKAPI_ATTR VkResult VKAPI_CALL __HOOKED_vkMapMemory(
...
     CREATE_TRACE_PACKET(vkMapMemory, sizeof(void*));
     result = mdd(device)->devTable.MapMemory(device, memory, offset, size, flags, ppData);
     vktrace_set_packet_entrypoint_end_time(pHeader);
     entry = find_mem_info_entry(memory);
 
-    // For vktrace usage, clamp the memory size to the total size less offset if VK_WHOLE_SIZE is specified.
+    // For vktrace usage, clamp the memory size to the total size if VK_WHOLE_SIZE is specified.
     if (size == VK_WHOLE_SIZE) {
-        size = entry->totalSize - offset;
+        size = entry->totalSize;
```
davidlunarg: I think entry>totalSize - offset is needed. Otherwise we would map the whole memory object, rather than the memory from offset to the end.

vktrace/vktrace_common/vktrace_trace_packet_utils.c
```diff
@@ -42,7 +42,32 @@
...
+}
+
+uint64_t vktrace_get_unique_packet_index()
+{
+    // Keep the s_packet_index scope to within this method, to ensure this method is always used to get a unique packet index.
+    static uint64_t s_packet_index = 0;
+
+    vktrace_enter_critical_section(&s_packet_index_lock);
+    // Store the value then increment within the critical section to avoid race conditions.
+    int index = s_packet_index;
```
plohrmann: Good catch!

vktrace/vktrace_layer/vktrace_lib_trace.cpp
plohrmann: Agreed, now that I know that "VK_WHOLE_SIZE" is only for the remainder of the buffer size.

vktrace/vktrace_layer/vktrace_lib_trace.cpp
```diff
@@ -1063,7 +1362,52 @@ VKTRACER_EXPORT VKAPI_ATTR VkResult VKAPI_CALL __HOOKED_vkCreateRenderPass(
...
+        trim::add_RenderPassCreateInfo(*pRenderPass, pCreateInfo);
+        if (pInfo != NULL)
+        {
+            pInfo->belongsToDevice = device;
+            pInfo->ObjectInfo.RenderPass.pCreatePacket = trim::copy_packet(pHeader);
+            
+            if (pCreateInfo->attachmentCount > 0)
+            {
+                pInfo->ObjectInfo.RenderPass.attachmentCount = pCreateInfo->attachmentCount;
+                pInfo->ObjectInfo.RenderPass.pAttachments = new trim::ImageTransition[pCreateInfo->attachmentCount];
```
plohrmann: will do

vktrace/vktrace_layer/vktrace_lib_trace.cpp
```diff
@@ -880,7 +1110,35 @@ VKTRACER_EXPORT VKAPI_ATTR VkResult VKAPI_CALL __HOOKED_vkCreateFramebuffer(
...
+        FINISH_TRACE_PACKET();
+    }
+    else
+    {
+        vktrace_finalize_trace_packet(pHeader);
+        trim::ObjectInfo* pInfo = trim::add_Framebuffer_object(*pFramebuffer);
+        pInfo->belongsToDevice = device;
+        pInfo->ObjectInfo.Framebuffer.pCreatePacket = trim::copy_packet(pHeader);
+        pInfo->ObjectInfo.Framebuffer.attachmentCount = pCreateInfo->attachmentCount;
+        pInfo->ObjectInfo.Framebuffer.pAttachments = new VkImageView[pCreateInfo->attachmentCount];
```
plohrmann: Will do.

vktrace/vktrace_layer/vktrace_lib_trace.cpp
```diff
@@ -1751,7 +2685,47 @@ VKTRACER_EXPORT VKAPI_ATTR void VKAPI_CALL __HOOKED_vkCmdBeginRenderPass(
...
+        {
+            pCommandBuffer->ObjectInfo.CommandBuffer.activeRenderPass = pRenderPassBegin->renderPass;
+
+            trim::ObjectInfo* pFramebuffer = trim::get_Framebuffer_objectInfo(pRenderPassBegin->framebuffer);
+            trim::ObjectInfo* pRenderPass = trim::get_RenderPass_objectInfo(pRenderPassBegin->renderPass);
+            assert(pRenderPass != nullptr && pFramebuffer != nullptr);
+            if (pRenderPass != nullptr && pFramebuffer != nullptr)
+            {
+                assert(pRenderPass->ObjectInfo.RenderPass.attachmentCount == pFramebuffer->ObjectInfo.Framebuffer.attachmentCount);
+                for (uint32_t i = 0; i < pRenderPass->ObjectInfo.RenderPass.attachmentCount && i < pFramebuffer->ObjectInfo.Framebuffer.attachmentCount; i++)
```
plohrmann: yes. Will do.

vktrace/vktrace_layer/vktrace_lib_trim_statetracker.cpp
plohrmann: will do

vktrace/vktrace_viewer/vktraceviewer_qtracefileloader.cpp
plohrmann: Ah yes, this change was to fix opening really large files. Instead of seeking from the start of the file every time, this change will just seek relative to the current position, so we advance the pointer by the size of the packet minus the sizeof(uint64_t) which holds packetSize (which we've already read in).

ardaLunarG:
Merged

PR #164, opened 2017-01-27
vktrace: modify help text, fix frame range parsing error

mizhen:
modify help text, fix frame range parsing error

modify help text of screenshot command for vktrace/vkreplay, fix frame range parsing error make it output error message for extra input.

Change-Id: I20fe77eae7c33a26f5354079d03ad0f9971d5ff9

### DIFF
--- a/layersvt/screenshot_parsing.cpp
+++ b/layersvt/screenshot_parsing.cpp
@@ -20,6 +20,34 @@ using namespace std;
 
 namespace screenshot {
 
+// check rangeString:
+//    1> rangeString is valid and every char is '0'-'9' or '-'
+//    2> char '-' count <= 2;
+// return:
+//    false if 1> or 2> fail;
+bool checkScreenShotFrameRangeString(const char *rangeString) {
+    bool checkResult = false;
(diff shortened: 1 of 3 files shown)


davidlunarg:
The method checkScreenShotFrameRangeString returns true if the string passed in is "2--2". It should return false.

When the range specified with -s for vktrace/replay has a bad format, the error message should include the help text. Also, the error message describing the error in the range doesn't need to be so verbose. Attached is a patch file with my proposed changes. If my changes are used, the method initScreenShotFrameRange could be simplified to return true or false, rather than a specific error number.

[p.patch.zip](https://github.com/LunarG/VulkanTools/files/736392/p.patch.zip)

mizhen:
The function is used to check the following: rangeString is valid, every char is '0'-'9' or '-', char '-' count <= 2. It should return true. initScreenShotFrameRange function can handle errors like frameCount < 0. 

The patch file looks good. Thanks, David.

davidlunarg:
Could you resubmit with my patch? I think we would then be good to go.

ardaLunarG:
As Ming approved David's changes I push both of them to master.

PR #166, opened 2017-01-31
vkreplay: Fix error in vkUpdateDescriptorSets

mizhen:
Add descriptorType handling when call VKTRACE_DELETE.

Change-Id: I5505ccf666238af008271a3848580ca9839afb54


davidlunarg:
+1

PR #171, opened 2017-02-02
fix error for copy app memory

mizhen:
Add additional process in vkFreeMemory in case target app doesn't call vkUnmapMemory. 

There are some apps call vkFreeMemory without call vkUnmapMemory on same memory object. In that situation, capture/playback runs into error.


mizhen:
yes, only second commit, thanks.

davidlunarg:
Second commit: +1

ardaLunarG:
Pushed to master

PR #172, opened 2017-02-07
via: Fix installation paths that contain spaces

joey-lunarg:
Put quotes around path when passing as arg to other programs.

Change-Id: Ia2bc070ea0ca1dacb6c4de3681181ef79854d479


ardaLunarG:
Merged

PR #175, opened 2017-02-08
vktraceviewer improvements

ardaLunarG:


ardaLunarG:
I have added a new patch on top which removes warnings (there are a lot of them)
However 32-bit build still fails due to %p.

Added Cody as reviewer

cnorthrop:
I pushed a fix for this on cody_fix_VktraceViewer that gets Android compiling again.  It is messy, generally needs to be refactored to call snprintf multiple times rather than one giant string.  Avoiding %p in the future will help cross platform portability.

ardaLunarG:
merged

PR #178, opened 2017-02-10
vktrace: Fix a 4 bytes alignment issue

jyangsh:
Both __HOOKED_vkFlushMappedMemoryRanges() and
vkFlushMappedMemoryRangesWithoutAPICall call
getALLChangedPackageSizeInMappedMemory(), but
only __HOOKED_vkFlushMappedMemoryRange rounds
the size to 4 bytes alignment.

Move the ROUNDUP_TO_4() into the function to
avoid calling it in two places.


davidlunarg:
+1

ardaLunarG:
merged

PR #179, opened 2017-02-10
vktrace: fix corruption and flickering problem of xcap-507.

mizhen:
page guard process miss following write access if read access happen on same page, this cause show corruption and flickering during capture in some titles. the change fix the problem.

XCAP-507

Change-Id: Icde4628641b98600072a85418a7824ccec55fa62

### DIFF
--- a/vktrace/vktrace_layer/vktrace_lib_pageguard.cpp
+++ b/vktrace/vktrace_layer/vktrace_lib_pageguard.cpp
@@ -108,17 +108,17 @@ bool getPageGuardEnableFlag()
     return EnablePageGuard;
 }
 
-bool getEnableReadPMBFlag()
-{
+bool getEnableReadProcessFlag(const char* name) {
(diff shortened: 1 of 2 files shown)


davidlunarg:
The change looks good, except that I'm concerned we're adding environment variables that change the behavior of vktrace, but those variables are not getting documented. The only way someone will find out about them is by reading the code. We need to decide how this (and existing) environment variables are to be documented. I see several ways that they could be documented:

- Additional command line option to vktrace
- A file (format TBD), that can be pulled in by LunarG's SDK build so that the description of environment variables is included in the Vulkan SDK.
- A file somewhere in the source tree describing the env variables
- No documentation whatsoever. This would be for variables typically used only by developers that might be used to control something such as debug output.

Could you propose what type of documentation would be needed for this new env variable?

ardaLunarG:
merged

Issue #181, opened 2017-02-11
Trim function does not work if frame range outside the captured frames


ardaLunarG:
vktrace -tr frames-0-99 -p cube -a "--c 100"                
vkreplay -o vktrace_out.vktrace ==>WORKS
vktrace -tr frames-0-100 -p cube -a "--c 100"
vkreplay -o vktrace_out.vktrace ==> DOES NOT WORK
vktrace -tr frames-1-120 -p cube -a "--c 100"
vkreplay -o vktrace_out.vktrace ==> DOES NOT WORK

ardaLunarG:
fixed

PR #190, opened 2017-02-14
vktrace trim: Fix tracking of object destruction

plohrmann:
While investigating #181 I came across a few objects whose destruction was not being tracked in the trim StateTracker. This change fixes the following objects:
VkCommandPool
VkPipelineCache
VkSwapchainKHR
VkSurfaceKHR
VkQueue


vktrace/vktrace_layer/vktrace_lib_trim.cpp
```diff
@@ -1579,6 +1579,20 @@ ObjectInfo &add_Device_object(VkDevice var) {
...
+
+    std::vector<VkQueue> queuesToRemove;
+    for (TrimObjectInfoMap::iterator info = s_trimGlobalStateTracker.createdQueues.begin(); info != s_trimGlobalStateTracker.createdQueues.end(); ++info) {
+        if (info->second.belongsToDevice == var)
+        {
+            queuesToRemove.push_back(static_cast<VkQueue>(info->first));
+        }
+    }
+
+    for (int i = 0; i < queuesToRemove.size(); i++)
```
ardaLunarG: size() returns size_t, 
Should not we use size_t instead of int here?

ardaLunarG:
merged

PR #192, opened 2017-02-16
32 bit trim compile

ardaLunarG:
Commit  	a0b6c65 does not belong to this PR. It is already merged and there is no change in that.
Please review other 4 patches.


ardaLunarG:
Merged

PR #195, opened 2017-02-20
Trim size

ardaLunarG:
Size's in trim was hitting to assert&3 when adding map and unmap packages.
This patch round them to 4.

Also add additional size check for map.


ardaLunarG:
merged

PR #196, opened 2017-02-22
vktrace: Set more variables only used in asserts

ardaLunarG:
U_ASSERT_ONLY moved common area and more variables set


ardaLunarG:
merged

PR #197, opened 2017-02-23
vktrace: Better way to skip vktrace build without xcb

Tony-LunarG:
Use BUILD_VKTRACE_REPLAY option to disable building vkreplay instead of heavy-handedly doing it later

Change-Id: I4a975f706e0dba67be2f78ef43266474f77fb654


ardaLunarG:
merged

PR #198, opened 2017-02-24
vktrace: Fixes to get macOS building again

cnorthrop:


vktrace/vktrace_common/vktrace_interconnect.c
```diff
@@ -125,7 +125,7 @@ BOOL vktrace_MessageStream_SetupHostSocket(MessageStream* pStream) {
 #if defined(PLATFORM_LINUX) || defined(PLATFORM_OSX)
     int yes = 1;
 #endif
-    struct addrinfo hostAddrInfo = {0};
+    struct addrinfo hostAddrInfo = {};
```
ardaLunarG: {0} was meant to clear up the struct members. Should we call memset 0 to stay on the safe side?

cnorthrop: This popped up as a warning:

>    .../VulkanTools/vktrace/vktrace_common/vktrace_interconnect.c:128:38: warning: missing field 'ai_family' initializer [-Wmissing-field-initializers]
       struct addrinfo hostAddrInfo = {0};
                                        ^
   1 warning generated.

But I see that -Wno-missing-field-initializers is specified in CMakeLists.txt, so this warning shouldn't be firing.  I'll resubmit without that portion and look into that separately.

ardaLunarG:
merged

PR #199, opened 2017-02-24
Vkreplay and vktraceviewer merged into singe folder

ardaLunarG:
Vkreplay and vktraceviewer merged into single folder.


ardaLunarG:
merged

PR #200, opened 2017-02-26
vktrace: Removing build_options.cmake

ardaLunarG:
Removing seperate cmake file and include content into regular cmake


ardaLunarG:
merged

PR #201, opened 2017-02-28
Screen capture color space format

ardaLunarG:
With this patch set screen capture color space format will be same as swapchain.
User can specify specific color space format.

### DIFF
--- a/layersvt/screenshot.cpp
+++ b/layersvt/screenshot.cpp
@@ -417,6 +417,12 @@ static void writePPM(const char *filename, VkImage image1) {
                 destformat = VK_FORMAT_R8G8B8A8_SINT;
             else
                 destformat = VK_FORMAT_R8G8B8_SINT;
+        } else {
+#ifdef ANDROID
+#else
+            fprintf(stderr, "Selected format:%s\nIs NOT in the list:\nUNORM, SNORM, USCALED, SSCALED, UINT, SINT, SRGB\n"
+                            "Swapchain Colorspace will be used instead\n", vk_screenshot_format);
+#endif
         }
     }
     local_free_getenv(vk_screenshot_format);
@@ -468,6 +474,11 @@ static void writePPM(const char *filename, VkImage image1) {
     //Still could not find the right format then we use UNORM
     if (destformat == VK_FORMAT_UNDEFINED)
     {
+#ifdef ANDROID
+#else
+        fprintf(stderr, "Swapchain format is not in the list:\nUNORM, SNORM, USCALED, SSCALED, UINT, SINT, SRGB\n"
+                        "UNORM colorspace will be used instead\n");
+#endif
         if (numChannels == 4)
             destformat = VK_FORMAT_R8G8B8A8_UNORM;
         else
@@ -808,6 +819,8 @@ static void writePPM(const char *filename, VkImage image1) {
 #ifdef ANDROID
         __android_log_print(ANDROID_LOG_DEBUG, "screenshot",
                             "Failed to open output file: %s.  Be sure to grant read and write permissions.", filename);
+#else
+        fprintf(stderr, "Failed to open output file:%s,  Be sure to grant read and write permissions\n", filename);
 #endif
         return;
     }
@@ -1127,6 +1140,7 @@ VKAPI_ATTR VkResult VKAPI_CALL QueuePresentKHR(VkQueue queue, const VkPresentInf
             fileName = base + ".ppm";
 #else
             fileName = to_string(frameNumber) + ".ppm";
+            printf("Screen Capture file is: %s \n", fileName.c_str());
 #endif
 
             VkImage image;
--- a/layersvt/screenshot.cpp
+++ b/layersvt/screenshot.cpp
@@ -120,8 +120,20 @@ static loader_platform_thread_mutex globalLock;
 
 const char *vk_screenshot_format = nullptr;
 
-bool printFormatWarning1 = true;
-bool printFormatWarning2 = true;
+bool printFormatWarning = true;
+
+typedef enum colorSpaceFormat {
+    UNDEFINED = 0,
+    UNORM = 1,
+    SNORM = 2,
+    USCALED = 3,
+    SSCALED = 4,
+    UINT = 5,
+    SINT = 6,
+    SRGB = 7
+} colorSpaceFormat;
+
+colorSpaceFormat userColorSpaceFormat = UNDEFINED;
 
 // unordered map: associates a swap chain with a device, image extent, format,
 // and list of images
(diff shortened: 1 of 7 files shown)


layersvt/screenshot.cpp
```diff
@@ -364,18 +366,126 @@ static void writePPM(const char *filename, VkImage image1) {
...
+                destformat = VK_FORMAT_R8G8B8_UINT;
+        } else if (!strcmp(vk_screenshot_format, "SINT")) {
+            if (numChannels == 4)
+                destformat = VK_FORMAT_R8G8B8A8_SINT;
+            else
+                destformat = VK_FORMAT_R8G8B8_SINT;
+        } else {
+#ifdef ANDROID
+#else
+            fprintf(stderr, "Selected format:%s\nIs NOT in the list:\nUNORM, SNORM, USCALED, SSCALED, UINT, SINT, SRGB\n"
```
davidlunarg: This message should be printed only once. If multiple screenshots are done, it looks like it will be printed for each screenshot.

layersvt/screenshot.cpp
```diff
@@ -364,18 +366,126 @@ static void writePPM(const char *filename, VkImage image1) {
...
         assert(0);
         return;
     }
-    if ((3 != numChannels) && (4 != numChannels)) {
+
+    // Initial dest format is undefined as we will look for one
+    VkFormat destformat = VK_FORMAT_UNDEFINED;
+
+    //Get users request is specific color space format required
+    const char *vk_screenshot_format = local_getenv(env_var_format);
```
davidlunarg: Parse the env var once, sometime during initiliazation, rather than each time a screenshot is generated.

layersvt/screenshot.cpp
```diff
@@ -92,6 +93,7 @@ static inline void local_free_getenv(const char *val) {}
 
 const char *env_var_old = "_VK_SCREENSHOT";
 const char *env_var = "VK_SCREENSHOT_FRAMES";
+const char *env_var_format = "VK_SCREENSHOT_FORMAT";
```
davidlunarg: These three lines are duplicated. Can #ifdef's be rearranged so that they occur only once?

layersvt/screenshot.cpp
```diff
@@ -364,18 +366,126 @@ static void writePPM(const char *filename, VkImage image1) {
...
+                destformat = VK_FORMAT_R8G8B8_SINT;
+        }
+    }
+
+    //Still could not find the right format then we use UNORM
+    if (destformat == VK_FORMAT_UNDEFINED)
+    {
+#ifdef ANDROID
+#else
+        fprintf(stderr, "Swapchain format is not in the list:\nUNORM, SNORM, USCALED, SSCALED, UINT, SINT, SRGB\n"
```
davidlunarg: Same as above - only print it once.

layersvt/screenshot.cpp
```diff
@@ -1029,6 +1140,7 @@ VKAPI_ATTR VkResult VKAPI_CALL QueuePresentKHR(VkQueue queue, const VkPresentInf
             fileName = base + ".ppm";
 #else
             fileName = to_string(frameNumber) + ".ppm";
+            printf("Screen Capture file is: %s \n", fileName.c_str());
```
davidlunarg: This could result in lots of output. But I realize the value of printing out the name of the capture file. Not sure what best solution is.

vktrace/vktrace_replay/vkreplay_main.cpp
```diff
@@ -341,6 +348,18 @@ int vkreplay_main(int argc, char** argv, vktrace_window_handle window = 0) {
         vktrace_set_global_var("VK_SCREENSHOT_FRAMES", "");
     }
 
+    // Set up environment for screenshot color space format
+    if (replaySettings.screenshotColorFormat != NULL && replaySettings.screenshotList != NULL) {
+        vktrace_set_global_var("VK_SCREENSHOT_FORMAT", replaySettings.screenshotColorFormat);
+        vktrace_free((char*)replaySettings.screenshotColorFormat);
+    }else if (replaySettings.screenshotColorFormat != NULL && replaySettings.screenshotList == NULL) {
+        vktrace_LogError("Screenshot format should be used when screenshot enabled!");
```
davidlunarg: Maybe make this a warning: "Screenshot format ignored when screenshot is not enabled"

vktrace/vktrace_trace/vktrace.cpp
```diff
@@ -317,6 +325,18 @@ int main(int argc, char* argv[]) {
             vktrace_set_global_var("VK_SCREENSHOT_FRAMES", "");
         }
 
+        // Set up environment for screenshot color space format
+        if (g_settings.screenshotColorFormat != NULL && g_settings.screenshotList != NULL) {
+            vktrace_set_global_var("VK_SCREENSHOT_FORMAT", g_settings.screenshotColorFormat);
+            vktrace_free((char*)g_settings.screenshotColorFormat);
+        }else if (g_settings.screenshotColorFormat != NULL && g_settings.screenshotList == NULL) {
+            vktrace_LogError("Screenshot format should be used when screenshot enabled!");
```
davidlunarg: Same comment as above

layersvt/screenshot.cpp
ardaLunarG: Agree

layersvt/screenshot.cpp
ardaLunarG: Agree

layersvt/screenshot.cpp
ardaLunarG: Agree

layersvt/screenshot.cpp
ardaLunarG: Agree

layersvt/screenshot.cpp
ardaLunarG: I will address this one later with a different change: as printing the directory instead of filenames. Lets keep it as is for now

vktrace/vktrace_replay/vkreplay_main.cpp
ardaLunarG: Agree

vktrace/vktrace_trace/vktrace.cpp
ardaLunarG: Agree

ardaLunarG:
Does vktrace/vkreplay unset VK_SCREENSHOT_FRAMES and VK_SCREENSHOT_FORMAT when -s and -sf are not on the command line?
Ans: when they are not set in the command line vktrace/replay clear the environment variables.

ardaLunarG:
pushed patches addressing comments

ardaLunarG:
Thanks for testing Cody!

ardaLunarG:
merged

PR #204, opened 2017-03-07
Trivial fix for logging

jyangsh:
missing value for "%d" in the logging message.

Current logging implementation, the vktrace_LogXXX() functions, doesn't check the print format specifier with the format data type strictly. That may cause some potential logging issues.


ardaLunarG:
Merged, Thanks!

PR #208, opened 2017-03-10
vktrace: Warning on memory requirement changes

jyangsh:
Rarely driver may change the memory requirements for image or buffer. But if it happens, a trace file captured on old driver may not be replayed properly on the new driver. So we'd better have a warning on it for catching such case.


jyangsh:
@davidlunarg, no problem, I am fine to close this PR.

PR #212, opened 2017-03-16
vktrace: handle mem types for better trace file portability

davidlunarg:
Change-Id: I512ceb6aaded8ae54a48ae6baea046bfb8c4a617

### DIFF
--- a/vktrace/vktrace_common/vktrace_trace_packet_identifiers.h
+++ b/vktrace/vktrace_common/vktrace_trace_packet_identifiers.h
@@ -27,10 +27,13 @@
 #define VKTRACE_TRACE_FILE_VERSION_3 0x0003
 #define VKTRACE_TRACE_FILE_VERSION_4 0x0004
 #define VKTRACE_TRACE_FILE_VERSION_5 0x0005
-#define VKTRACE_TRACE_FILE_VERSION VKTRACE_TRACE_FILE_VERSION_5
-#define VKTRACE_TRACE_FILE_VERSION_MINIMUM_COMPATIBLE VKTRACE_TRACE_FILE_VERSION_5
+#define VKTRACE_TRACE_FILE_VERSION_6 0x0006
+#define VKTRACE_TRACE_FILE_VERSION VKTRACE_TRACE_FILE_VERSION_6
+#define VKTRACE_TRACE_FILE_VERSION_MINIMUM_COMPATIBLE VKTRACE_TRACE_FILE_VERSION_6
 
-#define VKTRACE_MAX_TRACER_ID_ARRAY_SIZE 14
+#define VKTRACE_FILE_MAGIC 0xABADD068ADEAFD0C
+
+#define VKTRACE_MAX_TRACER_ID_ARRAY_SIZE 16  // Should be multiple of 8
 
 typedef enum VKTRACE_TRACER_ID {
     VKTRACE_TID_RESERVED = 0,
@@ -62,6 +65,8 @@ static const VKTRACE_TRACER_REPLAYER_INFO gs_tracerReplayerInfo[VKTRACE_MAX_TRAC
     {VKTRACE_TID_RESERVED, FALSE, "", ""},  // this can be updated as new tracers are added
     {VKTRACE_TID_RESERVED, FALSE, "", ""},  // this can be updated as new tracers are added
     {VKTRACE_TID_RESERVED, FALSE, "", ""},  // this can be updated as new tracers are added
+    {VKTRACE_TID_RESERVED, FALSE, "", ""},  // this can be updated as new tracers are added
+    {VKTRACE_TID_RESERVED, FALSE, "", ""},  // this can be updated as new tracers are added
 };
 
 typedef enum _VKTRACE_TRACE_PACKET_ID {
@@ -71,33 +76,51 @@ typedef enum _VKTRACE_TRACE_PACKET_ID {
     VKTRACE_TPI_MARKER_API_GROUP_BEGIN,
     VKTRACE_TPI_MARKER_API_GROUP_END,
     VKTRACE_TPI_MARKER_TERMINATE_PROCESS,
+    VKTRACE_TPI_PORTABILITY_TABLE,
     VKTRACE_TPI_BEGIN_API_HERE  // this enum should always be the last in the list. Feel free to insert new ID above this one.
 } VKTRACE_TRACE_PACKET_ID;
+
+#define VKTRACE_BIG_ENDIAN 0
+#define VKTRACE_LITTLE_ENDIAN 1
 
 typedef struct {
     uint8_t id;
     uint8_t is_64_bit;
 } vktrace_tracer_info;
 
-#if defined(__LP64__) || defined(_WIN64) || (defined(__x86_64__) && !defined(__ILP32__)) || defined(_M_X64) || defined(__ia64) || \
-    defined(_M_IA64) || defined(__aarch64__) || defined(__powerpc64__)
+// vktrace_trace_file_header and vktrace_trace_packet_header are written and read
+// by both host and target CPU, which can be different architectures.
+// All fields in these structs are 64-bit aligned.
 
 typedef struct {
     uint16_t trace_file_version;
+    uint16_t reserved1;
+    uint16_t reserved2;
+    uint16_t reserved3;
+    uint64_t magic;
     uint32_t uuid[4];
     uint64_t first_packet_offset;  // will be size of header including size of tracer_id_array and state_snapshot_path/binary
-    uint8_t tracer_count;          // number of tracers referenced in this trace file
+    uint64_t tracer_count;         // number of tracers referenced in this trace file
     vktrace_tracer_info
         tracer_id_array[VKTRACE_MAX_TRACER_ID_ARRAY_SIZE];  // array of tracer_ids and values which are referenced in the trace file
     uint64_t trace_start_time;
+    uint64_t portability_table_valid;
+
+    // Description of platform on which this trace was created
+    uint64_t endianess;
+    uint64_t ptrsize;
+    uint64_t arch;
+    uint64_t os;
+    uint64_t gpu;
+    uint64_t drv_vers;
 } vktrace_trace_file_header;
 
 typedef struct {
     uint64_t size;  // total size, including extra data, needed to get to the next packet_header
     uint64_t global_packet_index;
-    uint8_t tracer_id;   // TODO: need to uniquely identify tracers in a way that is known by the replayer
-    uint16_t packet_id;  // VKTRACE_TRACE_PACKET_ID (or one of the api-specific IDs)
-    uint32_t thread_id;
+    uint64_t tracer_id;  // TODO: need to uniquely identify tracers in a way that is known by the replayer
+    uint64_t packet_id;  // VKTRACE_TRACE_PACKET_ID (or one of the api-specific IDs)
+    uint64_t thread_id;
     uint64_t vktrace_begin_time;  // start of measuring vktrace's overhead related to this packet
     uint64_t entrypoint_begin_time;
     uint64_t entrypoint_end_time;
@@ -105,46 +128,6 @@ typedef struct {
     uint64_t next_buffers_offset;  // used for tracking the addition of buffers to the trace packet
     uintptr_t pBody;               // points to the body of the packet
 } vktrace_trace_packet_header;
-
-#else
-
-// vktrace_trace_file_header and vktrace_trace_packet_header are written and
-// read by both host and target CPU, which can be different architectures.
-// Specifically for Android, x86 and arm32 align 64-bit integers differently,
-// so we must require 8 byte alignment.
-// These changes are guarded by a 64-bit check above for now.
-
-#if defined(_WIN32) || defined(_WIN64)
-#define ALIGN8 __declspec(align(8))
-#else
-#define ALIGN8 __attribute__((aligned(8)))
-#endif
-
-typedef struct {
-    uint16_t trace_file_version;
-    uint32_t uuid[4];
-    ALIGN8 uint64_t first_packet_offset;  // will be size of header including size of tracer_id_array and state_snapshot_path/binary
-    uint8_t tracer_count;                 // number of tracers referenced in this trace file
-    vktrace_tracer_info
-        tracer_id_array[VKTRACE_MAX_TRACER_ID_ARRAY_SIZE];  // array of tracer_ids and values which are referenced in the trace file
-    ALIGN8 uint64_t trace_start_time;
-} vktrace_trace_file_header;
-
-typedef struct {
-    ALIGN8 uint64_t size;  // total size, including extra data, needed to get to the next packet_header
-    ALIGN8 uint64_t global_packet_index;
-    uint8_t tracer_id;   // TODO: need to uniquely identify tracers in a way that is known by the replayer
-    uint16_t packet_id;  // VKTRACE_TRACE_PACKET_ID (or one of the api-specific IDs)
-    uint32_t thread_id;
-    ALIGN8 uint64_t vktrace_begin_time;  // start of measuring vktrace's overhead related to this packet
-    ALIGN8 uint64_t entrypoint_begin_time;
-    ALIGN8 uint64_t entrypoint_end_time;
-    ALIGN8 uint64_t vktrace_end_time;     // end of measuring vktrace's overhead related to this packet
-    ALIGN8 uint64_t next_buffers_offset;  // used for tracking the addition of buffers to the trace packet
-    uintptr_t pBody;                      // points to the body of the packet
-} vktrace_trace_packet_header;
-
-#endif  // 64-bit
 
 typedef struct {
     vktrace_trace_packet_header* pHeader;
--- a/vktrace/vktrace_common/vktrace_trace_packet_identifiers.h
+++ b/vktrace/vktrace_common/vktrace_trace_packet_identifiers.h
@@ -92,41 +92,55 @@ typedef struct {
 // by both host and target CPU, which can be different architectures.
 // All fields in these structs are 64-bit aligned.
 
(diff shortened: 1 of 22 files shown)


vktrace/vktrace_common/vktrace_trace_packet_identifiers.h
```diff
@@ -71,33 +76,51 @@ typedef enum _VKTRACE_TRACE_PACKET_ID {
...
 
 typedef struct {
     uint64_t size;  // total size, including extra data, needed to get to the next packet_header
     uint64_t global_packet_index;
-    uint8_t tracer_id;   // TODO: need to uniquely identify tracers in a way that is known by the replayer
-    uint16_t packet_id;  // VKTRACE_TRACE_PACKET_ID (or one of the api-specific IDs)
-    uint32_t thread_id;
+    uint64_t tracer_id;  // TODO: need to uniquely identify tracers in a way that is known by the replayer
+    uint64_t packet_id;  // VKTRACE_TRACE_PACKET_ID (or one of the api-specific IDs)
+    uint64_t thread_id;
```
jyangsh: Does this change break the backward compatibility? The trace file captured with the old tracer may not be retraced with the new vkreplayer.

vktrace/vktrace_replay/vkreplay_main.cpp
```diff
@@ -404,8 +423,26 @@ int vkreplay_main(int argc, char** argv, vktrace_window_handle window = 0) {
...
             "older replayer.",
             fileHeader.trace_file_version, VKTRACE_TRACE_FILE_VERSION_MINIMUM_COMPATIBLE);
+        fclose(tracefp);
+        vktrace_free(pTraceFile);
+        vktrace_free(traceFile);
+        return -1;
+    }
+
+    // Make sure magic number in trace file is valid
+    if (fileHeader.magic != VKTRACE_FILE_MAGIC) {
```
jyangsh: How about the old trace file without the magic number?

vktrace/vktrace_replay/vkreplay_vkreplay.cpp
```diff
@@ -2256,16 +2112,182 @@ VkResult vkReplay::manually_replay_vkAllocateMemory(packet_vkAllocateMemory *pPa
         }
     }
 
-    if (!m_objMapper.m_adjustForGPU) {
-        uint32_t replayIdx;
-        if (getMemoryTypeIdx(pPacket->device, remappedDevice, pPacket->pAllocateInfo->memoryTypeIndex, &replayIdx)) {
-            *((uint32_t *)&pPacket->pAllocateInfo->memoryTypeIndex) = replayIdx;
+    if ((uint64_t)pPacket->pMemory == (uint64_t)0x22a5caeb580) {
+        pPacket->pMemory = pPacket->pMemory;
```
jyangsh: Not very understand what is this code block for? 0x22a5caeb580 is another magic number?

vktrace/vktrace_common/vktrace_trace_packet_utils.c
```diff
@@ -113,16 +114,163 @@ uint64_t vktrace_get_time() {
...
+    VkInstance instance;
+    uint32_t physDevCount;
+    VkPhysicalDevice physDevice;
+
+    static bool devPropertiesValid = false;
+    static VkPhysicalDeviceProperties devProperties;
+
+    if (devPropertiesValid) {
+        rval = &devProperties;
+        goto allDone;
```
ardaLunarG: this goto allDone doing nothing but returns, we can return right here as well and use less goto.

vktrace/vktrace_common/vktrace_trace_packet_utils.c
```diff
@@ -113,16 +114,163 @@ uint64_t vktrace_get_time() {
...
+    // We can't just save the pointer to the value, because getenv returns
+    // a pointer to where the value is stored, but we are going to change it.
+    char* pDL = vktrace_get_global_var("VK_DEVICE_LAYERS");
+    char* pIL = vktrace_get_global_var("VK_INSTANCE_LAYERS");
+    char *pSaveDL = NULL, *pSaveIL = NULL;
+    char* pEnvSave;
+    int lenDL = pDL ? strlen(pDL) : 0;
+    int lenIL = pIL ? strlen(pIL) : 0;
+    pEnvSave = vktrace_malloc(lenDL + lenIL + 2);
+    if (!pEnvSave) goto allDone;
```
ardaLunarG: same as above. we can return here

vktrace/vktrace_common/vktrace_trace_packet_utils.c
```diff
@@ -113,16 +114,163 @@ uint64_t vktrace_get_time() {
...
+    // We don't want any layers for this call to vKCreateInstance,
+    // especially the trace layer!
+    if (pDL) vktrace_set_global_var("VK_DEVICE_LAYERS", "");
+    if (pIL) vktrace_set_global_var("VK_INSTANCE_LAYERS", "");
+
+    memset(&createInfo, 0, sizeof(createInfo));
+    createInfo.sType = VK_STRUCTURE_TYPE_INSTANCE_CREATE_INFO;
+    if (VK_SUCCESS != vkCreateInstance(&createInfo, NULL, &instance)) goto restoreEnv;
+
+    if (VK_SUCCESS != vkEnumeratePhysicalDevices(instance, &physDevCount, NULL) || physDevCount != 1) {
```
ardaLunarG: I now we do not handle multigpu cases well yet but if here we saying if multiple gpus registered on one device do NOT run vktrace and replay. There a lot of devices in this case. We need to relax this condition.

vktrace/vktrace_common/vktrace_trace_packet_utils.c
```diff
@@ -113,16 +114,163 @@ uint64_t vktrace_get_time() {
...
+
+destroyInstance:
+    vkDestroyInstance(instance, NULL);
+
+restoreEnv:
+    if (pSaveDL) vktrace_set_global_var("VK_DEVICE_LAYERS", pSaveDL);
+    if (pSaveIL) vktrace_set_global_var("VK_INSTANCE_LAYERS", pSaveIL);
+    vktrace_free(pEnvSave);
+
+allDone:
```
ardaLunarG: this mark is not necessary as only returns after

vktrace/vktrace_common/vktrace_trace_packet_identifiers.h
davidlunarg: Yes, this change does break compatibility. And we don't take it lightly. We needed to add configuration information to the header, which is needed not only for this change, but will likely be required for future enhancements. A workaround  for moving forward old traces is to replay the old trace file with a compatible version of vkreplay,  and tracing the replay with a new vktrace:

`vktrace.new -p vkreplay.old -a '-t tracefile.old.vktrace' -o tracefile.new.vktrace`

vktrace/vktrace_replay/vkreplay_vkreplay.cpp
```diff
@@ -2237,9 +2065,37 @@ bool vkReplay::getMemoryTypeIdx(VkDevice traceDevice, VkDevice replayDevice, uin
...
     gpuMemObj local_mem;
+    VkMemoryRequirements memRequirements;
+    VkDeviceSize gimrSizeSum = 0;
+    VkDeviceSize alignment = 1;
+    uint32_t replayMemTypeIndex;
+    vktrace_trace_packet_header packetHeader1, packetHeader2;
+    VkDeviceMemory traceAllocateMemoryRval;
+    packet_vkBindImageMemory bimPacket;  // Note: we rely on the fact that packet_vkBindBufferMemory is the same size.....
+    packet_vkGetImageMemoryRequirements
+        gimrPacket;  // Note: we rely on the fact that packet_vkGetBufferMemoryRequires is the same size.....
```
ardaLunarG: this line would fit into one line no?

vktrace/vktrace_replay/vkreplay_vkreplay.cpp
```diff
@@ -2256,16 +2112,182 @@ VkResult vkReplay::manually_replay_vkAllocateMemory(packet_vkAllocateMemory *pPa
...
+
+            if ((packetHeader1.packet_id == VKTRACE_TPI_VK_vkBindImageMemory ||
+                 packetHeader1.packet_id == VKTRACE_TPI_VK_vkBindBufferMemory) &&
+                traceAllocateMemoryRval == bimPacket.memory) {
+                // A vkBIM/vkBBM binds memory allocated by this vkAM call.
+                foundBindMem = true;
+
+                // Search backwards for the vkGIMR/vkGBMR call.
+                int64_t j;
+                for (j = amIdx - 1; j >= 0; j--) {
```
ardaLunarG: This for compares int64_t to size_t, (signed vs unsigned). Be better define j uint64_t

vktrace/vktrace_replay/vkreplay_vkreplay.cpp
```diff
@@ -2256,16 +2112,182 @@ VkResult vkReplay::manually_replay_vkAllocateMemory(packet_vkAllocateMemory *pPa
...
+                    if (packetHeader1.packet_id == VKTRACE_TPI_VK_vkDestroyImage ||
+                        packetHeader1.packet_id == VKTRACE_TPI_VK_vkDestroyBuffer) {
+                        FREAD(&destroyImagePacket, sizeof(destroyImagePacket), 1, tracefp);
+                        if (destroyImagePacket.image == bimPacket.image) {
+                            // Found a destroy of this Buffer/Image, stop the back search.
+                            break;
+                        }
+                    }
+                }
+                if (j < 0) {
```
ardaLunarG: With above comment (uint64_t j) this would not work. We must change logic here.

vktrace/vktrace_replay/vkreplay_vkreplay.cpp
davidlunarg: That is debug code that was not removed. I used it to set a breakpoint when a certain vkAllocateMemory packet was processed. Will be removed.

vktrace/vktrace_common/vktrace_trace_packet_utils.c
```diff
@@ -113,16 +114,163 @@ uint64_t vktrace_get_time() {
...
+        strncpy(pSaveDL, pDL, lenDL + 1);
+    }
+    if (pIL) {
+        pSaveIL = pEnvSave + lenDL + 1;
+        strncpy(pSaveIL, pIL, lenIL + 1);
+    }
+
+    // Set VK_DEVICE_LAYERS and VK_INSTANCE_LAYERS to null.
+    // We don't want any layers for this call to vKCreateInstance,
+    // especially the trace layer!
```
cnorthrop: The Android loader can't enable and disable layers like this.  It also prevents validating vktrace itself.  Is this to prevent tracing of the inserted calls?  I believe there are other ways to make Vulkan calls to the layers below vktrace, or directly into the driver, like the vktrace_lib_trace.cpp already does.  Can we switch to those?

davidlunarg: On looking at this, I realized that this code is being run in the vktrace process, not the trace layer. The querying of the trace configuration should be done in the trace layer, because the vktrace command and the trace program can be on different systems. I'll have to change vktrace and the trace layer to create the file header inside the trace layer.

vktrace/vktrace_common/vktrace_trace_packet_identifiers.h
jyangsh: The workaround should be work. At least for now, because there are not many vulkan games or apps in market now. But in future, in case, vulkan becomes more and more popular(hope it will happen), and many vulkan games need be traced for performance regression. Then, at last, we will have a large trace file repository. Once the compatibility is broken, the repository need be regenerated, that may be a little pain.

vktrace/vktrace_replay/vkreplay_vkreplay.cpp
```diff
@@ -2256,16 +2166,183 @@ VkResult vkReplay::manually_replay_vkAllocateMemory(packet_vkAllocateMemory *pPa
...
+        }
+        if (amIdx == portabilityTable.size()) {
+            // Didn't find the current vkAM packet, something is wrong with the trace file.
+            // Just use the index from the trace file and attempt to continue.
+            vktrace_LogError("Replay of vkAllocateMemory() failed, trace file may be corrupt.");
             replayResult = m_vkFuncs.real_vkAllocateMemory(remappedDevice, pPacket->pAllocateInfo, NULL, &local_mem.replayGpuMem);
-        } else {
-            vktrace_LogError("vkAllocateMemory() failed, couldn't find memory type for memoryTypeIndex");
-            return VK_ERROR_VALIDATION_FAILED_EXT;
+            fseek(tracefp, saveFilePos, SEEK_SET);
```
ardaLunarG: we can replace this with FSEEK macro for consistency throughout the function

vktrace/vktrace_replay/vkreplay_vkreplay.cpp
```diff
@@ -2256,16 +2166,183 @@ VkResult vkReplay::manually_replay_vkAllocateMemory(packet_vkAllocateMemory *pPa
...
+                        replayGetImageMemoryRequirements[m_objMapper.remap_images(bimPacket.image)] = mem_reqs;
+                    } else {
+                        m_vkFuncs.real_vkGetBufferMemoryRequirements(
+                            remappedDevice, m_objMapper.remap_buffers((VkBuffer)bimPacket.image), &mem_reqs);
+                        replayGetBufferMemoryRequirements[m_objMapper.remap_buffers((VkBuffer)bimPacket.image)] = mem_reqs;
+                    }
+                }
+            }
+        }
+        FSEEK(tracefp, saveFilePos, SEEK_SET);
```
ardaLunarG: why do we need to FSEEK here again?

vktrace/vktrace_replay/vkreplay_vkreplay.cpp
```diff
@@ -2256,16 +2166,183 @@ VkResult vkReplay::manually_replay_vkAllocateMemory(packet_vkAllocateMemory *pPa
...
+            }
+
+            if ((packetHeader1.packet_id == VKTRACE_TPI_VK_vkBindImageMemory ||
+                 packetHeader1.packet_id == VKTRACE_TPI_VK_vkBindBufferMemory) &&
+                traceAllocateMemoryRval == bimPacket.memory) {
+                // A vkBIM/vkBBM binds memory allocated by this vkAM call.
+                foundBindMem = true;
+
+                // Search backwards for the vkGIMR/vkGBMR call.
+                bool found = false;
```
ardaLunarG: found is not getting updated in the loop

vktrace/vktrace_replay/vkreplay_vkreplay.cpp
```diff
@@ -2256,16 +2166,183 @@ VkResult vkReplay::manually_replay_vkAllocateMemory(packet_vkAllocateMemory *pPa
...
+                }
+            }
+        }
+        FSEEK(tracefp, saveFilePos, SEEK_SET);
+
+        if (!foundBindMem) {
+            // Didn't find vkBind{Image|Buffer}Memory call for this vkAllocateMemory.
+            // This isn't an error (the memory is probably allocated but never used),
+            // so just use the index from the trace file and continue.
+            replayResult = m_vkFuncs.real_vkAllocateMemory(remappedDevice, pPacket->pAllocateInfo, NULL, &local_mem.replayGpuMem);
```
ardaLunarG: we better use macro FSEEK here for consistency thought the function.

vktrace/vktrace_common/vktrace_trace_packet_utils.c
```diff
@@ -113,22 +114,53 @@ uint64_t vktrace_get_time() {
...
-    pHeader->trace_start_time = vktrace_get_time();
+uint64_t get_endianess() {
+    uint32_t x = 1;
+    return *((char*)&x) ? VKTRACE_BIG_ENDIAN : VKTRACE_LITTLE_ENDIAN;
+}
 
-    return pHeader;
+uint64_t get_arch() {
+    uint64_t rval = 0;
+#if defined(PLATFORM_LINUX)
```
ardaLunarG: which ifdef covers Android?

vktrace/vktrace_replay/vkreplay_vkreplay.cpp
davidlunarg: Can't use the FSEEK macro because we already logged an error and called real_vkAllocateMemory. If we called the FSEEK macro we would potentially do those again.

vktrace/vktrace_replay/vkreplay_vkreplay.cpp
davidlunarg: Will fix.

vktrace/vktrace_replay/vkreplay_vkreplay.cpp
davidlunarg: Need to restore the file positions after we changed it. However, this should be fseek not FSEEK - we already called vkAllocateMemory, want to avoid potentially calling it again. Will be make that change.

vktrace/vktrace_replay/vkreplay_vkreplay.cpp
davidlunarg: This fseek can be eliminated - we just called fseek with the same args on line 2306. Plan to change this.

vktrace/vktrace_common/vktrace_trace_packet_utils.c
davidlunarg: "#if defined (PLATFORM_LINUX)" is for both Linux and Android. "#if defined (ANDROID)" is used for Android-specific code. Cody, can you verify this?

cnorthrop: Yes, the Android build defines PLATFORM_LINUX and tries to inherit most sections unchanged.

PR #214, opened 2017-03-23
Fix macOS vktrace

cnorthrop:
Get the macOS build going again after PR 212.


cnorthrop:
Thanks, cherry-picked to master.

PR #215, opened 2017-03-24
vktrace: Fix vkCreateGraphicsPipelines crash

mizhen:
Add ROUNDUP_TO_4 and ceiling function process for some vk_struct_size_helper functions, the change fix some title crash on vkCreateGraphicsPipelines problem during capture.

XCAP-585

Change-Id: I55f536208cf919f774a4c7f875711ff6e9a95624


ardaLunarG:
@mizhen, we need to apply this patch to LVL repo. 
However as ROUNDUP4 is vktrace specific it would not compile.

We will look at the options to apply these changes.

ardaLunarG:
fixed in PR #217

Issue #182, opened 2017-02-11
Trim crashes when used with screenshot layer


ardaLunarG:
When I enable screenshot layer with trim
- Vktrace dumps heap and crashes
- vkreplay crashes

plohrmann:
I've identified the issue, and a fix is pending internal code review.
* Screenshot layer is caching a VkCommandPool object to use, but didn't handle the situation that the CommandPool got destroyed. The simple solution is to keep a most-recently-created list of command pools, remove items from the list if they are destroyed, then take the CommandPool that is at the front of the list when the screenshot is saved.

ardaLunarG:
fixed

PR #220, opened 2017-03-31
screenshot: Fix screenshot layer + trim compatibility.

plohrmann:
Screenshot layer was not handling the case that the most recently created
CommandPool object was also being destroyed, and thus was trying to use a
destroyed object. The layer now keeps a most-recently-created list of
CommandPools, and will utilize the one that was most recently created and
not yet destroyed. If there are no available CommandPools, then the
screenshot layer will not generate a file.

Fixes issue #182 

Change-Id: I8f248b44272f6b414f6ef578b962a2584c200e62

### DIFF
--- a/layersvt/screenshot.cpp
+++ b/layersvt/screenshot.cpp
@@ -678,8 +678,7 @@ static void writePPM(const char *filename, VkImage image1) {
 
     // Set up the command buffer.  We get a command buffer from a pool we saved
     // in a hooked function, which would be the application's pool.
-    if (deviceMap[device]->commandPools.empty())
-    {
+    if (deviceMap[device]->commandPools.empty()) {
         assert(!deviceMap[device]->commandPools.empty());
         return;
     }
--- a/layersvt/screenshot.cpp
+++ b/layersvt/screenshot.cpp
@@ -159,7 +159,7 @@ typedef struct {
     VkLayerDispatchTable *device_dispatch_table;
     bool wsi_enabled;
     VkQueue queue;
-    VkCommandPool commandPool;
+    std::list<VkCommandPool> commandPools;
     VkPhysicalDevice physicalDevice;
     PFN_vkSetDeviceLoaderData pfn_dev_init;
 } DeviceMapStruct;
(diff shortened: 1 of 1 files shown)


layersvt/screenshot.cpp
```diff
@@ -678,9 +678,17 @@ static void writePPM(const char *filename, VkImage image1) {
 
     // Set up the command buffer.  We get a command buffer from a pool we saved
     // in a hooked function, which would be the application's pool.
+    if (deviceMap[device]->commandPools.empty())
+    {
```
ardaLunarG: Does not fit to clang here

plohrmann: Will fix.

layersvt/screenshot.cpp
```diff
@@ -678,9 +678,16 @@ static void writePPM(const char *filename, VkImage image1) {
 
     // Set up the command buffer.  We get a command buffer from a pool we saved
     // in a hooked function, which would be the application's pool.
+    if (deviceMap[device]->commandPools.empty()) {
+        assert(!deviceMap[device]->commandPools.empty());
```
davidlunarg: Do we want this assert here? If there isn't a command pool to use, this will crash the process when compiled Debug, but will silently do nothing in Release mode.  Maybe the behavior should be the same for Release/Debug?

layersvt/screenshot.cpp
```diff
@@ -1207,16 +1235,16 @@ VKAPI_ATTR VkResult VKAPI_CALL QueuePresentKHR(VkQueue queue, const VkPresentInf
...
-                for (auto it = imageMap.begin(); it != imageMap.end(); it++) {
-                    ImageMapStruct *imageMapElem = it->second;
+                for (auto imageIter = imageMap.begin(); imageIter != imageMap.end(); imageIter++) {
+                    ImageMapStruct *imageMapElem = imageIter->second;
                     delete imageMapElem;
                 }
-                for (auto it = physDeviceMap.begin(); it != physDeviceMap.end(); it++) {
-                    PhysDeviceMapStruct *physDeviceMapElem = it->second;
+                for (auto physDeviceIter = physDeviceMap.begin(); physDeviceIter != physDeviceMap.end(); physDeviceIter++) {
+                    PhysDeviceMapStruct *physDeviceMapElem = physDeviceIter->second;
```
davidlunarg: Why these changes renaming the iterator variable in this pull request? Seems unrelated to the gist of this pull request. But at this point, leave the changes in.

layersvt/screenshot.cpp
plohrmann: I put the assert in because all the previous errors in this function which return, also have an assert. I agree they will have different behavior in release and debug, but was just following the existing behavior.

layersvt/screenshot.cpp
plohrmann: Right it's unrelated to this change, but I saw a build warning related to the duplicate used variable name (there's already an iterator called 'it', so changed this one to ensure there wasn't any conflicts.

layersvt/screenshot.cpp
davidlunarg: OK. Better to be consistent with existing code.

PR #230, opened 2017-04-12
add ROUNDUP_TO_4 process for adding buffer to trace_packet and vkUnmapMemory 

mizhen:
Some title is not 4 byte aligned when call vkMapMemory, it cause error at adding buffer to trace_packet and vkUnmapMemory.


ardaLunarG:
cherry picked to master

PR #231, opened 2017-04-26
VKTRACE: Move vkreplay_vk_objmapper.h to use vk.xml-based codegen

mark-lunarg:
This file is the first of the 10 or so codegen'd vktrace files which rely on vulkan.py to be moved to xml-based codegen.


mark-lunarg:
Pushed to Master.

PR #235, opened 2017-05-01
vktrace: Remove memcpy to shadow memory in map memory process

mizhen:
In current implementation, once vkMapMemory get captured, vktrace allocate shadow memory with same mapped size, then memcpy real mapped memory to shadow memory. The change (only for Windows) removes the memcpy, and adds additional process in page guard exception handler: only if a page is first time accessed, the handler memcpy that page from real mapped memory to shadow memory. This optimization significantly improve target title capture time from hours to minutes, because the title use lots of big size map/unmap, and only a small portion of mapped size accessed by host side.

XCAP-585

Change-Id: I29f3add4553576679b3196f0b1595c096b77143e


davidlunarg:
I just tested this change on Linux. I created a trace of Dota2, then replayed it. Replay generates several errors: 

    vkreplay error: Invalid handle in vkAllocateMemory pAllocate->pNext structure.

I don't know if Ming's changes introduced these errors, will investigate further.

PR #236, opened 2017-05-03
Move tracke packet id definitions out of codegen

mark-lunarg:
Moved definitions into the existing ID definition header file per @ardaLunarG's suggestion.

WAS: [Created a new header file in the common directory containing just the enum used for these definitions.  This seems to fix the weird (really weird) issues we were seeing yesterday, and may make it a bit tougher to break compatibility related to ID definitions in the future.]


davidlunarg:
Is it true that when new entry points are added, new packet id definitions won't be automatically generated? Or do I have that right?

mark-lunarg:
Everything will be generated as before EXCEPT for the ID.  So new APIs will
generate compilation errors and will need to be manually added to this
header file.

vktrace/vktrace_common/vktrace_packet_id_defs.h
```diff
@@ -0,0 +1,195 @@
...
+ * Author: Mark Lobodzinski <mark@lunarg.com>
+ */
+
+#pragma once
+
+// Note: Changing the following enums will break backwards-compatibility with
+//       previously generated traces.  New values should be appended, deprecated
+//       values should be left in or skipped.
+enum VKTRACE_TRACE_PACKET_ID_VK {
+    VKTRACE_TPI_VK_vkApiVersion = 7,
```
ardaLunarG: I understand the reason we start at 7 is the fact that this enum table continuing from VKTRACE_TRACE_PACKET_ID. in this first enum currently VKTRACE_BEGIN_API_HERE is 7 but this can change if new enums added to first enum table. Thus making this list start at 7 is still risky. I think we can merge VKTRACE_TRACE_PACKET_ID enum into this table (as this is not generated anymore) and keep them all in same palce

mark-lunarg: This is a brilliant idea, Arda.  Will update patch shortly.

mark-lunarg:
@Tony-LunarG, can you try this branch once again?

PR #238, opened 2017-05-04
Fix location of vk_layer_settings.txt on Android

cforfang:
Commit 969ab0680dd434f0 ('api_dump: Support config file and logging on Android')
added a custom location for vk_layer_settings.txt on Android. However, commit
750837a5dedf524c81c ('Merge branch 'trunk'') ended up partly overwriting these changes,
breaking the functionality.

As a result, using the config file on Android hasn't worked since October 2016. This commit
restores the changes in the original commit to fix this.

### DIFF
--- a/layers/vk_layer_config.cpp
+++ b/layers/vk_layer_config.cpp
@@ -200,7 +200,7 @@ const char *ConfigFile::getOption(const std::string &_option) {
             }
             parseFile(envPath.c_str());
         } else {
-            parseFile("vk_layer_settings.txt");
+            parseFile(m_fileName.c_str());
         }
     }
 
@@ -223,7 +223,7 @@ void ConfigFile::setOption(const std::string &_option, const std::string &_val) 
             }
             parseFile(envPath.c_str());
         } else {
-            parseFile("vk_layer_settings.txt");
+            parseFile(m_fileName.c_str());
         }
     }
 


cforfang:
I hit this issue trying to use api_dump, which seems to live here. 

Looks like [Vulkan-LoaderAndValidationLayers](https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/) has the same issue though -- I can submit a pull-request there as well.

cforfang:
Actually, looks like LoaderAndValidationLayers doesn't have 969ab06. @cnorthrop, should we move your changes there?

cnorthrop:
api_dump (and other utility layers) live in this repo.  I'm giving your changes a try, then we'll merge it.  We really need some sort of test to ensure api_dump doesn't break again. :\

cnorthrop:
Confirmed this gets things going again.  Many thanks for the patch!  I'll hook something up that invokes cube or smoke with api_dump, ensure the output file is created and contains something sensible.

cherry-picked to master

cnorthrop:
Actually, this file lives upstream doesn't it.  Bleh.  I'll apply it upstream as well and apologize to Arda.

cnorthrop:
As you astutely pointed out, the original Android changes weren't applied upstream, so I broke master upstream and apologized to everyone.  I'll look into adding Android support upstream, for layers other than just api_dump.

PR #244, opened 2017-05-26
vktrace: Unload the network for Android

jyangsh:
Some APPs create/destroy instance several times in their lifecycle, so the network need be unloaded in the vkDestroyInstance() call to avoid the following vkCreateInstance() calls fail.


davidlunarg:
The new _Unload function looks like it was duplicated from vktrace_layer/vktrace_lib.c. Is there a reason that the existing function could not be used? Can the existing function be moved to some other file so that only the function is defined only once in the source?

jyangsh:
The vktrace_lib.c is not compiled in Android build, so I copied it to the vktrace_lib_trace.cpp for Android.

Now I have moved it to the vktrace_lib_helpers.h and deleted the duplication.

davidlunarg:
I tried the latest change in which _Unload is moved to vktrace_lib_helpers.h on Linux. Since this header file is included by several .cpp and .c files, this results in _Unload being called 10 times when the trace library is unloaded.

vktrace/vktrace_layer/vktrace_lib.c
```diff
@@ -24,6 +24,7 @@
 #include "vktrace_interconnect.h"
 #include "vktrace_vk_vk.h"
 #include "vktrace_lib_trim.h"
+#include "vktrace_lib_helpers.h"
```
davidlunarg: I don't think this header file needs to be included here.

vktrace/vktrace_layer/vktrace_lib_helpers.h
```diff
@@ -25,6 +25,7 @@
 #include "vktrace_vk_vk.h"
 #include "vulkan/vk_layer.h"
 #include "vktrace_platform.h"
+#include "vktrace_lib_trim.h"
```
davidlunarg: I don't think this header needs to be included.

jyangsh:
Thanks for review.:)

davidlunarg:
Merged to master

PR #245, opened 2017-06-05
apidump HTML backend

jeremyk-lunarg:
Added an HTML backend to the apidump layer.


cnorthrop:
Any chance you have a test we can use to verify api_dump is still working as expected?

lenny-lunarg:
We don't have a test in the repo for api dump. We do run brief tests on it as a part of the SDK testing, but that isn't in the repo. It might be a good thing to add in the future, but since we haven't had api dump tests in the past, that's an improvement I don't think we need right now.

PR #250, opened 2017-06-09
Avoid modifying application memory in __HOOKED_vkCreateDevice()

jozefkucia:
Fixes a crash when an application uses a static const array for
ppEnabledLayerNames or ppEnabledExtensionNames.

Allocate memory for ppEnabledLayerNames and ppEnabledExtensionNames
arrays similarly how it's done in __HOOKED_vkCreateInstance().

### DIFF
--- a/vktrace/vktrace_layer/vktrace_lib_trace.cpp
+++ b/vktrace/vktrace_layer/vktrace_lib_trace.cpp
@@ -953,14 +953,6 @@ VKTRACER_EXPORT VKAPI_ATTR VkResult VKAPI_CALL __HOOKED_vkCreateDevice(VkPhysica
     // remove the loader extended createInfo structure
     VkDeviceCreateInfo localCreateInfo;
     memcpy(&localCreateInfo, pCreateInfo, sizeof(localCreateInfo));
-    for (uint32_t i = 0; i < pCreateInfo->enabledExtensionCount; i++) {
-        char** ppName = (char**)&localCreateInfo.ppEnabledExtensionNames[i];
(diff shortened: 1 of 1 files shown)


vktrace/vktrace_layer/vktrace_lib_trace.cpp
```diff
@@ -953,6 +953,13 @@ VKTRACER_EXPORT VKAPI_ATTR VkResult VKAPI_CALL __HOOKED_vkCreateDevice(VkPhysica
     // remove the loader extended createInfo structure
     VkDeviceCreateInfo localCreateInfo;
     memcpy(&localCreateInfo, pCreateInfo, sizeof(localCreateInfo));
+
```
karl-lunarg: I think that these new lines aren't needed.  

A better fix might be to simply remove lines 956-963.  `add_VkDeviceCreateInfo_to_packet()` is probably the only thing that references these lists of strings and all it does is copy the strings into the trace buffer.  The existing arrays of char* owned by the app can still be used for this since they don't need to be modified.

The existing code at 956-963 is modifying the app's memory when it doesn't need to.  And it is just copying the pointers back on top of themselves anyway.

The only reason to keep the existing code (and add the suggested lines) is if the lists of extensions or layers can be  modified someplace that I may have missed.

jozefkucia: The lists of extensions or layers don't appear to be modified. I've updated the fix.

davidlunarg:
Merged to master.

Issue #249, opened 2017-06-08
layersvt install libs into wrong destination (hardcoded)


sl1pkn07:
when configure cmake with:

`
-DCMAKE_INSTALL_PREFIX=/usr
`
and after build, try install the layersvt with

`
cd layersvt
make install
`

the lib files is installed into `install_staging` folder instead of `/usr`

https://github.com/LunarG/VulkanTools/blob/master/layersvt/CMakeLists.txt#L108

the fix is change the

`${PROJECT_BINARY_DIR}/install_staging`

to

`${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}`

greetings

karl-lunarg:
Thanks!   I think that there is a lot of other things missing from the `make install` support for this repo as well.  We'll work on it.

karl-lunarg:
Related:  https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/pull/1844

karl-lunarg:
Fixed with #251.

This PR addresses the original problem stated in this issue, plus `make install` now installs the vktrace layer and executables and also installs via.

PR #251, opened 2017-06-12
build: Add VT-specific artifacts to install

karl-lunarg:
Adding:
VT-specific layers
via executable
vktrace layer and trace/replay executables


karl-lunarg:
This fixes #249

PR #256, opened 2017-06-16
layersvt: Fix false UNUSED reports

krOoze:
- fix `VkPipelineViewportStateCreateInfo::pViewports` UNUSED when not
dynamic
- fix `VkPipelineViewportStateCreateInfo::pScissors` "UNUSED" when not
dynamic
- fix  `VkCommandBufferBeginInfo::pInheritanceInfo "UNUSED" when
secondary
- add flexibility that can help handle similar hard to implement cases

### DIFF
--- a/layersvt/api_dump.h
+++ b/layersvt/api_dump.h
@@ -38,6 +38,11 @@
 #include <string.h>
 #include <string>
 #include <type_traits>
+#include <map>
+#include <unordered_map>
+#include <vector>
+#include <unordered_set>
+#include <utility>
 
 #define MAX_STRING_LENGTH 1024
 
@@ -318,6 +323,7 @@ class ApiDumpInstance {
         loader_platform_thread_create_mutex(&output_mutex);
         loader_platform_thread_create_mutex(&frame_mutex);
         loader_platform_thread_create_mutex(&thread_mutex);
+        loader_platform_thread_create_mutex(&cmd_buffer_state_mutex);
     }
 
     inline ~ApiDumpInstance() {
@@ -326,6 +332,7 @@ class ApiDumpInstance {
         loader_platform_thread_delete_mutex(&thread_mutex);
         loader_platform_thread_delete_mutex(&frame_mutex);
         loader_platform_thread_delete_mutex(&output_mutex);
+        loader_platform_thread_delete_mutex(&cmd_buffer_state_mutex);
     }
 
     inline uint64_t frameCount() {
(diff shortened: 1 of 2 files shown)


layersvt/api_dump.h
```diff
@@ -366,6 +373,58 @@ class ApiDumpInstance {
...
+    inline void eraseCmdBuffers(VkDevice device, VkCommandPool cmd_pool, std::vector<VkCommandBuffer> cmd_buffers) {
+        loader_platform_thread_lock_mutex(&cmd_buffer_state_mutex);
+
+        const auto pool_cmd_buffers_iter = cmd_buffer_pools.find(std::make_pair(device, cmd_pool));
+        assert(pool_cmd_buffers_iter != cmd_buffer_pools.end());
+
+        for (const auto cmd_buffer : cmd_buffers) {
+            pool_cmd_buffers_iter->second.erase(cmd_buffer);
+
+            assert(cmd_buffer_level.count(cmd_buffer) > 0);
```
krOoze: Does not account for potential `NULL` elements in `pCommandBuffers`.

layersvt/api_dump.h
```diff
@@ -366,6 +373,58 @@ class ApiDumpInstance {
...
+        }
+
+        loader_platform_thread_unlock_mutex(&cmd_buffer_state_mutex);
+    }
+
+    inline void eraseCmdBufferPool(VkDevice device, VkCommandPool cmd_pool) {
+        loader_platform_thread_lock_mutex(&cmd_buffer_state_mutex);
+
+        const auto cmd_buffers_iter = cmd_buffer_pools.find(std::make_pair(device, cmd_pool));
+        assert(cmd_buffers_iter != cmd_buffer_pools.end());
```
krOoze: Does not account for `VK_NULL_HANDLE`.

krOoze:
@lenny-lunarg Also necessary, because nulls are perfectly valid values there.
I am not sure what your policy is otherwise. Since this is non-validation layer, I am assuming the App is a valid Vulkan usage. Should I also be protecting it against API abuse or not?
OK, done.

lenny-lunarg:
I had forgotten that nulls were legal there, so you're right then and the null checks are necessary. In general, you can assume valid Vulkan usage in non-validation layers. It's nicer if we can at least not crash as a result of invalid Vulkan use, but it's definitely not something we guarantee. There's a lot of places where you essentially can't guarantee there won't be a crash if the Vulkan use is invalid (such as if the user passes invalid pointers). So I think the way you went about it is just fine.

Assuming appveyor succeeds, I'll merge your change in once that finishes.

PR #257, opened 2017-06-17
layersvt: Document api_dump settings

krOoze:
i.e. add `output_format` and `show_shader` documentation in `vk_layer_settings.txt`.


lenny-lunarg:
Thanks for catching the documentation we forgot to update.

PR #260, opened 2017-06-22
layersvt: Create device simulation (devsim) layer

mikew-lunarg:
The Device Simulation Layer is a Vulkan layer that can modify the results of Vulkan PhysicalDevice queries based on a configuration file, thus simulating a different device than is actually in a system.  A goal is to simplify application testing for a wide range of simulated devices, without actually requiring a physical copy of every device.

This "version 0" prerelease is functional, but many things could change, depending on feedback, before a "version 1" release.


danginsburg:
+1 Tested with dota2 on Windows and appears to be working well.

PR #263, opened 2017-07-05
build: enable JsonCpp separate from VIA

mikew-lunarg:
As other tools beyond VIA begin to use JsonCpp, enable it
even if VIA is not being built.

Change-Id: I506345c26010e9f3d97cc5f2fbb2b56a054f927d


mikew-lunarg:
@MarkY-LunarG gave verbal go-ahead to push.

PR #268, opened 2017-07-19
vkreplay: Fix GH184 - Unmap swapchain images

mvicomoya:
VkImages obtained through vkGetSwapchainImagesKHR are inserted into the
images map, but never removed when the VkSwapchainKHR is destroyed.

It may cause issues with traces from applications that re-create a
VkSwapchainKHR multiple times (e.g. upon window resize) due to the fact
that VkImage handles might be re-used by the underlying driver.

A call to vkCreateImage could return the same handle of a swapchain
VkImage that no longer exists, but since vkreplay didn't clean those
mappings up, it thinks the image was already created and skip the real
vkCreateImage call.

This change adds a swapchainToImages map to keep track of what images
were obtained through vkGetSwapchainImagesKHR, can remove the
corresponding mappings at VkDestroySwapchainKHR time.

### DIFF
--- a/scripts/vktrace_file_generator.py
+++ b/scripts/vktrace_file_generator.py
@@ -589,6 +589,7 @@ class VkTraceFileOutputGenerator(OutputGenerator):
                                  'GetPhysicalDeviceSurfaceFormatsKHR',
                                  'GetPhysicalDeviceSurfacePresentModesKHR',
                                  'CreateSwapchainKHR',
+                                 'DestroySwapchainKHR',
                                  'GetSwapchainImagesKHR',
                                  'CreateXcbSurfaceKHR',
                                  'CreateXlibSurfaceKHR',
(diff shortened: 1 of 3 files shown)


davidlunarg:
The CLA needs to be signed before this pull request can be merged. As soon as it is signed I will merge it.

mvicomoya:
My understanding is that contributors from @NVIDIA aren't required to sign the individual CLA, as we have already sign an overarching corporate CLA that covers all our contributions.

Please, let me know if I'm mistaken.

davidlunarg:
VulkanTools is a LunarG github project. I'm not aware of the existence a CLA between nVidia and Lunarg for VulkanTools or any other LunarG github project. So a CLA is needed from each individual contributor.

BTW, the CLA for LunarG github projects is identical to the CLA for Khronous github projects.

davidlunarg:
mvicomoya - Are you still investigating the need for a CLA for contributing to VulkanTools? This is a valuable pull request, that addresses a bug likely to occur. I would really like to incorporate this code change.

mvicomoya:
Legal issues are finally sorted out. Sorry for the delay.

davidlunarg:
Merged

PR #272, opened 2017-08-03
vkreplay: Fix reset frame number when loopStartFrame is given.

justuswang:
This change sets m_frameNumber to loopStartFrame after each loop to
make the frame number counter m_frameNumber consistent with the frame
which startingPacket belongs to.

Otherwise the vkreplay will replay frames starts from loopStartFrame to
(loopStartFrame + loopEndFrame) in the second or later loop.  Because
the startingPacket is set to the first packet of frame "loopStartFrame"
but the m_frameNumber is reset to 0 which means the replayed content is
starts from the expected loopStartFrame but the internal frame counter
is starts from 0 instead of loopStartFrame.

### DIFF
--- a/vktrace/vktrace_replay/vkreplay.cpp
+++ b/vktrace/vktrace_replay/vkreplay.cpp
@@ -150,8 +150,8 @@ int VKTRACER_CDECL VkReplayGetFrameNumber() {
     return -1;
 }
 
-void VKTRACER_CDECL VkReplayResetFrameNumber() {
+void VKTRACER_CDECL VkReplayResetFrameNumber(int frameNumber) {
     if (g_pReplayer != NULL) {
-        g_pReplayer->reset_frame_number();
+        g_pReplayer->reset_frame_number(frameNumber);
     }
 }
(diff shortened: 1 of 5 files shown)


davidlunarg:
I build VulkanTools with this change and found a potential problem.

I created a trace file of cube, then replayed it:

> vkreplay -t cube.vktrace -lsf 6 -lef 50 -l 50

I get a few errors, then a segmentation fault:

> vkreplay info: Loop number 50 completed. Remaining loops:49
> vkreplay info: Loop number 49 completed. Remaining loops:48
> vkreplay info: Loop number 48 completed. Remaining loops:47
> vkreplay info: Loop number 47 completed. Remaining loops:46
> vkreplay error: Failed to remap pImageIndex.
> vkreplay error: Skipping vkQueuePresentKHR() due to invalid remapped pImageIndices.
> vkreplay error: Return value VK_ERROR_VALIDATION_FAILED_EXT from API call (vkQueuePresentKHR) does not match return value from trace file VK_SUCCESS.
> vkreplay error: Failed to replay packet_id 165, with global_packet_index 187.
> vkreplay info: Loop number 46 completed. Remaining loops:45
> Segmentation fault (core dumped)
> 

It replays successfully if I build VulkanTools from master.

I'm not sure why this is happening, or if the problem was introduced by this pull request. This needs to be further investigated.

justuswang:
I have found the root cause and added a new commit to fix it.

With the new commit, vkreplay will only record the location of looping start packet in the first loop when loopStartFrame is not 0.

Otherwise the startingPacket offset will be increased by one packet in each loop because "seq.get_bookmark(startingPacket)" is called after "seq.get_next_packet()" when "frameNumber == settings.loopStartFrame".

davidlunarg:
Merged to master

PR #273, opened 2017-08-04
vktrace: Fix trim trace file playback crash

mizhen:
vkreplay crash during playback trimmed trace file for some title. The
change fixes the problem by adding process in vktrace trim module to
record vkGetPhysicalDeviceProperties packet, the packet is needed by
portability process during playback.

Change-Id: I9692a99614818679123cdd1bbbca0db56da58434


davidlunarg:
Merged to master

PR #275, opened 2017-08-09
layersvt: Make DevSim work on Android

justuswang:
Make device simulation layer work on Android and update vktrace to not record it in the trace file.


layersvt/device_simulation.cpp
```diff
@@ -48,6 +48,8 @@
 #include <fstream>
 #include <mutex>
 
+using namespace std;
```
cnorthrop: I believe this is against convention.  Go ahead and include std:: in your uses of the library.

layersvt/device_simulation.cpp
```diff
@@ -103,26 +144,57 @@ std::string GetEnvarValue(const char *name) {
...
+void AndroidPrintf(const char *level, const char *fmt, va_list args) {
+    int requiredLength;
+    va_list argcopy;
+    va_copy(argcopy, args);
+    requiredLength = vsnprintf(NULL, 0, fmt, argcopy) + 1;
+    va_end(argcopy);
+
+    char *message = (char *)malloc(requiredLength);
+    vsnprintf(message, requiredLength, fmt, args);
+    __android_log_print(ANDROID_LOG_INFO, "devsim", "%s: %s", level, message);
```
cnorthrop: This should select the log priority based on level.  There's an example of this in [vktrace_lib.c](https://github.com/LunarG/VulkanTools/blob/master/vktrace/vktrace_layer/vktrace_lib.c#L104)

layersvt/device_simulation.cpp
mikew-lunarg: Agreed.

layersvt/device_simulation.cpp
```diff
@@ -73,16 +75,54 @@ const uint32_t kLayerPropertiesCount = (sizeof(kLayerProperties) / sizeof(kLayer
 const VkExtensionProperties *kExtensionProperties = nullptr;
 const uint32_t kExtensionPropertiesCount = 0;
 
+#if !defined(__ANDROID__)
```
mikew-lunarg: If this is for the unused variable, note that I have a change in the works that will use it.
So if the unused var is causing trouble, we can leave the #if in there for the moment, and I'll remove it in my upcoming PR.

layersvt/device_simulation.cpp
```diff
@@ -103,26 +144,57 @@ std::string GetEnvarValue(const char *name) {
     return value;
 }
 
+#if defined(__ANDROID__)
+void AndroidPrintf(const char *level, const char *fmt, va_list args) {
```
mikew-lunarg: As @cnorthrop mentioned above, perhaps this could also go into a common location for Android utilities.

I'm open to refactoring the debug and error message reporting interfaces, if it would reduce the need for #if  conditionals below.

build-android/jni/Application.mk
```diff
@@ -1,21 +1,21 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+APP_ABI := armeabi-v7a arm64-v8a x86 x86_64 mips mips64
+APP_PLATFORM := android-22
+APP_STL := gnustl_static
+APP_MODULES := layer_utils VkLayer_core_validation VkLayer_parameter_validation VkLayer_object_tracker VkLayer_threading VkLayer_unique_objects VkLayer_api_dump VkLayer_screenshot VkLayerValidationTests VulkanLayerValidationTests VkLayer_vktrace_layer vkreplay VkLayer_device_simulation
```
justuswang: Sure.  The Windows line endings have been added back to reduce the diff to a single line.

layersvt/device_simulation.cpp
justuswang: Got it.  Will use atoi() for Android and std::atoi() for none Android explicitly.

layersvt/device_simulation.cpp
justuswang: Yes, it is just to avoid an unused variable to get rid of the "unused variable" error reported by the ndk-build.

layersvt/device_simulation.cpp
```diff
@@ -73,16 +75,54 @@ const uint32_t kLayerPropertiesCount = (sizeof(kLayerProperties) / sizeof(kLayer
...
 const char *const kEnvarDevsimExitOnError = "VK_DEVSIM_EXIT_ON_ERROR";  // a non-zero integer will enable exit-on-error.
+#endif
 
 // Various small utility functions ///////////////////////////////////////////////////////////////////////////////////////////////
 
+#if defined(__ANDROID__)
+#include <android/log.h>
+const uint32_t MAX_BUFFER_SIZE = 255;
+
+const char *AndroidGetEnv(const char *key) {
```
justuswang: Yes, I have noticed the similar code in screenshot but not sure where should the common location be.

layersvt/device_simulation.cpp
justuswang: Thanks very much for pointing it out.  I will update it accordingly.

layersvt/device_simulation.cpp
```diff
@@ -103,26 +144,57 @@ std::string GetEnvarValue(const char *name) {
...
     va_start(args, fmt);
+#if defined(__ANDROID__)
+    AndroidPrintf("Errors", fmt, args);
+#else
     vfprintf(stderr, fmt, args);
+#endif
     va_end(args);
     if (kExitLevel > 0) {
+#if defined(__ANDROID__)
+        __android_log_print(ANDROID_LOG_INFO, "devsim", "Errors: devsim exiting on error as requested");
```
justuswang: Will change it to ANDROID_LOG_ERROR.

layersvt/device_simulation.cpp
cnorthrop: Yeah, the print routines are a bit messy.  Maybe just have wholly different Android functions for `DebugPrintf` an `ErrorPrintf`.  We can refactor them into `VulkanTools/common` and migrate the other layers to use them.  I see screenshot has picked up at least one `fprintf` getting dropped on Android.  I don't think we need to hold up this PR for that though.  We may even want that to happen upstream.

cnorthrop:
@mikew-lunarg pointed me at [tests/devsim_layer_test.sh](https://github.com/LunarG/VulkanTools/blob/master/tests/devsim_layer_test.sh) for validation.  I'll port that to Android to ensure this works and keeps working.  I also need to check in the script I've been using to track VulkanTools quality on our farm.

mikew-lunarg:
@cnorthrop Going forward, shall I add you as a reviewer for future PRs?

cnorthrop:
Sure, although I probably won't add a lot of value to devsim specific issues.  I just try to keep Android support chugging along.  I'll let you know when I've got the test script running.  I might be able to extend the approach to api_dump, which has no Android testing at the moment.

cnorthrop:
cherry-picked to master

PR #280, opened 2017-08-16
android: Enable vkjson_info and add test for devsim layer

cnorthrop:
This does a trivial enable of `vkjson_info` for Android.  It can be expanded to an APK with some boilerplate work, but this exe and test script gets device simulation layer validation off the ground for devices with root access.

The test script can be run like so, from build-android:
`./devsim_layer_test_android.sh --serial <serial> --abi <abi>`

It currently passes in 64-bit mode, but 32-bit fails due to know differences in how `vkjson` emits 64-bit values.

### DIFF
--- a/libs/vkjson/vkjson.cc
+++ b/libs/vkjson/vkjson.cc
@@ -34,11 +34,16 @@
 
 #include <cJSON.h>
 #include <vulkan/vk_sdk_platform.h>
+#include <stdlib.h>
 
 namespace {
 
 inline bool IsIntegral(double value) {
+#if defined(ANDROID)
+  return trunc(value) == value;
+#else
   return std::trunc(value) == value;
+#endif
 }
 
 template <typename T> struct EnumTraits;
--- a/BUILDVT.md
+++ b/BUILDVT.md
@@ -274,6 +274,14 @@ Run your application with the following layer enabled:
 ```
 VK_LAYER_LUNARG_device_simulation
 ```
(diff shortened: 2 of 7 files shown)


mikew-lunarg:
@karl-lunarg This has changes to vkjson, albeit minor, due to lack of a better tool for capturing devsim configuration locally.  May want to increase priority of Jira VT-47 for creating such a tool.

libs/vkjson/vkjson.cc
```diff
@@ -34,11 +34,16 @@
 
 #include <cJSON.h>
 #include <vulkan/vk_sdk_platform.h>
+#include <stdlib.h>
 
 namespace {
 
 inline bool IsIntegral(double value) {
+#if defined(ANDROID)
+  return trunc(value) == value;
```
mikew-lunarg: Android toolchain won't support explicit namespace qualification?
Is there an implicit using decl, or is there no std namespace at all, etc?

cnorthrop: The NDK's STL support is partial, trunc isn't included (although it might be in NDK r16).  This instead uses trunc from the NDK in sysroot/usr/include/math.h

cnorthrop:
Ugh, first time I've tried to contribute to vkjson... I'll split this up and move those parts upstream.

cnorthrop:
I'll reopen another PR with just the new test and documentation.

PR #281, opened 2017-08-17
android: Add test for devsim layer

cnorthrop:
This script targets Android and performs the same golden output check as the desktop version.

The test script can be run like so, from build-android:
`./devsim_layer_test_android.sh --serial <serial> --abi <abi>`

It currently passes in 64-bit mode, but 32-bit fails due to know differences in how vkjson emits 64-bit values.


cnorthrop:
This is checked in and in use on our end.

PR #282, opened 2017-08-18
vkreplay: Support Android device replays on viewer

g-keen:
Adds support on both the replayer APK and VkTraceViewer to connect via 
ADB, allowing synchronized playback of a trace to the UI timeline on a 
mobile device.

This commit adds:
- a UI checkbox to toggle replay location (remote device or local PC).
- A new argument for the replayer APK to signify how to replay a trace.
- A second loop method on the APK to better replay a trace from viewer.
- Connection using ADB and abstract UNIX sockets.
- Minor fix to some type comparisons.


cnorthrop:
Cool, thanks for doing this!  Adding Android support to VkTraceViewer has been on my TODO list for quite a while.  I'll need to familiarize myself with the tool before I can do a detailed review.

After a quick glance through it, a question for @davidlunarg, it looks like the commit assumes all remote replays will be on Android.  Is that going to be the case?  Might want to rename the functions to specify that they are for Android, or at least add a check for what the target platform is somehow.

Also, when I last looked into this when trying to load an Android trace for inspection, I thought WSI functions for Android were missing.  Have they been added since?  Worth double checking.

g-keen:
Hi cnorthrop,
Thanks for the comment, I hope this contribution will help many with debugging their apps.

From my observations, Android support in VkTraceViewer seemed like it was already there in terms of successfully displaying API calls in the UI. What was lacking, and what this contribution adds, was a method of replaying these Android traces whilst being able to observe the progress in the timeline, and doing so by communicating with the device and replaying through the APK. 

As for your second point, I have tried to keep the code as device-ambiguous as possible. You are correct in mentioning there is currently only Android support, however these should only exist as the system ADB calls which can be easily relocated or switched with other OS alternatives. Due to my current testing environment, Android was the only OS I could verify as functional. I do agree there should be some sort of check to test the target platform, if not a user-defined one.

Could I get some clarification as to what you mean about the "WSI functions for Android missing"?

VkTrace recording and replaying works fine on Android as is, using the APK and shell script, and WSI appears to be fine.

Thanks,
Gary

cnorthrop:
By WSI I meant calls specific to Android, i.e.: `VkAndroidSurfaceCreateInfoKHR`, `vkCreateAndroidSurfaceKHR`.  Those found in the `VK_USE_PLATFORM_ANDROID_KHR` section of `vulkan.h`.

I haven't had a chance to try the changes yet, but I plan to this week.  Sorry for the delay!

cnorthrop:
I tried this out.  It looks like vktraceviewer doesn't support 32-bit traces?  I wasn't able to load an armeabi-v7a trace.  One from arm64-v8a worked fine.

Once loaded, the replay buttons were grayed out.  How are you enabling them?  Maybe this CL needs documentation additions?  Or... *is* there any vktraceviewer documentation?

g-keen:
Unfortunately I myself haven't been able to test 32-bit traces but they should be working so long as your test application is built with the layers for the corresponding ABI and a matching replayer APK. I'm aware there was an issue raised with recording 32-bit traces on a 64-bit vktrace, perhaps that is related to the problem?

As for the options being greyed out you will need to manually switch ENABLE_REPLAY to true in vktrace/vktrace_viewer/vktraceviewer_QReplayWorker.h to enable the replayer buttons. I decided to keep it set to false as this contribution doesn't necessarily focus on the problems that caused the replayer to be disabled in the first place.

cnorthrop:
arm32 trace/replay by itself is working currently, you just have to ensure you are using 32-bit vktrace server.  We test this case with internal CI using the following invocation:
```
./vktracereplay.sh \
 --serial $serial \
 --abi armeabi-v7a \
 --apk ../demos/android/cube-with-layers/bin/cube-with-layers.apk \
 --package com.example.CubeWithLayers \
 --activity android.app.NativeActivity \
 --frame 100
```
I'm unable to build vktraceviewer in 32-bit right now, I believe due to my Qt setup.  I'd rather not muck with that at the moment, so arm64 testing will suffice.

With your suggested manual enable of replay, I'm able to load an arm64 trace and step through it, seeing the results on an Android device, so that's cool.

I'll work with @davidlunarg on next steps.  Thanks!

davidlunarg:
@cnorthrop, @g-keen : What is the status of this pull request? I don't have an Android setup to build and test this myself. No problem leaving it open if it hasn't been tested yet.

cnorthrop:
The code is functional for arm64 once you hand enable replay, and the ADB commands are fine, but vktraceviewer doesn't seem to have an owner.  I'm fine with the code going in, could be very useful in the future.  I would prefer there was someone with an eye towards the future of this project who could verify the direction.  You should also make sure it doesn't change desktop behavior.

PR #284, opened 2017-08-28
vktrace: fix gcc version query

tpalli:
Version returned by gcc might not have minor version, this results
to following cmake error:

"CMake Error at vktrace/CMakeLists.txt:223 (list):
  list index: 1 out of range (-1, 0)"

When version returned by gcc equals to '7'.

fix this by setting minor as 0 in such case.

Signed-off-by: Tapani Pälli <tapani.palli@intel.com>


davidlunarg:
LGTM. Merged to master.

PR #286, opened 2017-08-28
vktrace_layer: require xcb-keysyms package

tpalli:
Otherwise cmake passes fine but we get failure only when linking:

fatal error: xcb/xcb_keysyms.h: No such file or directory

Signed-off-by: Tapani Pälli <tapani.palli@intel.com>


davidlunarg:
LGTM. Merged to master.

PR #290, opened 2017-09-01
vkreplay: Reduce CPU overhead in the main loop when using portability table

justuswang:
Move portability table lookup (in vkAM) out from main loop to reduce
CPU overhead in the main loop when portability table is being used.

Encapsulate all the direct call of fread/ftell/fseek with FileLike

Do alignment for the memory size needed in vkAM call to make sure it
allocates memory with the expected size during replay.  Only works when
portablility table is besing used and the alignment in vkGBMR is
different in trace platform and replay platform.

This pull request also includes:
* Print useful info when user tries to replay a 32bit trace file using 64bit vkreplay and vice versa.

### DIFF
--- a/scripts/vktrace_file_generator.py
+++ b/scripts/vktrace_file_generator.py
@@ -614,6 +614,7 @@ class VkTraceFileOutputGenerator(OutputGenerator):
                                  'DestroyDescriptorUpdateTemplateKHR',
                                  'UpdateDescriptorSetWithTemplateKHR',
                                  'CmdPushDescriptorSetWithTemplateKHR',
+                                 'BindBufferMemory',
                                  ]
         # Map APIs to functions if body is fully custom
         custom_body_dict = {'CreateInstance': self.GenReplayCreateInstance,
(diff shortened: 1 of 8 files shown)


davidlunarg:
I attempted to implement something similar to this when I first wrote this code. I found one serious problem that prevented me from doing this -- creating a new trace file by retracing an existing trace file does not work:

    vktrace -o cNew.vktrace -p vkreplay -a "-o cOld.vktrace"

This is because the addition of the vkCreateInstance during replay initialization to query the driver and gpu causes problems when it is retraced. (I can't remember the exact reason why, but I encountered this problem when I ran the code in this pull request.)

Also, this change requires that every vkAM packet in the trace file be read during startup. The vkAM packets may be scattered throughout the trace file, so many portions of the trace file needs to be pulled into memory by the file system, which can cause a delay in starting the replay. The algorithm of reading forwards and backwards from the current vkAM during replay of vkAM doesn't pull in additional portions of the trace file, since the packets that the vkAM portability code looks for tend to be near the vkAM packet in the file. I would prefer that a large portion of the trace file not be read during startup, but instead read as the packets are processed. 

I like the change of encapsulating all the direct calls of fread/ftell/fseek with FileLike.

justuswang:
@davidlunarg , thanks for pointing this problem out.  I will revert the change of moving portability table out from main_loop and create another pull request for "Encapsulate all the direct calls of fread/ftell/fseek with FileLike".

I'm going to close this pull request.

PR #291, opened 2017-09-01
toolchain: Update Windows build to track upstream changes

cnorthrop:
Updates parameters, initializes jsoncpp variables.


cnorthrop:
Only the Windows script had its options changed, breaking the build on our end (we don't build jsoncpp and the script can't evaluate options that aren't initialized).

cherry-picked to master, thanks!

PR #293, opened 2017-09-04
vkreplay: Encapsulate direct file read calls and fix alignment in vkBBM

justuswang:
* Encapsulate all the direct calls of fread/ftell/fseek with FileLike

* Do alignment for the memory size needed in vkAM call to make sure it
allocates memory with the expected size during replay.  Only works when
portability table is being used and the alignment in vkGBMR is
different in trace platform and replay platform.
(Haven't met the alignment issue in vkBIM, so leave vkBIM unchanged)

* Print useful info when user tries to replay a 32bit trace file using
64bit vkreplay and vice versa.

### DIFF
--- a/vktrace/vktrace_common/vktrace_filelike.c
+++ b/vktrace/vktrace_common/vktrace_filelike.c
@@ -63,6 +63,26 @@ BOOL vktrace_Checkpoint_read(Checkpoint* pCheckpoint, FileLike* _in) {
 // ------------------------------------------------------------------------------------------------
 // ------------------------------------------------------------------------------------------------
 // ------------------------------------------------------------------------------------------------
+size_t vktrace_FileLike_GetFileLength(FILE* fp) {
+    size_t byte_length = 0;
+
+    long length = 0;
+    if (fseek(fp, 0, SEEK_END) != 0) {
+        vktrace_LogError("Failed to fseek to the end of tracefile for replaying.");
+    } else {
+        length = ftell(fp);
+        if (length == -1L) {
+            vktrace_LogError("Failed to get the length of tracefile for replaying.");
+            length = 0;
+        }
+    }
+    rewind(fp);
+
+    byte_length = length;
+    return byte_length;
+}
+
+// ------------------------------------------------------------------------------------------------
 FileLike* vktrace_FileLike_create_file(FILE* fp) {
     FileLike* pFile = NULL;
     if (fp != NULL) {
@@ -70,6 +90,7 @@ FileLike* vktrace_FileLike_create_file(FILE* fp) {
         pFile->mMode = File;
         pFile->mFile = fp;
         pFile->mMessageStream = NULL;
+        pFile->mMemLen = vktrace_FileLike_GetFileLength(fp);
     }
     return pFile;
 }
@@ -82,6 +103,7 @@ FileLike* vktrace_FileLike_create_msg(MessageStream* _msgStream) {
         pFile->mMode = Socket;
         pFile->mFile = NULL;
         pFile->mMessageStream = _msgStream;
+        pFile->mMemLen = 0;
     }
     return pFile;
 }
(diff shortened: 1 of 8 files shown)


vktrace/vktrace_common/vktrace_filelike.c
```diff
@@ -63,13 +63,34 @@ BOOL vktrace_Checkpoint_read(Checkpoint* pCheckpoint, FileLike* _in) {
 // ------------------------------------------------------------------------------------------------
 // ------------------------------------------------------------------------------------------------
 // ------------------------------------------------------------------------------------------------
+size_t vktrace_FileLike_GetFileLength(FILE* fp) {
```
davidlunarg: A side effect of this function is that it resets the file pos to the beginning of the file. Should add a comment warning about this.

vktrace/vktrace_common/vktrace_filelike.h
```diff
@@ -37,6 +37,7 @@ typedef struct FileLike FileLike;
 typedef struct FileLike {
     enum { File, Socket } mMode;
     FILE* mFile;
+    size_t mMemLen;
```
davidlunarg: mMemLen doesn't describe what this variable holds. Maybe change it to mFileLen?

vktrace/vktrace_replay/vkreplay_vkreplay.cpp
```diff
@@ -102,6 +102,36 @@ int vkReplay::init(vktrace_replay::ReplayDisplay &disp) {
...
+                replayBitLen = "64bit";
+                break;
+            default:
+                replayBitLen = "Unknown";
+                break;
+        }
+        vktrace_LogError("%s trace file is not supported by %s vkreplay.", traceBitLen.c_str(), replayBitLen.c_str());
+        return -1;
+    }
+
```
davidlunarg: This switch statement, string constance, and call to vktrace_LogError can probably be replace with a single statement:

vktrace_LogError("%d-bit trace file is not supported by %d-bit vkreplay.", m_pFileHeader->ptrsize, m_replay_ptrsize);

vktrace/vktrace_replay/vkreplay_vkreplay.cpp
```diff
@@ -2399,37 +2422,59 @@ VkResult vkReplay::manually_replay_vkAllocateMemory(packet_vkAllocateMemory *pPa
...
+wrapItUp:
+
+    if (doAllocate) {
         replayResult = m_vkFuncs.real_vkAllocateMemory(remappedDevice, pPacket->pAllocateInfo, NULL, &local_mem.replayDeviceMemory);
     }
 
-wrapItUp:
+    if (saveFilePos) {
+        vktrace_FileLike_SetCurrentPosition(traceFile, saveFilePos);
+    }
```
davidlunarg: Current file position is not reset to saved position if return at line 2457 is executed.

vktrace/vktrace_common/vktrace_filelike.h
justuswang: Yes, mFileLen is more clear to describe this variable.  Will update it.

vktrace/vktrace_replay/vkreplay_vkreplay.cpp
justuswang: Will reset file position before the return at line 2457.  Maybe the file position should also be reset before the return at line 2374, 2396, and 2409?  Though the replay may (should?) fail when there's an error happens in vkAM.

vktrace/vktrace_replay/vkreplay_vkreplay.cpp
justuswang: Thanks!  Will update it with 

vktrace_LogError("%d-bit trace file is not supported by %d-bit vkreplay.", m_pFileHeader->ptrsize * 8, m_replay_ptrsize * 8);

to convert byte length to bit length.

vktrace/vktrace_common/vktrace_filelike.c
justuswang: Maybe it is better to save file position at the beginning and reset it before return the file length in this function?
I didn't reset the file position because this function was only called by vktrace_FileLike_create_file() which means the file position is always the beginning of the file when calling this function.  I realized that it may confuse people now.  Will update it in next commit.

justuswang:
@davidlunarg , Thanks for your comments.  I will separate the unrelated changes into different pull requests in the future.

vktrace/vktrace_common/vktrace_filelike.c
davidlunarg: Either way is fine - restore the file pos or add a comment that the file pos is changed when this function is called. I would lean toward the latter since the function is only called from one place.

vktrace/vktrace_replay/vkreplay_vkreplay.cpp
davidlunarg: Agree that the replay will probably fail if errors occur, but would be best to be consistent - makes code easier to understand.

vktrace/vktrace_common/vktrace_filelike.c
justuswang: A comment warning about the reset of file position has been added.  Please help to review.

vktrace/vktrace_replay/vkreplay_vkreplay.cpp
justuswang: Updated to reset file position to saved position before every return (at line 2352, 2375, 2389, 2438 in the latest code) to make it consistent.

davidlunarg:
Merged to master.

davidlunarg:
This pull request resulted in a problem in playing back certain trace files. Justus, could you review pull request  https://github.com/LunarG/VulkanTools/pull/297

justuswang:
Thanks David, pull request #297 solved the problem of this pull request #293 when BindBufferMemory is called before GetBufferMemoryReqs.

PR #297, opened 2017-09-08
vktrace: add GetBufferMemoryReqs call to replay of BindBufferMemory

davidlunarg:
Some apps do not call vkGetBufferMemoryRequirements before
calling vkBindBufferMemory. The replay of vkBindBufferMemory
when trace/replay platforms are not identical is dependent
on the results of vkBindBufferMemory. If the trace
replay has not done so already, vkBindBufferMemory calls
vkGetBufferMemoryReqs and saves the result.

Change-Id: Iffb2e6540cccfd89ea12d1ac23612c1222f51dc5


justuswang:
Thanks David, this pull request solved the problem introduced by my pull request #293 when BindBufferMemory is called before GetBufferMemoryReqs.

davidlunarg:
Merged to master -- commit a0f08a9f2412776dc39243add2eab1675017fadf.

PR #298, opened 2017-09-11
vktrace: add next n frames trim capture support

mizhen:
Add support to capture the next n frames after pressing a trim hotkey once.

   The original command line option for trim hotkey is “hotkey-\<hotkeyname\>”, the change in this pull request is to add support for an additional format which is " hotkey-\<hotkeyname\>-\<framecount\>".  If user starts vktrace with this type of command line option or from server mode, set “ hotkey-\<hotkeyname\>-\<framenumber\> “ to  original environment variable for hotkey, once user press hotkey, trim will automatically stop capture after specified frames (framecount) . The feature also support user press hotkey to stop the trim capture before it reach the specified frames.

XCAP-729

Change-Id: I83ee50aa4c90b656b51bb6bf05c249ff7b9ff747

### DIFF
--- a/vktrace/vktrace_layer/vktrace_lib_trace.cpp
+++ b/vktrace/vktrace_layer/vktrace_lib_trace.cpp
@@ -3096,16 +3096,42 @@ VKTRACER_EXPORT VKAPI_ATTR VkResult VKAPI_CALL __HOOKED_vkQueuePresentKHR(VkQueu
     }
 
     if (g_trimEnabled) {
+        g_trimFrameCounter++;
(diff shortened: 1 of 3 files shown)


mizhen:
Cherrypick from amd-xcap-729-a + adding help info for hotkey command line option.

davidlunarg:
Looks good. I will merge this change to master.

However, I did find a problem when I type:

    vktrace -tr F1-20

or

    vktrace -tr xyzzy

Both of the above commands should generate an error.  They don't. This problem was not introduced by this pull request, but should be addressed. I'll submit a new issue.

davidlunarg:
Merged. Closing pull request.

PR #302, opened 2017-09-13
vktrace: optimize the memory allocations used by trim feature

mizhen:
Some titles hang during trim capture due to the needed GPU
memory allocations beyond the limitation of driver. The
commit fix the problem, it change the order of the sub-process
in trim snapshot_state_tracker function to optimize the
memory allocations used by trim.

XCAP-745

Change-Id: Idb390e0404ac83080cc71bcb19567e0ae9f5d4ae


davidlunarg:
Merged to master.

PR #304, opened 2017-09-14
layersvt: DevSim 1.0.3 minor code reorgs

mikew-lunarg:
Bump version to 1.0.3

This commit is a collection of small changes that don't affect existing
functionality.  There are general code improvements, and some preparation
for new functionality in development.

Among minor code reorganizations, function renames, comments added, also:

1) Use JsonCpp value.type() rather than value.is*().
An old JsonCpp design decision (and tagged as will-not-change): isObject()
unfortunately returns TRUE for nullValue.  Therefore, use explicit test of
type() to ensure a value is really only an objectValue.  isArray() has same
issue.  Changing all type checks to avoid value.is*().

2) Reduce type checking code, since types are enforced by schema validation.

3) Rename ApplyOverrides() to GetValue().  Split the value argument into
separate parent and name arguments.

Change-Id: Id660685cdecd3b137ea24e269815045e0376dec3

### DIFF
--- a/layersvt/device_simulation.cpp
+++ b/layersvt/device_simulation.cpp
@@ -57,11 +57,12 @@ namespace {
 
 // Global constants //////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
-// When making minor changes to this layer, please increment the patch level, and be sure the new version is also updated in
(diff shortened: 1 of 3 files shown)


mikew-lunarg:
Solved a subtle bug.  This should be ready to push.

PR #307, opened 2017-09-19
vktrace: Fix endianess

justuswang:
This change fixed the return value of get_endianess() to represent the
correct endianess.

### DIFF
--- a/vktrace/vktrace_common/vktrace_trace_packet_identifiers.h
+++ b/vktrace/vktrace_common/vktrace_trace_packet_identifiers.h
@@ -255,8 +255,8 @@ typedef enum _VKTRACE_TRACE_PACKET_ID_VK {
     VKTRACE_TPI_VK_vkGetPhysicalDeviceWaylandPresentationSupportKHR = 180
 } VKTRACE_TRACE_PACKET_ID_VK;
 
-#define VKTRACE_BIG_ENDIAN 0
-#define VKTRACE_LITTLE_ENDIAN 1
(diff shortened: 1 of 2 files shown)


davidlunarg:
Yes, the determination of whether the system is big endian or little endian is incorrect. I'm not sure how that got through our code reviews.

With this change, however, traces created prior to this fix that are played back with a version of vkreplay that was built after this fix on an otherwise identical platform (same gpu, driver version, os, arch, endianess and ptrsize) will fail the platform match test at line 2200 in vkreplay_vkreplay.cpp. A possible fix for this is to also swap the definitions of the constants in vktrace_trace_packet_identifiers.h:

>  -#define VKTRACE_BIG_ENDIAN 0
 -#define VKTRACE_LITTLE_ENDIAN 1
 +#define VKTRACE_LITTLE_ENDIAN 0
 +#define VKTRACE_BIG_ENDIAN 1

With the change in the constant definitions, traces created prior to this fix that are played back with a version of vkreplay that was built after this fix on a platform that is exactly identical except for endianess will pass the platform match test when it shouldn't. I don't think two platforms exist that are identical in the above characteristics except for endianess, so I don't think changing the constants will be a problem. Am I correct on this?

justuswang:
Yes, you are correct.  I have submitted a new commit to this pull request to fix the backward compatibility.  So existing trace files won't be impacted by this pull request.

davidlunarg:
Merged to master.

Issue #296, opened 2017-09-08
DevSim should emit warning when exceeding device limits


Jasper-Bekkers:
From the README.md:

"Also, if the configuration attempts to add capabilities beyond what the actual device is natively capable of providing, then the results are undefined."

Please output a (non fatal) warning to prevent accidental mistakes.

mikew-lunarg:
Thanks for the suggestion.

This is checking against limits when configuring DevSim.  Related is [LVL 2063](https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/issues/2063) for validating against a simulated device.

mikew-lunarg:
Please see pull request #308

mikew-lunarg:
This change has been pushed to master.

PR #308, opened 2017-09-20
layersvt: DevSim 1.0.4 GH296 Add debug msgs

mikew-lunarg:
Bump version to 1.0.4

See issue https://github.com/LunarG/VulkanTools/issues/296
This depends upon pull request https://github.com/LunarG/VulkanTools/pull/304

Add some informational messages to help diagnose when values of a simulated device configuration are not compatible with the VkPhysicalDeviceLimits of the underlying actual device.  Those messages are enabled, along with other debug messages, using: export VK_DEVSIM_DEBUG_ENABLE="1"

### DIFF
--- a/layersvt/device_simulation.cpp
+++ b/layersvt/device_simulation.cpp
@@ -39,7 +39,7 @@
 #include <stdarg.h>
 #include <stdio.h>
 #include <stdlib.h>
-#include <stdint.h>
+#include <inttypes.h>
 #include <string.h>
 
 #include <functional>
@@ -62,7 +62,7 @@ namespace {
 // When making ANY changes to the version, be sure to also update layersvt/{linux|windows}/VkLayer_device_simulation.json
 const uint32_t kVersionDevsimMajor = 1;
 const uint32_t kVersionDevsimMinor = 0;
-const uint32_t kVersionDevsimPatch = 3;
+const uint32_t kVersionDevsimPatch = 4;
 const uint32_t kVersionDevsimImplementation = VK_MAKE_VERSION(kVersionDevsimMajor, kVersionDevsimMinor, kVersionDevsimPatch);
 
 const VkLayerProperties kLayerProperties[] = {{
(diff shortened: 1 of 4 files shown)


mikew-lunarg:
This passes our internal CT, but TravisCI isn't happy?

karl-lunarg:
Things have been failing since Sep 7, which is when Travis-CI upgraded their Trusty images, which included updating clang from 3.5 to 3.9.

Your failing log shows that a CMake file somewhere in glslang may be requiring clang 3.5.

It looks like CMake ran for glslang and emitted the clang version error, but the subsequent build of VT worked because the external cache had been restored, so all the needed bits where there.

But it looks like that there was a trunk merge just performed on VT and a very recent build on master succeeded.

So maybe you should pick up the latest master into your branch and see if that helps.  My best guess is that you need the latest glslang revision file in VT to pick up a glslang that doesn't have this problem.

karl-lunarg:
It is also (more) likely that this was caused by a stale Travis cache, which may have contained enough cached CMake info to "freeze" the compiler version CMake uses.   The recent VT trunk merge also contains the fix to flush the caches "ourselves" when needed, so the same suggestion applies:  Update your branch with the contents of master, which may invalidate the cached external directory.

mikew-lunarg:
rebased to latest master, we'll see what Travis says.

karl-lunarg:
OK, so that worked.  We were lucky in the sense that one of the external revisions changed and that caused the new cache-checking script to delete the external directory contents and rebuild it.

The problem is, going forward, is that if they (Travis-CI) update the C/C++ compiler again in the Trusty OS image, there will be a similar failure until a cache invalidate/flush/rebuild is triggered by the completely unrelated event of a change in the external revisions.  This particular failure, CMake not finding a specific C/C++ compiler, was benign in that a CMake error was posted, but the code didn't need to be rebuilt anyway and the overall build succeeded even though the CMake step error caused the red flag.

Hopefully the frequency of Travis-CI compiler upgrades is far less than updating external revisions, so this problem simply repairs itself quickly with the next external revision change.   I suppose the ideal solution would be to check also for C/C++ version changes as part of the caching, but I'm not inclined to worry about that, since we'd have to consider all tools used in the build in the same manner.

mikew-lunarg:
cool.  can I get reviews of this and #304?  :-)

karl-lunarg:
The commit message in this PR suggests that there is a new env var that is used to activate the debug messages, yet I see no code in this PR that appears to read this env var and/or alter behavior based on the setting of the env var.  Also, there is no mention of it in the updated doc file.   Did you decide to just leave the new messaging on all the time?   (I think it should be on all the time)

mikew-lunarg:
There is not a new envar.  
The new messages use the existing DebugPrintf() mechanism with its envar to enable the output.

PR #311, opened 2017-09-22
vktrace: add one AMD Extension support to fix replay crash

mizhen:
Some title use AMD Rasterization Order Extension on AMD hardware, it cause
replay crash due to missing support for this extension. the change add
support for this extension to fix the problem.

XCAP-759

Change-Id: Id0885686a6f169a5ce43b2c47ba3a91464cf58c1


davidlunarg:
Merged to master

PR #312, opened 2017-09-26
vktrace: Fix the incorrect bufferAddress in generateMapUnmap()

justuswang:
In generateMapUnmap() when doing trim, no need to adjust the pointer of bufferAddress
if the memory is actually mapped in generate::vkMapMemory() because the
pointer is re-populated to point to the beginning of the image memory
in generate::vkMapMemory().


davidlunarg:
@plohrmann, @mizhen : Would you be able to review this pull request?

mizhen:
Looks good to me.

davidlunarg:
Merged to master.

PR #314, opened 2017-09-27
vktrace: Fix descriptor sets recording in trim

justuswang:
This change fixes three minor issues related to recording write descriptor
sets in trim.

* Reset writeDescriptorCount for each DescriptorSet before processing
write descriptor sets in __HOOKED_vkUpdateDescriptorSets in trim. This
is to make sure writeDescriptorCount is consistent with the number of
pWriteDescriptorSets for each descriptor set in the latest call of
vkUpdateDescriptorSets for each descriptor set.

* Add check for the destination address (it should not be nullptr) when
recording ImageInfo, BufferInfo or TexelBufferView for each write
descriptor set via memcpy.

* Go through all the WriteDescriptorSets to copy their valid ImageInfo,
BufferInfo and TexelBufferView data in StateTracker::operator=().


davidlunarg:
Merged to master.

mizhen:
Could you also change copyDescriptor part in same function?

justuswang:
@mizhen, sure, I will create another pull request for copyDescriptor.

PR #315, opened 2017-09-27
vkreplay: Grant SD card read permission in vkreplay APK installation

justuswang:
Set the target SDK version to a version before 23 makes it possible for
vkreplay to grant SD card read/write permission during installation on
Android.

With this change, user will not need to manually run following commands
after vkreplay installation.

* adb shell pm grant com.example.vkreplay android.permission.READ_EXTERNAL_STORAGE
* adb shell pm grant com.example.vkreplay android.permission.WRITE_EXTERNAL_STORAGE

This has been tested on Android N and Android O with vkreplay built by Android SDK Tools v24.4.1.

### DIFF
--- a/build-android/vkreplay/AndroidManifest.xml
+++ b/build-android/vkreplay/AndroidManifest.xml
@@ -2,7 +2,7 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android" package="com.example.vkreplay" android:versionCode="1" android:versionName="1.0">
 
     <!-- This is the platform API where NativeActivity was introduced. -->
-    <uses-sdk android:minSdkVersion="23" android:targetSdkVersion="23"/>
+    <uses-sdk android:minSdkVersion="22" android:targetSdkVersion="22"/>
 
     <!-- This allows reading trace file from sdcard -->
     <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE"/>


davidlunarg:
LGTM. @cnorthrop, what do you think?

cnorthrop:
Sorry for the delay - crazy travel week.

This is an interesting find.  There aren't any known Vulkan implementations before Android-23, so I've turned a blind eye to versions before that.  Since we aren't depending on libvulkan.so being present on the device (we're using vulkan_wrapper), we could lower the requirement to Android-22 and pick up these more relaxed security checks.

FWIW, we can get the same effect by using `adb install -g <package>` (which automatically grants all requested permissions), but generally retargeting our projects to Android-22 might be an easier route forward.  I am a bit worried about any unknowns that come along with that, but I don't have any concrete reasons not to.

I can't test this at the moment due to #316, so let's get that cleared up before we merge this.

justuswang:
@cnorthrop, thanks for your comments!

`adb install -g <package>` works great on my Android platforms.  I was not aware of this can get the same effect.  Shall I close this pull request and leave the target SDK version 23 unchanged?

cnorthrop:
Yes, if that works for you, I'd rather go the conservative route (not changing the SDK version).

justuswang:
Thanks, @cnorthrop.  Will close this pull request.

cnorthrop:
It is also worth noting that we've figured out a way to avoid this requirement altogether by writing directly to the application's files on sdcard.  See this [PR #739](https://github.com/baldurk/renderdoc/pull/739) for RenderDoc.  We can probably do the equivalent for vktrace.  Would need to coordinate that with an update to our testing scripts.

PR #318, opened 2017-09-28
vktrace: Fix copy descriptor sets recording in trim

justuswang:
Reset copyDescriptorCount for each DescriptorSet before processing
copy descriptor sets in __HOOKED_vkUpdateDescriptorSets in trim. This
is to make sure copyDescriptorCount is consistent with the number of
pCopyDescriptorSets for each descriptor set in the latest call of
vkUpdateDescriptorSets for each descriptor set.


mizhen:
Looks good to me

davidlunarg:
Merged to master

PR #319, opened 2017-09-28
vktrace: fix error for trim snapshot state tracker

mizhen:
When saving tracked state for DescriptorSets in trim
starting process, it use wrong number of descriptorWrite and
descriptorCopy and cause some title crash during tracing with
trim, the change fix the problem.

XCAP-759
Change-Id: I8459328b3dd35720611ceb02fb8182b2037313b7


davidlunarg:
Merged to master

justuswang:
@mizhen , seems `writeDescriptorCount` and `copyDescriptorCount` in `struct _DescriptorSet` are no longer needed after this change.  Maybe they can be removed from `struct _DescriptorSet` in vktrace_lib_trim_statetracker.h.

Is my understanding correct?  Maybe the two "reset counter" loops I added in pull request #314 and #318 can be removed then.

PR #328, opened 2017-10-10
vktrace: fix error in vkUpdateDescriptorSets trim tracking

mizhen:
The current process cannot handle multiple bindings update in pDescriptorWrites,
the change fix the error.

XCAP-759
Change-Id: I4ba297e1171c652b59527d405bfba60e2be7a9a1

### DIFF
--- a/vktrace/vktrace_layer/vktrace_lib_trace.cpp
+++ b/vktrace/vktrace_layer/vktrace_lib_trace.cpp
@@ -1824,6 +1824,164 @@ VKTRACER_EXPORT VKAPI_ATTR VkResult VKAPI_CALL __HOOKED_vkAllocateDescriptorSets
     return result;
 }
 
+// When define DescriptorSet Layout, the binding number is also defined. by Doc,
+// the descriptor bindings can be specified sparsely so that not all binding
+// numbers between 0 and the maximum binding number. the function is used to
+// convert the binding number to binding index starting from 0.
+uint32_t get_binding_index(VkDescriptorSet dstSet, uint32_t binding) {
+    uint32_t binding_index = 0;
+    trim::ObjectInfo* pInfo = trim::get_DescriptorSet_objectInfo(dstSet);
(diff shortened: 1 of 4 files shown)


vktrace/vktrace_layer/vktrace_lib_trace.cpp
```diff
@@ -1824,6 +1824,164 @@ VKTRACER_EXPORT VKAPI_ATTR VkResult VKAPI_CALL __HOOKED_vkAllocateDescriptorSets
...
+    uint32_t binding_index = 0;
+    trim::ObjectInfo* pInfo = trim::get_DescriptorSet_objectInfo(dstSet);
+    for (uint32_t i = 0; i < pInfo->ObjectInfo.DescriptorSet.numBindings; i++) {
+        if (binding == pInfo->ObjectInfo.DescriptorSet.pWriteDescriptorSets[i].dstBinding) {
+            binding_index = i;
+            break;
+        }
+    }
+    return binding_index;
+}
```
davidlunarg: This function will return 0 if binding is not found in the descriptor set. 0 can also be a valid return value when binding is found. Is this what you want?

mizhen: yes, that mean the target app has error or something wrong with trim descriptorset tracking. now will add a commit to add assert fail and let trim ignore the binding update for release build.

vktrace/vktrace_layer/vktrace_lib_trim.h
```diff
@@ -90,6 +91,8 @@ enum enum_key_state {
 // return if hotkey triggered;
 bool is_hotkey_trim_triggered();
 
+static const uint32_t INVALID_BINDING_INDEX = std::numeric_limits<uint32_t>::max();
```
davidlunarg: Why is this defined in vktrace_lib_trim.h? It's not specific to trim functionality. Maybe it should be defined in vktrace_common.h?

davidlunarg:
Merged the first two commits, please create a new pull request for the third commit.

PR #329, opened 2017-10-11
vktrace: use 64 bit fseek and ftell on Windows

davidlunarg:
Change-Id: I1430b22d58659633312be390e9d894f1b25dc52e


davidlunarg:
Merged to master

PR #331, opened 2017-10-20
layers: Fix frame range option of screenshot layer not work on Android

justuswang:
Make "<start>-<count>-<interval>" and "all" work when taking screenshot
on Android platform via screenshot layer by doing a right strip of " ",
"\r", "\t", "\n" for property debug.vulkan.screenshot.


davidlunarg:
Merged.

PR #332, opened 2017-10-20
vktrace: Fix sequence issue of primary/secondary command buffers in trim

justuswang:
This change avoids error in vkCmdExecuteCommands when replaying trimed
trace file by making sure secondary command buffer calls always being
replayed before primary command buffer calls.

Because the secondary command buffers should be created and constructed
before calling vkCmdExecuteCommands to execute them from a primary
command buffer.


davidlunarg:
@mizhen: Could you review this pull request?

mizhen:
ok. Hi Justus, did you find the problem when you tested some title?

justuswang:
@mizhen , yes, I met a segmentation fault in vkCmdExecuteCommands when I was testing an internal Vulkan demo on Android.

The problem was caused by the secondary command buffers (used by vkCmdExecuteCommands) not created before executing vkCmdExecuteCommands().

davidlunarg:
Merged to master.

PR #335, opened 2017-10-23
Add trigger to listen on a socket.

fluxlalonde:
Use -tr port[-nFrames[-portNumber]]
The default port is 8100.
Then send a message on the port to trigger the capture; if
the message is a single integer, it overrides nFrames passed
on the command line.
The nFrames parameter is ignored when -portNumber is added.
It's convenient to use nc (netcat) to send the message:
  echo 200 | nc localhost 8100

### DIFF
--- a/vktrace/README.md
+++ b/vktrace/README.md
@@ -119,7 +119,7 @@ The vktrace tool provides the capability to trace Vulkan applications.  vktrace 
 <td>-tr &lt;string&gt;<br/>  
 ‑‑TraceTrigger &lt;string&gt;</td>
 
-<td>Start/stop trim by hotkey or frame range.<br/>String arg is one of:<br>&nbsp;&nbsp;&nbsp;&nbsp;hotkey-[F1-F12|TAB|CONTROL]<br>&nbsp;&nbsp;&nbsp;&nbsp;hotkey-[F1-F12|TAB|CONTROL]-&lt;framecount&gt<br>&nbsp;&nbsp;&nbsp;&nbsp;frames-&lt;startframe&gt;-&lt;endframe&gt</td>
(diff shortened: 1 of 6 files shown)


fluxlalonde:
Yurgh.  Windows build.  Remediating.

davidlunarg:
I haven't reviewed the code, but here's my feedback after compiling the new code using the new port option:

- The commands "vktrace -tr port-50 -p cube -o t &" followed by "echo a | nc localhost 8100" do not seem to trace 50 frames.
- "vktrace --help" output should include the new port option in the output.
- Need more details in the README.md file about how to use the new port option. I figured out how to use it by reading the commit message. That information should be in the README.md file.
- Ignoring the nframes on the command line when a port number is specified makes it not possible to specify the number of frames from the command line.
- The -tr command line option is getting complex. The solution may be to separate out the hotkey, frame, and port option into separate options, which if specified, enable trimming. I think this needs to be addressed before we add another option to the -tr option.

fluxlalonde:
Thanks David - yes, this PR is functional, but not ready.  I wanted to pump it through the build checks - is there some way to flag a build as not yet ready for review?

davidlunarg:
> Thanks David - yes, this PR is functional, but not ready. I wanted to pump it through the build checks - is there some way to flag a build as not yet ready for review?

I think adding a comment to the PR soon after you submit the PR would suffice.

fluxlalonde:
I have not addressed -tr,  but I have cleaned up the parsing significantly.  
port's nframes is now specified like with the (undocumented) frames- ',' option.  It's now specified as
"-tr port[:<portnum>][,<nFrames>]"
I still have to guard the port work against windows (I have no windows environment in which to replicate the work).

fluxlalonde:
And I failed in that upload.  Please disregard.
I'll start a new PR and copy in your comments when it's ready.
Thanks,
   Paul

PR #338, opened 2017-10-25
Add trigger to listen on a socket.

fluxlalonde:
Use -tr port[:<portnum>][,frameCount]
The default port is 8100.
Then send a message on the port to trigger the capture; if
the message is a single integer, it overrides nFrames passed
on the command line.
It's convenient to use nc (netcat) to send the message:
  echo 200 | nc localhost 8100

Only enabled on PLATFORM_LINUX.

### DIFF
--- a/vktrace/README.md
+++ b/vktrace/README.md
@@ -119,9 +119,9 @@ The vktrace tool provides the capability to trace Vulkan applications.  vktrace 
 <td>-tr &lt;string&gt;<br/>  
 ‑‑TraceTrigger &lt;string&gt;</td>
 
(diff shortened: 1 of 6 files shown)


fluxlalonde:
There's one piece of this I don't like at all:  I use printf in trim::initialize() instead calling vktrace_Log* because the Log calls cause a crash; I don't know what I should be doing differently, or if there's some initialization of vktrace_Log that I have to perform ahead of its use.  This was most vexing to track down.

fluxlalonde:
This change also doesn't address David's concern about the complexity of the -tr option.  Since hotkey, frame, and port are all exclusive it doesn't seem too painful to specify them as part of trim, though the mini-language involved is ugly.

Issue #301, opened 2017-09-12
vktrace:  Why page guard is disabled on Android?


justuswang:
Is it possible to enable page guard on Android?  Some application may not be traced correctly on Android platform without page guard.

cnorthrop:
Mostly because it hasn't been staffed.  If you have time to do it, that would be fantastic.  We can't use the Linux solution directly due to lack of `/proc/self/pagemap`.  Although that is just an optimization - should be able to get it working without.

I had some early exploration of it on the [pmb_hacks](https://github.com/LunarG/VulkanTools/tree/pmb_hacks) branch, but wasn't able to finish it.  I believe I was trying to use exceptions to detect when a page had been modified, but I was getting a ton of false positives (exceptions on pages that weren't from the application).  I believe the smoke demo was working on that branch, but not Talos Principles, the other known app that required it.

justuswang:
Thanks for your answers.

But I still have a question:
I have found that /proc/self/pagemap exists on Android system (at least on Android M).  Is it different from the /proc/self/pagemap on Linux OS or not accessible by the trace layer on Android?

justuswang:
This issue has been addressed by https://github.com/LunarG/VulkanTools/pull/397

PR #339, opened 2017-10-26
Pageguard for Android

jyangsh:
Implement pageguard for android by mprotect().

The implementation is similar to windows pageguard, but because mprotect() doesn't provide some functionality as windows, some cases may not work, e.g., tracking the read access on a mapped buffer. But it works on most android vulkan app we had.


jyangsh:
@cnorthrop, @davidlunarg, could you help review this change? This change can address issue #301. Thanks!

cnorthrop:
@jyangsh, very cool that you've worked on this, thanks!  Can you give some reasoning for why you added it to the Windows path instead of the Linux path, which is already using `sigaction`?  Granted, we can't use `/proc/pid/pagemap` on Android, but it is just an optimization in the Linux path.  It might greatly help future maintenance if it were on a shared path?

jyangsh:
@cnorthrop, actually I added the code before the linux pageguard was introduced. In that time, I did the pageguard in similar schema that is used in the windows pageguard, so I added the code in the windows code path. I don't know the linux path very well because it used the /proc/pid/pagemap, in my mind always, the mprotect() in the linux path is for an optimization.

justuswang:
Thanks @jyangsh for addressing my issue #301.

@cnorthrop , I think @jyangsh may be correct.  In my understanding, /proc/pid/pagemap is the way of doing page guard in the Linux path and mprotect() is just an optimization because it is only used when examining and clearing dirty bits in /proc/pid/pagemap.  If there's no other threads writes to the protected memory when calling getMappedDirtyPagesLinux(), mprotect() will not take any effect in the Linux path.

Seems it is very hard to share the path between Linux and Android unless change the Linux path to use mprotect() as the main way to do page guard instead of using /proc/pid/pagemap.

cnorthrop:
Okay, I must be confusing what's in the Linux path with where we wanted the Android path to go (mprotect() as the main mechanism).  @davidlunarg, what do you think?  Are you okay with the Android path going through Windows code paths?  Or could the Linux path be updated to use mprotect() first so we can align with Android?

davidlunarg:
My original reason for not using mprotect was that, in order to be POSIX compliant, it cannot be used in a signal handler. But the final code does call mprotect from a signal handler, I had to add it in order to avoid missing writes to pages.

Another reason we used pagemap on Linux is for performance -- writes to a pmb do not result in a signal and a signal handler getting called. (In fact, performance is so much better on Linux than Windows that an app like DOTA2 can write to pmb's much faster, resulting in more pmb data getting written to the file, resulting in a larger trace file. I'm not sure if the faster handling of pmb writes is a good thing or a bad thing.)

I don't know whether Android should use the Windows code path, or if the Linux path should be modified to use mprotect. I need to look at this further.

jyangsh:
@davidlunarg, is it possible to let the change is merged in firstly? Because the vulkan on android is more and more popular now, many people and companies are requesting use the vktrace to capture the vulkan games/apps. But without the pageguard, some titles cannot be captured correctly.

So could you let this change in firstly, if it take too much time to make the linux path work on android?

Thanks

davidlunarg:
Without support for /proc/pid/pagemap on Android, it seems that mprotect has be used instead. So I think we have to use this pull request. I will review it over the next 2-3 days.

Sometime in the future, I would like to prototype the trace layer to only use mprotect on Linux, to see how performance and trace file size are affected. If they are affected positively, we should consider changing the Linux path to use mprotect only.

jyangsh:
Fine, thanks!:)

jyangsh:
ping...

davidlunarg:
Heart the ping... Yes, I need to review this, but right now I have higher priority work to get done (i.e. Vulkan 1.1). Hopefully I'll get to it this week.

jyangsh:
np. :)

vktrace/vktrace_layer/vktrace_lib_pageguard.cpp
```diff
@@ -388,6 +414,29 @@ LONG WINAPI PageGuardExceptionHandler(PEXCEPTION_POINTERS ExceptionInfo) {
...
 }
+#elif defined(ANDROID)
+void PageGuardExceptionHandler(int sig, siginfo_t *si, void *unused)
+{
+    if (sig == SIGSEGV)
+    {
+        VkDeviceSize OffsetOfAddr;
+        PBYTE pBlock;
+        VkDeviceSize BlockSize;
+        PBYTE addr = (PBYTE)si->si_addr;
```
mizhen: Would it be possible to add process for reading memory and capture (CPU needed) data change coming from GPU side?

jyangsh: The  siginfo_t struct seems not provide the access type info for the exception. Now we just add the write protect to the mapped memory address, so we can always assume the fault exception is caused by the write access.

So far, we didn't meet a read case on android. Mapping a memory permanently for writing might be a normal case, but mapping a memory permanently for reading might be very rare case.

mizhen: Is there some context info can be used?  of course, it's ok if no such case on Android. That's quite different with Windows titles.

davidlunarg:
I tried out this pull request, but I made additional changes so that Linux uses the same path as Android (eliminating the pagemap path). I trace a simple pmb program (smoketest), and the replay is identical as the trace. (I verified this with screenshots.)  I hope to try it next on Dota2. It uses lots of PMB. However, there's currently a problem with Dota2 so that it doesn't trace/replay successfully (I'm working on that), so I can't test this code as well as I should before using it. I'm considering removing the pagemap path in Linux vktrace, but I need to think about the implications of that and make sure it works on Dota2.

jyangsh:
I think pagemap is better than the mprotect. Is it necessary to fall back the linux pagemap path to the android mprotect path? Though it looks very strange to use two different pathes for linux and android.

davidlunarg:
I think, but I'm not sure, that for some apps using mprotect on Linux would be better performance than pagemap. It's counter-intuitive, but what I think happens in Dota2 is that pagemap performs so much better that the app is able to make changes to PMB at a much higher rate, resulting in lots more PMB data getting written to the trace file. Eventually, the disk cache in system memory fills up, that the trace process becomes disk I/O bound. Adding more memory to the system improves trace performance, but the trace file ends up being 2X larger than on Windows. I think if we used the same mechanism for both Linux and Android, the code would be easier to read and be more maintainable. I would like to further investigate this.

davidlunarg:
I applied this patch and modified the resulting code so that Linux uses mprotect. (A new patch file is attached.) I also applied the code change from pull request #353 to fix a bug in replay. I attempted to trace Dota2 on Linux so as to compare trace performance and trace file size when using mprotect and pagemap. Unfortunately, the trace files produced with mprotect on Linux are not correct, so I wasn't able to get a performance/file size comparison. I think the the trace layer misses changes to PMB. (See comment at Line 421 in vktrace_lib_pageguardmappedmemory.cpp.) This will need to be addressed in the Android solution. (And on Linux, should we decide to use mprotect on Linux.)

(Note that I will be on vacation for a couple days, so I probably won't be looking at this pull request for the rest of this week.)

[MPROTECT_FOR_LINUX4.zip](https://github.com/LunarG/VulkanTools/files/1472043/MPROTECT_FOR_LINUX4.zip)

jyangsh:
If you decided to use the mprotect on Linux path, may I close this pull request? Once you finish the mprotect Linux path later,  and then, we can try it on Android and to make it work on Android. Is it fine?

davidlunarg:
Please don't close this pull request. I still need to determine whether pagemap or mprotect has better performance on Linux. If pagemap has better performance, I think we would want to leave the Linux implementation as is.

I'm concerned about the mprotect path missing some pmb changes as pointed out in my previous comment. The fix for that would be to use pagemap, which Android doesn't support. I'm not sure how to  solve that problem on Android.

jyangsh:
About the "comment at Line 421 in vktrace_lib_pageguardmappedmemory.cpp", the mprotect() seems not able to provide the similar function provided by GetWriteWatch() on windows. But so far, we haven't met the issue described in the comment on Android.

vktrace/vktrace_layer/vktrace_lib_pageguard.cpp
```diff
@@ -388,6 +406,27 @@ LONG WINAPI PageGuardExceptionHandler(PEXCEPTION_POINTERS ExceptionInfo) {
...
+        if (pMappedMem) {
+            uint64_t index = pMappedMem->getIndexOfChangedBlockByAddr(addr);
+
+            pMappedMem->setMappedBlockChanged(index, true, BLOCK_FLAG_ARRAY_CHANGED);
+
+            if (mprotect(pMappedMem->getMappedDataPointer() + index * pageguardGetSystemPageSize(),
+                         (SIZE_T)pMappedMem->getMappedBlockSize(index), (PROT_READ | PROT_WRITE)) == -1) {
+                vktrace_LogError("Clear memory protect on page(%d) failed !", index);
+            }
+        }
```
stroyan: If pMappedMem is NULL then this should chain to any previous signal handler.
If g_old_sa is not NULL, then it can be called to deal with the signal.
If g_old_sa is NULL, then returning will only repeat the same instruction that caused the signal.
That will loop endlessly.  It would be better to abort.

jyangsh: Thanks, patched.:)

davidlunarg:
I have pushed the changes to branch jyangsh-pageguard_android that use mprotect for implementing PMB support. I like the fact that the same code is used for Linux and Android, greatly reducing code complexity and making it much more readable/maintable.

I would like to merge this branch to master, but there's a still a few issues remaining:

- When using debug compiled code on Linux, creating a trace of Dota2 on fails intermittently - Dota2 crashes.
- When using release compiled code on Linux, create a trace of Dota2 seems to alway fail. Again, Dota2 crashes.

I tested the above two scenarios on Ubuntu 16.04 with NVIDIA GeForce 660Ti with driver 384.98.

Also, Linux PMB implementation used to use checksums to determine if a page marked dirty was in fact not dirty. (Dota2 seems to write the same data to PMB memory causing it to be marked dirty. Checksums were used to detect this condition so that pages that were marked dirty but still contained identical data were not written to the trace file. This resulted in about a 30% smaller trace file for Dota2.) Unfortunately, the code to do the checksums is more difficult to implement when using mprotect/signals and I haven't gotten it to work. I would like to get it to working again, but I don't think it is a must.

jyangsh:
Thanks David for merging the both code paths, we can help verify the code changes on android if you have any further changes in the merging. @justuswang

davidlunarg:
@jyangsh :  What testing has been done with the code in this pull request? The code seems to work except for trace and replay of Dota2 on Linux seems to have some problems - sometimes it crashes during tracing, other times rendering is incorrect in what looks like texture buffers that weren't updated. I'm not sure if these issues are cause by a trace layer problem or a Dota2 problem.

jyangsh:
We tested some unpublished benchmarks and some internal demos. Also we tested it with android game Vainglory Vulkan version.

davidlunarg:
The problem with Dota2 not tracing/replaying successfully and consistently on Linux is what is holding up this pull request. Whatever problem is causing the problem on Linux would probably manifest itself on Android, given an app that behaves the same as Dota2. I have spend some time debugging the problem on Linux, but haven't been able to isolate it. I have other responsibilities at the moment that are preventing me from looking into it further. @jyangsh: would you be able to look at this problem on Linux?

It's also possible there a a bug in Dota2 that causes the trace/replay problems, but I would like to determine that is the case before merging in this pull request.

jyangsh:
Thanks, David for looking into the Linux. Unfortunately, we also don't have enough resource to handle the x86 linux platform games. We can put it to our backlog, but we have to treat it as a very low priority.

jyangsh:
merged

PR #346, opened 2017-10-31
layersvt: DevSim 1.2.0 add new features

mikew-lunarg:
Add ability to override the results of vkGetPhysicalDeviceFormatProperties()

DevSim currently supports only Vulkan v1.0.  Check and cleanly emit an error
message if the application requests a Vulkan version other than v1.0.

Rework the test script to use 'jq' to extract sections of the test's JSON
output, sort, and reformat for consistent comparison against gold file.

Disable some of the noisier debug outputs.

Fix VisualStudio warnings

Update the .md documentation file.

Bump version to v1.2.0

Change-Id: I54deec8772056ad8649e3b0ae16a6d6c16148ba5

### DIFF
--- a/layersvt/device_simulation.cpp
+++ b/layersvt/device_simulation.cpp
@@ -62,7 +62,7 @@ namespace {
 // For any changes, at least increment the patch level.
 // When making ANY changes to the version, be sure to also update layersvt/{linux|windows}/VkLayer_device_simulation.json
 const uint32_t kVersionDevsimMajor = 1;
-const uint32_t kVersionDevsimMinor = 1;
+const uint32_t kVersionDevsimMinor = 2;
 const uint32_t kVersionDevsimPatch = 0;
 const uint32_t kVersionDevsimImplementation = VK_MAKE_VERSION(kVersionDevsimMajor, kVersionDevsimMinor, kVersionDevsimPatch);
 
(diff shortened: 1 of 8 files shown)


layersvt/device_simulation.cpp
```diff
@@ -236,6 +424,24 @@ std::mutex global_lock;  // Enforce thread-safety for this layer's containers.
 uint32_t loader_layer_iface_version = CURRENT_LOADER_LAYER_INTERFACE_VERSION;
 
 typedef std::vector<VkQueueFamilyProperties> ArrayOfVkQueueFamilyProperties;
+typedef std::unordered_map<uint32_t /*VkFormat*/, VkFormatProperties> ArrayOfVkFormatProperties;
```
mikes-lunarg: Why not use VkFormat as the key?

layersvt/device_simulation.cpp
```diff
@@ -476,7 +684,32 @@ class JsonLoader {
...
+            // Get a format structure from JSON.
+            DevsimFormatProperties devsim_format_properties = {};
+            GetValue(value, i, &devsim_format_properties);
+            // Split the JSON-acquired data into VkFormat and VkFormatProperties.
+            const VkFormat format = devsim_format_properties.formatID;
+            VkFormatProperties vk_format_properties = {};
+            vk_format_properties.linearTilingFeatures = devsim_format_properties.linearTilingFeatures;
+            vk_format_properties.optimalTilingFeatures = devsim_format_properties.optimalTilingFeatures;
+            vk_format_properties.bufferFeatures = devsim_format_properties.bufferFeatures;
+            if (IsFormatSupported(vk_format_properties)) {
```
mikes-lunarg: Does this check make it impossible to spoof an unsupported format?

layersvt/device_simulation.cpp
```diff
@@ -986,7 +1245,7 @@ VKAPI_ATTR void VKAPI_CALL GetPhysicalDeviceMemoryProperties(VkPhysicalDevice ph
     const auto dt = instance_dispatch_table(physicalDevice);
 
     PhysicalDeviceData *pdd = PhysicalDeviceData::Find(physicalDevice);
-    DebugPrintf("GetPhysicalDeviceMemoryProperties physicalDevice %p pdd %p\n", physicalDevice, pdd);
+    // DebugPrintf("GetPhysicalDeviceMemoryProperties physicalDevice %p pdd %p\n", physicalDevice, pdd);
```
mikes-lunarg: Not a huge fan of commented out code, do we need some sort of log level to deal with debug verbosity?

layersvt/device_simulation.cpp
```diff
@@ -1001,15 +1260,30 @@ VKAPI_ATTR void VKAPI_CALL GetPhysicalDeviceQueueFamilyProperties(VkPhysicalDevi
     const auto dt = instance_dispatch_table(physicalDevice);
 
     PhysicalDeviceData *pdd = PhysicalDeviceData::Find(physicalDevice);
-    DebugPrintf("GetPhysicalDeviceQueueFamilyProperties physicalDevice %p pdd %p\n", physicalDevice, pdd);
+    // DebugPrintf("GetPhysicalDeviceQueueFamilyProperties physicalDevice %p pdd %p\n", physicalDevice, pdd);
```
mikes-lunarg: See previous comment

layersvt/device_simulation.cpp
```diff
@@ -890,6 +1141,14 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateInstance(const VkInstanceCreateInfo *pCreat
                                                   dt->GetPhysicalDeviceQueueFamilyProperties(physical_device, count, results);
                                                   return VK_SUCCESS;
                                               });
+        // Query every standard VkFormat's properties.
```
mikes-lunarg: Why pre-populate pdd.array_of_format_properties if GetPhysicalDeviceFormatProperties will call down the chain on lookup failures?

layersvt/device_simulation.cpp
mark-lunarg: Yes, please kill it with fire.

mark-lunarg: Please kill it with fire.

mikew-lunarg:
"meaningful ~~emulation~~ simulation of vkGetPhysicalDeviceImageFormatProperties" :-)

The combination of the 5 input parameters suggested a large set, so I postponed doing for the first version.  Certainly if someone really needs it, I'm glad to reconsider.

layersvt/device_simulation.cpp
mikew-lunarg: My Ubuntu 16.04's gcc 5.4 doesn't like it.
Apparently it's fixed in [gcc 6.1](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=60970)

layersvt/device_simulation.cpp
mikew-lunarg: Devsim tries not to make anything impossible.  You can certainly enable a format that is not consistent with nor supported by the actual implementation.

IsFormatSupported() just checks if the format's 3 featureflags fields are all zero.  [Per the spec that format is unsupported](https://github.com/LunarG/VulkanTools/blob/mikew_devsim120/layersvt/device_simulation.cpp#L442) and Devsim doesn't bother adding to its internal container.

layersvt/device_simulation.cpp
mikew-lunarg: The lookup failure that causes it to call down the chain is if the PhysicalDevice has not been overridden.  

If the format lookup fails, an "unsupported" format (all flags==0) is returned, and it does not call down.

layersvt/device_simulation.cpp
mikes-lunarg: But if you don't add it to the internal container, won't devsim use the underlying device's format properties? What if I want to restrict the formats available to an app by specifying an unsupported format in the json?

layersvt/device_simulation.cpp
mikew-lunarg: I leave those in because they're really for my use only, and it was convenient to re-enable as needed.  There's already a boolean to enable debug messages (which was getting spammed), I thought log levels would be overkill.
Will remove.

layersvt/device_simulation.cpp
mikes-lunarg: Aha! I misread that the first time, ignore my other comment

layersvt/device_simulation.cpp
mikew-lunarg: Gone, in a poof of smoke

PR #347, opened 2017-11-04
docs: Add Android build steps

streetworld:


cnorthrop:
Steps for building and running Android versions of the tool are already in BUILDVT.md, so I don't think we need to re-merge this commit.  Let us know if something is missing though, thanks!

PR #349, opened 2017-11-07
vktrace: Fix incorrect calculation of packet size in trace

justuswang:
This change fixes an incorrect calculation of packet size "customSize"
used by CREATE_TRACE_PACKET() in __HOOKED_vkCmdPipelineBarrier()


davidlunarg:
Merged to master

PR #351, opened 2017-11-08
vktrace: fix error in reading PMB flag

mizhen:
The change fixes an error in reading PMB flag, the error cause two PMB flags
reading has conflict, the change fix the error by using different static
variable to store different flag.

Change-Id: Ifff47d2a097e30651aa78f546f52d24396b59d3b


davidlunarg:
Merged to master

PR #352, opened 2017-11-08
vktrace: fix error in trim process for staging buffer and image

mizhen:
If an image or buffer bound to memory which is not host visable, trim will
use staging buffer to copy its data, the error is in the process when try
to reuse command buffer. It cause vkBeginCommandBuffer fail, the change
fix the error.

Change-Id: I03b7ef94982a9f32db113336d23b3df42afb322f


davidlunarg:
Merged to master

PR #353, opened 2017-11-14
vktrace: xlate of handles in descriptor set replay

davidlunarg:
Replay of vkUpdateDescriptorSetWithTemplateKHR should only
translate handles that are really going to be used by the
structures in pData. Modified the code that does the
translation.

Also, modified vkCreateDescriptorSetLayout replay so it
translates the pImmutableSamplers pointer.

Change-Id: I6dde3bc770d34fce8c9bd732433fdbfcfd97bbc4


davidlunarg:
AppVeyor build failure is do to 60 minute timeout.

davidlunarg:
Merged to master

PR #355, opened 2017-11-15
layersvt:Add GPDP2 entrypoints to DevSim layer

tobine:
Add corresponding GPDP2 entrypoints for all GPDP functions that DevSim
intercepts. Just pass them through to corresponding GPDP function.


cnorthrop:
cherry-picked to master

PR #356, opened 2017-11-22
vktrace: fix error for trim starting process

mizhen:
When starting to trim, the process include dumping the content of all
tracked images and buffers to trace file. Some images or buffers might
be just created, not bound to any memory at that time. Such case make
some title run into crash if tracing with trim. The change fix the
problem.

Change-Id: I7650b18645d0493c218e11718e062de97b00a4e9


davidlunarg:
Merged to master

PR #357, opened 2017-11-22
Devsim 1.2.1 multiple input files

mikew-lunarg:
Device Simulation Layer can now load multiple configuration files by specifying a delimited list of filenames in the VK_DEVSIM_FILENAME environment variable.  The delimited list borrows the syntax of the PATH environment variable: on Windows, separate the filenames with ';', on Linux separate with ":".

Add documentation for the multiple configuration filename to the .md file.

Some minor rearrangement of comments and debug messages.

Restructure/rename the regression test input files to exercise and make use of the new list of configuration files feature.
    
The "test1" input configuration now consists of several file fragments that are all loaded at runtime to create the complete configuration.
    
This feature permits configuration files to be modular, with boilerplate data in separate files.

### DIFF
--- a/BUILDVT.md
+++ b/BUILDVT.md
@@ -255,16 +255,16 @@ Result screenshot will be in:
 /sdcard/Android/<framenumber>.ppm
 ```
 ### device_simulation
-To enable, set a property with the path of configuration file to load:
+To enable, use a setting with the path of configuration file to load:
 ```
-adb shell setprop debug.vulkan.devsim.filepath <path/to/DevSim/JSON/configuration/file>
+adb shell settings put global debug.vulkan.devsim.filepath <path/to/DevSim/JSON/configuration/file>
 ```
 Example of a DevSim JSON configuration file: [tiny1.json](https://github.com/LunarG/VulkanTools/blob/master/layersvt/device_simulation_examples/tiny1.json)
 
-Optional: set properties to enable debugging output and exit-on-error:
+Optional: use settings to enable debugging output and exit-on-error:
 ```
-adb shell setprop debug.vulkan.devsim.debugenable 1
-adb shell setprop debug.vulkan.devsim.exitonerror 1
+adb shell settings put global debug.vulkan.devsim.debugenable 1
+adb shell settings put global debug.vulkan.devsim.exitonerror 1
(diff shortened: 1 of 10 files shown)


mikew-lunarg:
RE motivation for loading multiple DevSim configuration files...

The primary reason is Vulkan extensions will have their own schemas, and so must be loaded as separate files.  This PR is groundwork for adding extension support.

Since multiple file loading is available, it seems convenient to be able to split configuration data into separate files if desired; the tests in this PR use that capability.  That modularity allows large data (e.g.: the array of VkFormats) to be shared with files containing smaller configuration tweaks.  Using that feature is optional.

cnorthrop:
I tried updating the Android test script for this [here](https://github.com/LunarG/VulkanTools/commit/d3155817c2be905dc390986e8460b45ed1ec2535).

There's a problem on Android when trying to set such a long value for filenames paths.  Since we're currently using a system property to pass it to devsim, there is a size limit of (I believe) 91 characters, so we can't support the default test case.  There are a few ways to address this:
- We could shorten the list of files for Android to essentially creating a simpler test. 
- We might be able to move to using Android Settings instead of system properties.  I'll explore that soon, but probably not going to get to it today.
- We could break up the current filename setting into a series of settings, similar to how validation layers can be specified one at a time, in order.

mew-cx:
@cnorthrop Ok, I can certainly dial-back the splitting of the configuration files, and use only 2 or 3 files to keep the name length down below 90 chars.  Probably do want to investigate a less constrained mechanism for future use.

mikew-lunarg:
@cnorthrop Per your request, I've reduced the splitting of config files to two, which should fit in your current constraint.

cnorthrop:
Thanks @mikew-lunarg but even two files is still too long, since we provide full path to each, i.e.: `/sdcard/Android/devsim_test1_in.json:/sdcard/Android/devsim_test1_in_ArrayOfVkFormatProperties.json`


I'll quickly try to get settings going instead of system props since they have no limit on size, when we can move back to the 5-ish files you had before.

cnorthrop:
I added a commit that gets things working on Android now, using settings instead of system properties to populate input files.

PR #360, opened 2017-11-28
vktrace: Add process in trim to get memory size before binding an image

mizhen:
Trim statetracker get an image memory size when target title call
vkGetImageMemoryRequirements. But for some images, target title may not
call vkGetImageMemoryRequirements before binding the image to memory if
no such need on specific condition. Such case make trim cannot dump the
image data to trace file, the change fix the problem.

XCAP-759

Change-Id: Iea5a6a75c2b1f0238305c33ee2308207049a783f


davidlunarg:
Merged to master

PR #370, opened 2017-12-29
Vulkan Layer Factory

mark-lunarg:
First swipe at getting the Layer Factory into the repo.  Just for kicks, added a Demo Layer that uses a different code structure, has more output options and uses a global pre-call function.  

For anyone surprised by this, please see README.md in the layer_factory directory.

@cnorthrop, I'd like to remove the APP_MODULES line from Application.mk in LVL, if that's cool with you -- might make the merges a bit easier and cut down on LVL source changes as well.  Please take a look at the Android changes as our coverage on VT is currently lacking.  
**UPDATE**: This was done in LVL, commit 4801ac47.

Cheers!


cnorthrop:
Looks like APP_MODULES is optional so I think it should be okay, as long as we want to build everything, which we do.

VT is broken for Android at the moment due to trunk merge, so I can't verify your changes.  Hope to have it cleared up in the next day or two.

layer_factory/demo_layer/demo.cpp
```diff
@@ -0,0 +1,80 @@
...
+        std::stringstream message;
+        message << "Memory Allocation Count: " << _number_mem_objects << "\n";
+        message << "Total Memory Allocation Size: " << _total_memory << "\n\n";
+
+        // Various text output options:
+        // Call through simplified interface to log_msg()
+        MemDemo::Information(message.str());
+
+        // Call log_msg() directly
+        log_msg(vlf_report_data, VK_DEBUG_REPORT_INFORMATION_BIT_EXT, VK_DEBUG_REPORT_OBJECT_TYPE_DEVICE_MEMORY_EXT, 0, __LINE__,
```
jzulauf-lunarg: Is there any threading protection present (or needed) for the log_msg call (which I've seen called with a lock held in other layers)

layer_factory/starter_layer/high_water_mark.h
```diff
@@ -0,0 +1,74 @@
+/*
```
jzulauf-lunarg: same comment re: naming.  I'd think a high_water_mark would track a persistent "highest ever" value...

layer_factory/starter_layer/high_water_mark.h
```diff
@@ -0,0 +1,74 @@
...
+#include <sstream>
+#include <unordered_map>
+
+static uint32_t display_rate = 60;
+
+class MemAllocLevel : public layer_factory {
+public:
+    // Constructor for interceptor
+    MemAllocLevel() : layer_factory(this) {
+        _number_mem_objects = 0;
```
jzulauf-lunarg: Google name post-pends _

layer_factory/demo_layer/demo.cpp
mark-lunarg: Nope.

PR #372, opened 2018-01-05
vktrace: addressed all compiler warnings

davidlunarg:
Change-Id: I6ebdfaecd19bbd6198cdc3f4715e7ad1583e1b56

### DIFF
--- a/vktrace/vktrace_replay/vkreplay_objmapper_class_defs.h
+++ b/vktrace/vktrace_replay/vkreplay_objmapper_class_defs.h
@@ -67,7 +67,7 @@ class gpuMemory {
         PageGuardChangedBlockInfo *pChangedInfoArray = (PageGuardChangedBlockInfo *)pSrcData;
         if (pChangedInfoArray[0].length) {
             PBYTE pChangedData = (PBYTE)(pSrcData) + sizeof(PageGuardChangedBlockInfo) * (pChangedInfoArray[0].offset + 1);
-            DWORD CurrentOffset = 0;
+            size_t CurrentOffset = 0;
             for (DWORD i = 0; i < pChangedInfoArray[0].offset; i++) {
                 if ((size_t)pChangedInfoArray[i + 1].length) {
                     memcpy(mr.pData + (size_t)pChangedInfoArray[i + 1].offset, pChangedData + CurrentOffset,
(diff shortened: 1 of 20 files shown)


vktrace/vktrace_replay/vkreplay_objmapper_class_defs.h
```diff
@@ -67,7 +67,7 @@ class gpuMemory {
         PageGuardChangedBlockInfo *pChangedInfoArray = (PageGuardChangedBlockInfo *)pSrcData;
         if (pChangedInfoArray[0].length) {
             PBYTE pChangedData = (PBYTE)(pSrcData) + sizeof(PageGuardChangedBlockInfo) * (pChangedInfoArray[0].offset + 1);
-            DWORD CurrentOffset = 0;
+            size_t CurrentOffset = 0;
             for (DWORD i = 0; i < pChangedInfoArray[0].offset; i++) {
                 if ((size_t)pChangedInfoArray[i + 1].length) {
```
karl-lunarg: While you are here, you can remove three `(size_t)` casts from this line and the two that follow.  They are not needed since  `.length` and `.offset` are both `size_t`.  Casts suggest purposeful conversions, and putting them here can be confusing.

vktrace/vktrace_replay/vkreplay_objmapper_class_defs.h
```diff
@@ -67,7 +67,7 @@ class gpuMemory {
         PageGuardChangedBlockInfo *pChangedInfoArray = (PageGuardChangedBlockInfo *)pSrcData;
         if (pChangedInfoArray[0].length) {
             PBYTE pChangedData = (PBYTE)(pSrcData) + sizeof(PageGuardChangedBlockInfo) * (pChangedInfoArray[0].offset + 1);
-            DWORD CurrentOffset = 0;
+            size_t CurrentOffset = 0;
             for (DWORD i = 0; i < pChangedInfoArray[0].offset; i++) {
```
karl-lunarg: I don't understand what this code is doing, but using a DWORD loop index with a size_t upper bound seems wrong from a type safety point of view.  But I have a hard time imagining this loop iterating enough to require a size_t, so I wonder if `pChangedInfoArray[0].offset` is really the right value to use for the loop upper bound.  As I said, I can't tell what this code is supposed to do, but I think it would be good to take a closer look at it.  It may also be worth adding a comment or two to the function to explain it.

davidlunarg:
Abandoning this pull request, resubmitting as Pull Request 379.

PR #377, opened 2018-01-09
vktrace/replay: Fix pageguard delta structure type

joey-lunarg:
Made an an assumption when cleaning up warnings, undoing change.

Change-Id: If46589d893130a9e8b7fbe6b2b45dc175a04ece7


Tony-LunarG:
Explains why only 64 bit traces were failing - LGTM

PR #385, opened 2018-01-30
VLF Parallel Build Cmake Fixes

mark-lunarg:
@karl-lunarg, this seems to work fine on Linux, but (at least on my system) will consistently fail on Windows.  I have more than a hunch that this is due to some quirk with Visual Studio and the Generated File facilities of CMake.

These two commits make the VLF cmake file a bit easier to understand and are more consistent with VT and with internet consensus.

Note: I still see multiple 'Generating layer_factory.cpp' outputs in the top of master-submodules-cleanup branch, though I do NOT see them with this branch.

### DIFF
--- a/layer_factory/CMakeLists.txt
+++ b/layer_factory/CMakeLists.txt
@@ -30,7 +30,6 @@ set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/../layers)
 set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/../layers)
 set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/../layers)
 
-
 include_directories(
     ${CMAKE_CURRENT_SOURCE_DIR}
     ${CMAKE_CURRENT_SOURCE_DIR}/../layers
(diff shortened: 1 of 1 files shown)


karl-lunarg:
I made a few minor changes and the result seems to work (better) on Windows.  I basically changed the add_custom_command to use the TARGET form instead of the OUTPUT form.  I think that CMake has an easier time tracking the dependency of a target instead of the file list in the OUTPUT form.  The result is that the layer_factory files get built only once near the start of the build, which avoids the access conflicts when they get built more than once at the same time.

But if you rebuild the built solution, it will regenerate the layer_factory files, causing the 3 factory layers to rebuild as well.  Maybe we can fix this by improving the add_custom_command to not run if the generated files already exist? 

In any case, I think this may be good enough to merge and we can address the rebuild later.

I also looked at this [article](https://cmake.org/pipermail/cmake/2008-October/024492.html) and implemented the dependency chain to serialize the build of the factory layers.  This worked, but it wasn't as satisfying as the above solution because it built the factory files for each layer (3 times).  Plus, one of the designers of CMake said that we shouldn't have to do serialization.  But I think that the thing he was missing is that the "touch" generation used in his example doesn't re-write a file like we do.

If we can't solve the rebuild problem and the serialization approach fixes the rebuild problem, we could go that way.

karl-lunarg:
This isn't part of this change, but you should remove the INSTALL_LVL_FILES checks.  These checks are supposed to be only in LVL CMake files to prevent the installation of LVL content when building VT content, which includes LVL content by default.  Since this CMakefile is in a VT-only folder, it does not build any LVL content and should never appear here.

karl-lunarg:
Looks like my commit broke the Linux build.  I reverted it.

mark-lunarg:
Consolidated the useful parts of the previous set of commits, added the dependency chaining mods, cleaned up and squashed, and pushed for CI.

PR #388, opened 2018-02-01
Move VulkanTools temporary & worker targets into a solution explorer subfolder

mark-lunarg:
This unclutters the crazy list of targets in the MSVC solution explorer and has no effect on functionality, so that in MSVC you'll wind up with a folder named CMakePredefinedTargets (containing ALL_BUILD and ZERO_CHECK (!)), and a new one called vt_cmake_targets holding all the crud such as:

*json-copy
*-def-file
generate*

If this flies, I'll do a similar one for LVL.


mikew-lunarg:
Thank you for decrudding.

PR #391, opened 2018-02-02
Fix demo_layer cmake failure

mark-lunarg:
Parallel build cmake changes broke VLF layers with *.cpp and header files.

### DIFF
--- a/layer_factory/CMakeLists.txt
+++ b/layer_factory/CMakeLists.txt
@@ -32,13 +32,9 @@ set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/../layers)
 
 include_directories(
     ${CMAKE_CURRENT_SOURCE_DIR}
-#    ${CMAKE_CURRENT_SOURCE_DIR}/../layers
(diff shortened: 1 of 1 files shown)


layer_factory/CMakeLists.txt
```diff
@@ -58,38 +54,12 @@ else()
...
-if (WIN32)
-    add_library(VkLayer_utils_vlf STATIC ${VKLAYER_UTILS_VLF_SOURCES})
-else()
-    add_library(VkLayer_utils_vlf SHARED ${VKLAYER_UTILS_VLF_SOURCES})
-    if(INSTALL_LVL_FILES)
-        install(TARGETS VkLayer_utils_vlf DESTINATION $CMAKE_INSTALL_LIBDIR})
-    endif()
-endif()
-add_dependencies(VkLayer_utils_vlf generate_helper_files)
+add_dependencies(VkLayer_utils generate_helper_files)
```
karl-lunarg: Do you need to add this dependency?

VkLayer_utils is a target that is built within LVL and so its dependencies should (already) be specified there.

Below, your layer targets defined in this file specify VkLayer_utils as their own dependencies, which is desired.

mark-lunarg: Probably not, I'll pull it and retest.

PR #393, opened 2018-02-05
apidump : Display object name with handle

shannon-lunarg:
This change addresses LVL github issue [#2107](https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/issues/2107).

VK_LAYER_LUNARG_api_dump layer now displays object handle and name if set.


shannon-lunarg:
@lenny-lunarg Yup! I made this change locally about a month ago but never pushed it up.. Woops. It's changed now.

PR #394, opened 2018-02-07
Android fixes

cnorthrop:
Screenshot has been broken since end of December merge from LVL.  Found time to track it down today.  The switch from gnustl_static to c++_static STL libraries exposed a bug where a pointer to contents of a local string were being returned.  Writing the value back to our global envvar gets things working again.

Also, updated some scripts that were missed during move to submodules.

vktrace/vkreplay and screenshot are working now.  devsim has one difference I'll work with @mikew-lunarg  on.


cnorthrop:
cherry-picked to master

PR #398, opened 2018-02-13
via: Quiet annoying GCC warning message

mikew-lunarg:
For gcc version 5.4.0 of Ubuntu 16.04.3 LTS, resolve this compilation message:
warning: ignoring return value of ‘int chdir(const char*)’, declared with attribute warn_unused_result [-Wunused-result]

Change-Id: Idcc2f867d997d96e28c4109e62f9efbd61eca87b


MarkY-LunarG:
Thanks @mikew-lunarg and @davidlunarg , I was too busy on other things to notice this.

PR #399, opened 2018-02-13
tests: Add vlf_test to travis CI

mark-lunarg:
Also updated linux script for travis script invocation and normalized test output with other tests.

Took a look at the android versions of the test scripts and decided to leave that to The Experts.


cnorthrop:
Ack this.  Will get an Android version spun up soon-ish.  Thanks!

PR #405, opened 2018-02-23
tests: Add vktrace/replay regression for Windows

joey-lunarg:
Add vktracereplay.py. For now, this will handle scanning a directory of
old traces, tracing replays of them, and playing them back.

Change-Id: I3fe380dbdb00b72f0eecff83bceccfe3269dafb4


davidlunarg:
Could you add a file documenting how to use the tests directory? There now exists three vktracereplay test scripts (.bat, .sh, .py). When are they used? How do you set up a directory of trace files to trace/replay? How do you add additional tests? How do you run the tests? How do you know if they passed?

PR #415, opened 2018-03-01
vkreplay: Remap memoryOffset for vkBindImageMemory

justuswang:
This change remaps the memoryOffset in vkBindImageMemory when the
aligment on replay platform is different from the trace platform.

It only takes effect when portability table is being used and the
alignment in vkGBMR on replay platform is different from the
trace platform.


davidlunarg:
cherry-picked to master

PR #417, opened 2018-03-05
Assistant Layer: Add check for too many queues

mark-lunarg:
Added an interceptor to the assistant layer which warns if an app specifies a queueCount > 1 while also specifying a sharing mode of EXCLUSIVE. See github issue #2137.


layer_factory/assistant_layer/exclusive_qfi.h
```diff
@@ -0,0 +1,86 @@
+/*
+ * Copyright (c) 2015-2018 Valve Corporation
```
jzulauf-lunarg: Since you just wrote this, shouldn't the copyright just be 2018?

mark-lunarg: Heh, I wrote the copyright message in 2015.

PR #420, opened 2018-03-08
vkreplay: Detect space key during replay and pause/restart

davidlunarg:
Also, detect escape key and abort replay.

Added for Windows only, these features were already implemented on Linux.

Change-Id: If89af4f7ae18b6686ad5cc3e8d2f3ef798e81049


davidlunarg:
Merged to master.

PR #425, opened 2018-03-16
scripts: Rename xml generator script

mark-lunarg:
The name was carried over from the LVL repo, but this file is distinct and contains content unrelated to that file.

This shouldn't get pushed until after the 1.1 rebase!


mark-lunarg:
Rebased and pushed to Master.

PR #426, opened 2018-03-21
vktrace: Fix aligment fault on ARMv8 platform

justuswang:
This change fixes aligment fault when replaying a 32-bit trace file
on ARMv8 platform by making sure the additional data are stored with
64-bit aligned initial offset.

There's a possibility that a 32-bit GPU driver on ARMv8 platform
using instruction "vst1.64" which requires 64-bit alignment.

### DIFF
--- a/vktrace/vktrace_common/vktrace_trace_packet_utils.c
+++ b/vktrace/vktrace_common/vktrace_trace_packet_utils.c
@@ -174,7 +174,7 @@ vktrace_trace_packet_header* vktrace_create_trace_packet(uint8_t tracer_id, uint
                                                          uint64_t additional_buffers_size) {
     // Always allocate at least enough space for the packet header
     uint64_t total_packet_size =
-        ROUNDUP_TO_4(sizeof(vktrace_trace_packet_header) + ROUNDUP_TO_8(packet_size) + additional_buffers_size);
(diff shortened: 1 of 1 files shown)


davidlunarg:
This change rounds up packet size to a multiple of 8. Other places in the code rounds up to multiple of 4. Do those need to be changed too?

justuswang:
No need to change other code according to what I have tested and observed for now.

This change only makes sure the starting address of the additional data and the starting address of each packet in the trace file to be 8-byte aligned.  I think this is the minimum change to fix the alignment fault on ARMv8 platform without any uncertain side effect.

For the other places, the 8-byte aligned data's alignment (e.g. struct with uint64_t member will be recorded as 8-byte aligned on ARMv8 platform.) will not be changed by the ROUNDUP_TO_4, so there should not be a need to change other places to ROUNDUP_TO_8.

I have thought of if there are multiple additional buffers with different alignment requirements.  (e.g. the first additional buffer requires 4-byte alignment but the second one requires 8-byte alignment.)  I think it won't be a problem because ARMv8 platform supports unaligned access and the compiler won't generate vst1.64 in that condition according to what I have observed.

davidlunarg:
Merged to master

PR #429, opened 2018-03-23
docs: Move docs from LunarHub-Docs to repos

jeremyk-lunarg:
Moved the following documentation from
LunarHub-Docs to the VulkanTools repo:

api_dump_layer.md
monitor_layer.md
screenshot_layer.md
vktrace.md


karl-lunarg:
api_dump_layer.md needs to be here.

davidlunarg:
vktrace.md looks good

jeremyk-lunarg:
@mark-lunarg, I don't think I'm supposed to include SDK documentation at all.  I think it's supposed to stay in the LunarHub-Docs repo and used only for SDKs.

@karl-lunarg, thanks for the correction.  Fixed it.

cnorthrop:
There are Android usage instructions in [BUILD.md](https://github.com/LunarG/VulkanTools/blob/master/BUILD.md) that should be migrated to these docs I guess?

jeremyk-lunarg:
@cnorthrop, I'm not sure what you are asking.

cnorthrop:
I don't know where these docs are coming from, but if they are becoming canonical documentation, they should include directions for all supported platforms.

jeremyk-lunarg:
These are coming from a LunarG private repo that we use for managing documentation for our SDKs.  Recently we have realized that some of this documentation would be better off in the repos that contain what they are documenting as they are more general and less SDK focused.  There is no Android specific documentation in our private repo. If you have Android documentation you think should be in this repo please add it in another PR.

cnorthrop:
> If you have Android documentation you think should be in this repo please add it in another PR.

There is already documentation for Android in this repo.  It lives in [BUILD.md](https://github.com/LunarG/VulkanTools/blob/master/BUILD.md), but needs a better home.

Actually, it looks like the vktrace documentation you are bringing over is nearly identical to the existing vktrace documentation in its [README.md](https://github.com/LunarG/VulkanTools/blob/master/vktrace/README.md)

Follow ups (or maybe a sanity pass on documentation for the repo, similar to what was done in LVL) will be needed.

karl-lunarg:
@cnorthrop is right. 

* Since the docs coming over from the private repo were mostly intended only for the SDK, they need to be reviewed to make sure that they are not exclusively SDK-centric.  That means either carefully change them so that they agnostic to the SDK or include SDK-specific and repo-specific instructions.

* Looks like some of the docs contain both WIndows and Linux specific sections, even though they are being shipped in Windows-specific and Linux-specific SDKs.  It then sort of makes sense to pull the Android-specific info from the BUILD.md (which was probably not the greatest place) into each doc.  This has the side-effect of all SDK's seeing instructions for other platforms, but I guess we go that way for now.

* And yes, need to check for collisions with existing docs, as already pointed out.

* The VLF files that were removed should be here too, but they need to be modified to either be agnostic to the SDK or include instructions for using both in the SDK and repo.  I tend to prefer the latter.

It would probably be better to resolve all this before merging.

jeremyk-lunarg:
Thank you for the clarification.  I have moved parts of the Android section in BUILD.md into their respective locations.

None of the markdown files I added seem to mention the SDK at all.

@davidlunarg informed me that the vktrace.md file was the updated version so I removed the vktrace README.md and replaced it with vktrace.md.

The README.md file in layer_factory/ looks like it is already an agnostic version of the vlf documentation.

davidlunarg:
There's a short section in VulkanTools/BUILD.md that describes how to run vktrace/replay on Android. Should this section be moved into its own file in VulkanTools/vktrace? There's a comment that it "is preliminary and needs to be beefed up". Cody, can this comment be removed, or are changes needed?

cnorthrop:
@jeremyk-lunarg, thanks for bringing the Android steps over, looks good!

@davidlunarg, those steps appear to still be correct, although they could still be beefed up and updated a bit.  The build commands can remain, but the remaining usage bits can be brought over along with the other vktrace/replay steps, so all the vktrace Android usage documentation lives together.  I can do that as a follow up, since the text there is a bit mixed build/usage.

vkjson_info usage is still in BUILD.md.  I think it could also be split out into its own file, just like the other tools.  Then we're down to just build steps in BUILD.md.

jeremyk-lunarg:
@cnorthrop, vkjson_info seems to live in LVL.  If I put the Android usage instructions for vkjson_info into its own file, where do you think a good place to put it would be?

cnorthrop:
Ah, okay.  It looks like the LVL `BUILD.md` file still has usage steps for every platform, and contains a snippet for `vkjson_info`. 

I'd say move the bits in LVL to `libs/vkjson/README.md`, then when it makes it downstream to VT, copy over any extra info and then remove it from `BUILD.md`.

jeremyk-lunarg:
Due to the urgency of these changes, if there are no further comments by 11:00PM tomorrow I will assume the PR is acceptable in its current state and merge it.

PR #431, opened 2018-03-27
vkreplay: Add option to pre-load trace file

justuswang:
This change adds a new option "-pltf" to help user to pre-load trace
file to memory when running vkreplay. (with "-pltf true")

It helps to replay a trace file faster especially on a system with slow
file I/O.

But the pre-load requires NumLoops to be 1 because the memory
pre-loaded with a trace file may be overwritten during replay in the
first loop which will cause problem in the second or later loops.

### DIFF
--- a/vktrace/vktrace_common/vktrace_filelike.c
+++ b/vktrace/vktrace_common/vktrace_filelike.c
@@ -69,7 +69,7 @@ uint64_t vktrace_FileLike_GetFileLength(FILE* fp) {
     // Get file length
     int64_t length = 0;
     if (Fseek(fp, 0, SEEK_END) != 0) {
-        vktrace_LogError("Failed to fseek to the end of tracefile for replaying.");
+        vktrace_LogError("Failed to Fseek to the end of tracefile for replaying.");
     } else {
         length = Ftell(fp);
         if (length == -1L) {
@@ -88,14 +88,57 @@ uint64_t vktrace_FileLike_GetFileLength(FILE* fp) {
 }
 
 // ------------------------------------------------------------------------------------------------
-FileLike* vktrace_FileLike_create_file(FILE* fp) {
+FileLike* vktrace_FileLike_create_file(FILE* fp, BOOL preload) {
     FileLike* pFile = NULL;
     if (fp != NULL) {
         pFile = VKTRACE_NEW(FileLike);
-        pFile->mMode = File;
-        pFile->mFile = fp;
-        pFile->mMessageStream = NULL;
+        pFile->mFile = NULL;
         pFile->mFileLen = vktrace_FileLike_GetFileLength(fp);
+        pFile->mMessageStream = NULL;
+        pFile->mMemAddr = NULL;
+        pFile->mMemCurrAddr = NULL;
+
+        if (pFile->mFileLen == 0) {
+            vktrace_LogError("Failed to create FileLike, file length is 0!");
+            return pFile;
+        }
+
+        if (preload) {
+            vktrace_LogDebug("Preload trace file to memory.");
+            pFile->mMode = Memory;
+            char* addr = NULL;
+
+#if !defined(WIN32)
+            // mmap does not work on Windows
+            addr = (char*)mmap(NULL, pFile->mFileLen, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_POPULATE, fileno(fp), 0);
+            if (addr == MAP_FAILED) {
+                vktrace_LogError("Failed to mmap tracefile for replaying.");
+                addr = NULL;
+            }
+#else
+            // TODO: Find a better solution to make it work with a large trace file
+            addr = VKTRACE_NEW_ARRAY(char, pFile->mFileLen);
+            if (addr == NULL) {
+                vktrace_LogError("Failed to malloc memory to load tracefile for replaying.");
+            } else {
+                if (1 != fread(addr, (size_t)pFile->mFileLen, 1, fp)) {
+                    if (ferror(fp) != 0) {
+                        perror("fread error");
+                    } else if (feof(fp) != 0) {
+                        vktrace_LogVerbose("Reached end of file.");
+                    }
+                    VKTRACE_DELETE(addr);
+                    addr = NULL;
+                }
+            }
+#endif
+            pFile->mMemAddr = addr;
+            pFile->mMemCurrAddr = addr;
+        } else {
+            vktrace_LogDebug("Do not preload trace file.");
+            pFile->mMode = File;
+            pFile->mFile = fp;
+        }
     }
     return pFile;
 }
@@ -109,8 +152,22 @@ FileLike* vktrace_FileLike_create_msg(MessageStream* _msgStream) {
         pFile->mFile = NULL;
         pFile->mMessageStream = _msgStream;
         pFile->mFileLen = 0;
+        pFile->mMemAddr = NULL;
+        pFile->mMemCurrAddr = NULL;
     }
     return pFile;
+}
+
+// ------------------------------------------------------------------------------------------------
+void vktrace_FileLike_free(FileLike* pFileLike) {
+    if (pFileLike->mMemAddr) {
+#if !defined(WIN32)
+        munmap(pFileLike->mMemAddr, pFileLike->mFileLen);
+#else
+        vktrace_free(pFileLike->mMemAddr);
+#endif
+    }
+    vktrace_free(pFileLike);
 }
 
 // ------------------------------------------------------------------------------------------------
@@ -131,22 +188,38 @@ uint64_t vktrace_FileLike_Read(FileLike* pFileLike, void* _bytes, uint64_t _len)
 // ------------------------------------------------------------------------------------------------
 BOOL vktrace_FileLike_ReadRaw(FileLike* pFileLike, void* _bytes, uint64_t _len) {
     BOOL result = TRUE;
-    assert((pFileLike->mFile != 0) ^ (pFileLike->mMessageStream != 0));
+    assert((pFileLike->mFile != 0) ^ (pFileLike->mMessageStream != 0) ^ (pFileLike->mMemAddr != NULL));
 
     switch (pFileLike->mMode) {
         case File: {
-            if (1 != fread(_bytes, (size_t)_len, 1, pFileLike->mFile)) {
-                if (ferror(pFileLike->mFile) != 0) {
-                    perror("fread error");
-                } else if (feof(pFileLike->mFile) != 0) {
-                    vktrace_LogVerbose("Reached end of file.");
+            if (pFileLike->mFile != NULL) {
+                if (1 != fread(_bytes, (size_t)_len, 1, pFileLike->mFile)) {
+                    if (ferror(pFileLike->mFile) != 0) {
+                        perror("fread error");
+                    } else if (feof(pFileLike->mFile) != 0) {
+                        vktrace_LogVerbose("Reached end of file.");
+                    }
+                    result = FALSE;
                 }
+            } else {
                 result = FALSE;
             }
             break;
         }
         case Socket: {
             result = vktrace_MessageStream_BlockingRecv(pFileLike->mMessageStream, _bytes, _len);
+            break;
+        }
+        case Memory: {
+            if (pFileLike->mMemAddr == NULL) {
+                result = FALSE;
+            } else if (pFileLike->mMemCurrAddr + _len > pFileLike->mMemAddr + pFileLike->mFileLen) {
+                vktrace_LogVerbose("Reached end of file.");
+                result = FALSE;
+            } else {
+                memcpy(_bytes, pFileLike->mMemCurrAddr, _len);
+                pFileLike->mMemCurrAddr += _len;
+            }
             break;
         }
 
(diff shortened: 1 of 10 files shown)


davidlunarg:
Here my feedback on this pull request. These points are intended as a springboard for discussion, so don't go off and implement them right away, let's come to an agreement on them first.

The -pltf and --PreloadTraceFile options shouldn't need a boolean argument. The presence of the option should be sufficient to turn on file preloading.

Maybe use -p instead of -pltf?

There's a large delay from when vktrace is invoked until it starts to draw because it is preloading the file. Printing a message such as "Preloading trace file..." could indicate why the delay was occurring.

I wasn't able to initially quantify the performance improvement when the --pltf option is enabled. This might make a case for implementing an option that prints out the frames per second for which time starts after the file is preloaded. The performance improvement would then be obvious.

Would it be difficult to remove the restriction that the replay can only loop one time?

vktrace/vktrace_common/vktrace_filelike.c
```diff
@@ -88,14 +88,57 @@ uint64_t vktrace_FileLike_GetFileLength(FILE* fp) {
...
-        pFile->mFile = fp;
-        pFile->mMessageStream = NULL;
+        pFile->mFile = NULL;
         pFile->mFileLen = vktrace_FileLike_GetFileLength(fp);
+        pFile->mMessageStream = NULL;
+        pFile->mMemAddr = NULL;
+        pFile->mMemCurrAddr = NULL;
+
+        if (pFile->mFileLen == 0) {
+            vktrace_LogError("Failed to create FileLike, file length is 0!");
```
davidlunarg: Avoid using a term like "FileLike". This is a concept internal to the vkreplay code, give an error message that someone that isn't familiar with vkreplay internals can use to identify the problem.

vktrace/vktrace_common/vktrace_filelike.c
```diff
@@ -88,14 +88,57 @@ uint64_t vktrace_FileLike_GetFileLength(FILE* fp) {
...
+            // TODO: Find a better solution to make it work with a large trace file
+            addr = VKTRACE_NEW_ARRAY(char, pFile->mFileLen);
+            if (addr == NULL) {
+                vktrace_LogError("Failed to malloc memory to load tracefile for replaying.");
+            } else {
+                if (1 != fread(addr, (size_t)pFile->mFileLen, 1, fp)) {
+                    if (ferror(fp) != 0) {
+                        perror("fread error");
+                    } else if (feof(fp) != 0) {
+                        vktrace_LogVerbose("Reached end of file.");
```
davidlunarg: Why would this condition occur?

vktrace/vktrace_common/vktrace_filelike.c
justuswang: Sure.  That makes sense, will update the message to "Failed to read file, file length is 0!"

vktrace/vktrace_common/vktrace_filelike.c
justuswang: It was added to handle empty file, but yes, the 0 file length is already handled by line 101 "if (pFile->mFileLen == 0)", I will remove this condition check.

justuswang:
> The -pltf and --PreloadTraceFile options shouldn't need a boolean argument. The presence of the option should be sufficient to turn on file preloading.

I think a boolean argument is needed to make the usage of vkreplay options consistent with vktrace options.  Both "-ptm, --PrintTraceMessages <BOOL>" and "-P, --PMB <BOOL>" options of vktrace need a boolean argument.

> Maybe use -p instead of -pltf?

I think "-p" may cause some confusion because there's already a "-p" option in vktrace which specifies "The program to trace".
The options exist/shared in both vkreplay and vktrace are "-o", "-s", "-sf", and "-v" and each of them is identical in both vkreplay and vktrace.

> There's a large delay from when vktrace is invoked until it starts to draw because it is preloading the file. Printing a message such as "Preloading trace file..." could indicate why the delay was occurring.

There's a message added to vktrace_FileLike_create_file() printing debug message "Preload trace file to memory.".  Looks like "Preloading trace file..." is better.  Will update the message and change it to use vktrace_LogAlways() instead of vktrace_LogDebug().

> I wasn't able to initially quantify the performance improvement when the --pltf option is enabled. This might make a case for implementing an option that prints out the frames per second for which time starts after the file is preloaded. The performance improvement would then be obvious.

That makes sense, maybe no need to add another option.  Maybe always print the frame per second is very useful and won't disturb user.  I will create another pull request to print frame per second.

> Would it be difficult to remove the restriction that the replay can only loop one time?

Yes, this is because currently lots of replay APIs (both auto generated and manually added) are using elements in pPacket as parameters not only for setting configurations but also getting data back.  This makes it difficult to remove the restriction.
There's a workaround though: 

- NumLoops == 1: Load the file to memory at the begining.  Change current memory address pointer when reading a new packet.  (current pre-load method)

- NumLoops > 1: Load the file to memory at the beginning.  Allocate memory for each packet, memcpy data from the pre-loaded memory to the packet's memory when reading a new packet, and free it after a packet is finished replaying.  (another pre-load method which is slower than the current one)

Do you think the workaround is acceptable to remove the restriction?  This may cause the performance not consistent when user replays the same trace file for a single loop or multiple loops.

davidlunarg:
Agree on that -pltf should be used rather than -p to avoid confusion, and that it should have a boolean argument to be consistent with other options.

Thanks for submitting the fps pull request.

I overlooked the fact that replay modifies packets when it replays them, so I see the difficulty in implementing -pltf with loop > 1. However, the workaround you suggested would work, and I suspect if it was always used, the difference in performance would not be detectable. Might be worth prototyping this with the fps pull request in place to measure the difference.

A concern I have is what if the copy of the trace file doesn't fit in memory? On a system without vm, the read of the trace file would fail. But on a system with vm, the copy of the trace file would get swapped out to disk by the vm subsystem, so the option would be a no-op, and the user may not realize this is happening. The problem could be solved by locking the memory for the trace file using mlock() on Linux and VirtualLock() on Windows. If the lock fails, print an error message and exit.

justuswang:
> However, the workaround you suggested would work, and I suspect if it was always used, the difference in performance would not be detectable. Might be worth prototyping this with the fps pull request in place to measure the difference.

If the workaround is always used no matter the loop number is 1 or more than 1, I agree there should not be any detectable difference in performance.  But the performance may not be improved a lot since each packet in the trace file will go through malloc, memcpy, and free processes.

justuswang:
> A concern I have is what if the copy of the trace file doesn't fit in physical memory? On a system without vm, the read of the trace file would fail. But on a system with vm, the copy of the trace file would get swapped out to disk by the vm subsystem, so the option would be a no-op, and the user may not realize this is happening. The problem could be solved by locking the memory for the trace file using mlock() on Linux and VirtualLock() on Windows. If the lock fails, print an error message and exit.

I have tried with mlock() on linux but it needs to run vkreplay as root to make it work, otherwise it will return error ENOMEM "Cannot allocate memory".  Any suggestions?

davidlunarg:
We can't require that the user be root to run vkreplay, so my mlock idea won't work.

But I think we need to make sure we replay the file from ram and that it is not being played back from swapped out pages.

An idea we came up with is to create a ram disk and play back the trace file from there. I patched in the code from your print fps pull request into my copy of this patch and replayed traces of cube, hologram, and vkquake from a SATA hard drive, from a ram disk, and using -pltf. I made sure I cleared the disk cache between each run using "echo 3 >/proc/sys/vm/drop_caches". Here's the results I got:

vkreplayresults

It looks like there is not much difference in performance beween using a ram disk vs using the -pltf option. The ram disk guarantees that the trace file is in memory, and would allow the looping option to be used.

Might ram disk work as a solution?

justuswang:
I haven't tried it on PC, but looks like the ram disk works as a solution on PC from your results.

But on an Android platform I have tried (file system is on NAND flash).  It has some performance benefit when using -pltf option.

Here's the output of replaying an internal Vulkan application trace: (I have run each of the scenarios below 3 times and the fps looks similar, so only pasted one of those 3 runs.)

- Do NOT use -pltf option:
`27.714030 fps, 46.077744 seconds, 1277 frames, framerange 0-1276`

- Use -pltf option:
`30.294061 fps, 42.153477 seconds, 1277 frames, framerange 0-1276`

It has about 8.5% improvement using -pltf option on the trace file I have tried.  Maybe the -pltf option has more benefit on some platform than others.

davidlunarg:
A few questions...

The your test case on Android was run from a NAND file system. Is it possible to create a ramdisk in main memory on Android in the same way it is done on Linux and then copy the file to the ramdisk?:

>\# mkdir /tmp/ramdisk; chmod 777 /tmp/ramdisk
>\# mount -t tmpfs -o size=256M tmpfs /tmp/ramdisk

This ties up main memory with a ramdisk, but allocating a big chunk of memory within vkreplay does that too. The tmpfs filesystem should be available on Android, or am I mistaken?

Did you try more than one workload?

Do you know for sure that when you ran with -pltf that the whole file fit in memory and wasn't swapped out? I guess this wouldn't be an issue if the kernel on Android did not have have CONFIG_SWAP enabled.

justuswang:
> The your test case on Android was run from a NAND file system. Is it possible to create a ramdisk in main memory on Android in the same way it is done on Linux and then copy the file to the ramdisk?

I have tried to create a ramdisk in main memory on Android but vkreplay is not able to read files in the directory mounted with tmpfs.  Following are the commands I have tried:

```
# On Android:
mkdir /sdcard/ramdisk; chmod 777 /sdcard/ramdisk
mount -t tmpfs -o size=968M tmpfs /sdcard/ramdisk
# On host:
adb push test.vktrace /sdcard/ramdisk/
adb shell am start -a android.intent.action.MAIN -c android-intent.category.LAUNCH -n com.example.vkreplay/android.app.NativeActivity --es args "-v\ full\ -o\ /sdcard/ramdisk/test.vktrace"
```

but the file is not able to be opened with output: `12-15 12:20:10.046  3899  3914 E vkreplay: Cannot open trace file: '/sdcard/ramdisk/test.vktrace'.`
And the error I got via `strerror(errno)` was `No such file or directory`.

The same trace file can be replayed successfully when I was replaying it in the same directory with the same permissions but NOT mount tmpfs to that directory.


> Did you try more than one workload?
Do you mean try it with more trace files?  Yes, I have tried with several trace files I currently have and the benefits are different among different trace files (large trace files benefit more than smaller trace files).  e.g. 6.846912 fps without -pltf option and 7.035109 fps with -pltf option which is about 2.7% improvement, the trace file size is 82% of the one in my previous comment.

 

> Do you know for sure that when you ran with -pltf that the whole file fit in memory and wasn't swapped out? I guess this wouldn't be an issue if the kernel on Android did not have have CONFIG_SWAP enabled.
I'm sure won't be swapped out on the default Android configuration.  The kernel on Android does have CONFIG_SWAP=y but the swap is not enabled by default.  Confirmed by `cat /proc/meminfo` and got:
```
SwapTotal:             0 kB
SwapFree:              0 kB
```
Also found an answer here: https://stackoverflow.com/a/11274096

karl-lunarg:
@justuswang  There may be some value in using a ramdisk instead of complicating `vktrace` with this `-pltf` support.  The `-pltf` code is accomplishing basically the same thing as using a ramdisk.  Yes, I know that workflows are different and `-pltf` may be slightly more convenient in that you don't have to set up the (right-sized) ramdisk beforehand.  But `-pltf` has the problem of not supporting vktrace functionality that requires modifying the input stream (looping).

I hope that you can figure out that apparently minor problem with your ramdisk test on Android.  It might be a problem related to [this](https://stackoverflow.com/questions/24012942/copying-files-into-a-tempfs-in-android-native).  Perhaps there was a permission problem and you need to set a `umask` when mounting.

In any case, I think it would be good if we could demonstrate that performance of the ramdisk solution is very close to the `-pltf` option without any of the drawbacks.  Then we should consider not implementing `-pltf`.

justuswang:
Looks like the problem I have met was related to the directory I was using.  I switched to `/data/local/ramdisk` and it works.  Following are the commands I have tried:

```
adb root
adb shell setenforce 0
adb shell mkdir /data/local/ramdisk
adb shell mount -t tmpfs -o size=968M tmpfs /data/local/ramdisk
adb push test.vktrace /data/local/ramdisk/
adb shell am start -a android.intent.action.MAIN -c android-intent.category.LAUNCH -n com.example.vkreplay/android.app.NativeActivity --es args "-v\ full\ -o\ /data/local/ramdisk/test.vktrace"
```

The result 29.924793 fps is similar to the result of using -pltf option 30.294061 fps.

I think it is OK to use the ramdisk to do pre-load instead of using -pltf option.

karl-lunarg:
Yes, the ramdisk solution does have some additional overhead compared to `-pltf`, but in this case it was only about 1%.  There's a memory-to-memory copy, filesystem overhead, alloc, free...  But it does retain the looping feature.

In this workflow, you are probably also benefitting from copying the file to ramdisk, which is likely faster than copying to persistent storage.  The trace file "goes away" when you reboot the device or destroy the ramdisk, so you don't clutter your persistent storage.  And your timings may not have included the time it took for the `-pltf` option to preload the file, which of course could be amortized over the entire playback.  That makes any speed difference less significant.

At this point are you willing to withdraw this pull request?   Since some good work went into this, I think it would be great if you kept it on a branch for a while, in case we think of a need for it.

justuswang:
> At this point are you willing to withdraw this pull request? Since some good work went into this, I think it would be great if you kept it on a branch for a while, in case we think of a need for it.

Yes, I will withdraw this pull request and keep it on branch justuswang:vkreplay_preload here: https://github.com/justuswang/VulkanTools/tree/vkreplay_preload

justuswang:
Just came up an idea when I was copying a large file to the ram disk to replay, do you think there's a need to make vkreplay able to read compressed trace file?  (e.g. .gz file which compressed by gzip)

Of cause the fps result from the compressed trace file is not comparable with the fps result from replaying uncompressed trace file because there are some additional overhead when reading a compressed file.  But the fps results from compressed trace files are comparable with each other.

davidlunarg:
Yes, support for compressed trace files is something we are currently investigating. We are looking at compressing each packet, rather than the whole file, because the trace file sometimes needs to be randomly access (i.e. the portability table at the end of the trace file is contains file offsets). Stay tuned.

justuswang:
Great! That will be very useful when the ram disk is not large enough for some large trace files.  And it will also be very useful in storing trace files into a trace repository.

PR #432, opened 2018-03-28
vkreplay: Print frame per second

justuswang:
This change prints frame per second after finish replaying a trace file
for user to easily measure the performance.

If a trace file is replayed with multiple loops "NumLoops > 1", the
frame per second will be printed after each loop.

### DIFF
--- a/vktrace/vktrace_replay/vkreplay_main.cpp
+++ b/vktrace/vktrace_replay/vkreplay_main.cpp
@@ -134,6 +134,7 @@ int main_loop(vktrace_replay::ReplayDisplay display, Sequencer& seq, vktrace_tra
     seq.get_bookmark(startingPacket);
     unsigned int totalLoops = settings.numLoops;
     while (settings.numLoops > 0) {
+        uint64_t start_time = vktrace_get_time();
         while (trace_running) {
             display.process_event();
             if (display.get_quit_status()) {
@@ -199,6 +200,7 @@ int main_loop(vktrace_replay::ReplayDisplay display, Sequencer& seq, vktrace_tra
                                 // record the location of looping start packet
                                 seq.record_bookmark();
                                 seq.get_bookmark(startingPacket);
+                                start_time = vktrace_get_time();
                             }
 
                             if (frameNumber == settings.loopEndFrame) {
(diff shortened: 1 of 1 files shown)


davidlunarg:
When I run "vkreplay -l 5 -lsf 30 -lef 38 -t cube.vktrace", a line similar to this is repeated 5 times:

`vkreplay info: fps 59.717045 (loop duration(ns) 133965102, frame number 8 (30 , 37))`

It's confusing in that the command requests that frames 30-38 be replayed in the loop, but the output seems to indicate that frames 30-37 were replayed. Have we stumbled into an off by one error in the start/end frame code?

I think we can omit the frame numbers in the fps output. How about something like this for each iteration of the loop:

`vkreplay info: 69 frames, 1.3441556 seconds, 59.516903 fps`

I made up the above numbers, but you get the idea. I think I would prefer the total time reported to be in seconds rather than nanoseconds. Is there be a case where reporting in ns might be better?

justuswang:
> It's confusing in that the command requests that frames 30-38 be replayed in the loop, but the output seems to indicate that frames 30-37 were replayed. Have we stumbled into an off by one error in the start/end frame code?

Thanks for pointing this out, seems the current implementation of the option "-lef" is not correct.  Because the comparing of current frame number and loopEndFrame happens right after a vkQueuePresentKHR() packet when a frame is finished replaying where the current frame number is just added by 1.  That means the current frame number is "the correct current frame number + 1", I will fix this and include the fix in this pull request.

> I think we can omit the frame numbers in the fps output.

Seems the frame numbers in the fps output is needed.  The frame range info is useful for some performance analysis needs because different frame range may have different fps result.
After talked with some internal users I have got some information that user may want to see the info of "fps", "total time cost in a single loop", and "which frame range does the fps calculation applies to" in a single output message.
And keep those three information in the same place will help automated performance data collection for some users.

> I made up the above numbers, but you get the idea. I think I would prefer the total time reported to be in seconds rather than nanoseconds. Is there be a case where reporting in ns might be better?

Reporting the total time in nanoseconds is better because it can help to show the time more accurate than using seconds when replaying less frames.  (e.g. only 1 or 2 frames)

justuswang:
Yeah, it looks better.  Will change to this output.

davidlunarg:
Fixed the clang-format error and cherry-picked to master.

PR #434, opened 2018-03-29
layersvt: Devsim 1.2.3 bug fix

mikew-lunarg:
The DevSim implementation of vkGetPhysicalDeviceQueueFamilyProperties2KHR() was incorrectly populating the destination properties array.

This also restores the QueueFamilyProperties test that was temporarily disabled to avoid breaking CI.

Change-Id: I2ff07574e7ab7502932dae86b299e7abae1d5b32

### DIFF
--- a/layersvt/device_simulation.cpp
+++ b/layersvt/device_simulation.cpp
@@ -63,7 +63,7 @@ namespace {
 // When making ANY changes to the version, be sure to also update layersvt/{linux|windows}/VkLayer_device_simulation.json
 const uint32_t kVersionDevsimMajor = 1;
 const uint32_t kVersionDevsimMinor = 2;
-const uint32_t kVersionDevsimPatch = 2;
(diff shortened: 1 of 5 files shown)


layersvt/device_simulation.cpp
```diff
@@ -1279,13 +1279,29 @@ VKAPI_ATTR void VKAPI_CALL GetPhysicalDeviceQueueFamilyProperties(VkPhysicalDevi
...
+        return;
+    }
+
+    const uint32_t src_count = static_cast<uint32_t>(pdd->arrayof_queue_family_properties_.size());
+    if (!pQueueFamilyProperties2) {
+        *pQueueFamilyPropertyCount = src_count;
+        return;
+    }
+
+    // Careful: cannot use EnumerateProperties() here! (because src and dst structs are not the same type)
```
mikes-lunarg: :+1: for the warning

PR #450, opened 2018-04-13
vkreplay: Fix a corner case in replaying vkAM

justuswang:
Corner case:
A dstBuffer's memory is allocated too early and at the same time there's
a srcBuffer with the same buffer handle being created and used in
vkCmdCopyBuffer before the real dstBuffer being created and bind to the
dstBuffer's memory.

Sequence without portability table being enabled:
1. memory#dstBuffer is allocated
2. srcBuffer is created with handle#1
3. memory#srcBuffer is allocated
4. handle#1 is bind to memory#srcBuffer
5. handle#1 is used in vkCmdCopyBuffer as srcBuffer
6. handle#1 is destroyed
7. dstBuffer is created with handle#1
8. handle#1 is bind to memory#dstBuffer
9. handle#1 is used in vkCmdCopyBuffer as dstBuffer

The portability table feature will change the sequence to: (It will
create buffer before calling vkAllocateMemory if the buffer binds to the
memory is not created before allocation.)
1. dstBuffer is created with handle#1
2. memory#dstBuffer is allocated
3. srcBuffer is NOT created with handle#1, and vkreplay will use
   handle#1 directly for srcBuffer because vkreplay does not know the
   srcBuffer is different from the dstBuffer though they have the same
   buffer handle.
4. memory#srcBuffer is allocated
5. handle#1 is bind to memory#srcBuffer
6. handle#1 is used in vkCmdCopyBuffer as srcBuffer (Error happens here!
   because handle#1 was created with VK_BUFFER_USAGE_TRANSFER_DST_BIT)

This is fixed by destroy buffer/image created in
manually_replay_vkAllocateMemory() and do a clean up in object mappings.

### DIFF
--- a/vktrace/vktrace_replay/vkreplay_vkreplay.cpp
+++ b/vktrace/vktrace_replay/vkreplay_vkreplay.cpp
@@ -1956,6 +1956,7 @@ bool vkReplay::modifyMemoryTypeIndexInAllocateMemoryPacket(VkDevice remappedDevi
     VkImage remappedImage = VK_NULL_HANDLE;
     size_t bindMemIdx;
     VkImage bindMemImage = VK_NULL_HANDLE;
+    bool doDestroyImage = false;
 
     // Should only be here if we have a valid portability table and the replay platform does not match the trace platform
     assert(m_pFileHeader->portability_table_valid && m_platformMatch == 0);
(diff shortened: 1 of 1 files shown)


davidlunarg:
Looks good in general.

Could you add some comments as to why the image/buffer needs to be destroyed?  Probably add the comments after the comment on why it needs to be created in the first place, around line 2070.

Also, with this change, might lines 456 and 490 that check to see if the image/buffer has already been created no longer be needed?

justuswang:
> Could you add some comments as to why the image/buffer needs to be destroyed? Probably add the comments after the comment on why it needs to be created in the first place, around line 2070.

Sure, will do.

> Also, with this change, might lines 456 and 490 that check to see if the image/buffer has already been created no longer be needed?

Yes, those lines are no longer be needed.  Seems there's no other place to create image/buffer before replaying the packet of vkCreateImage/vkCreateBuffer.

davidlunarg:
Merged to master

PR #459, opened 2018-04-18
vktrace: store image in linear when doing trim

justuswang:
With this change, not only images being bound to device only memory but
also images with optimal tiling will be saved in linear layout using a
staging buffer via vkCmdCopyImageToBuffer.

It helps the trimmed trace file to have possibility to work on a
different GPU by saving image data in linear because image with optimal
tiling may have different memory layout on different GPUs.


davidlunarg:
@plohrmann, @mizhen: The two of you have worked with the trim code. What do you think of this change?

PR #460, opened 2018-04-19
vktrace: Fix some typos in scripts

justuswang:
This change fixes following typos in vktrace_file_generator.py:
* vkDestroyDevice, vkDestroySwapchainKHR, vkDestroyInstance do not have
return value so remove the check of "replayResult == VK_SUCCESS"

* vkAcquireNextImage has return value so add a check of "replayResult
== VK_SUCCESS"

* Remove the duplicate "vk" when generating vktrace_vk_packet_id.h

* Check pointer is not NULL before comparing *pointer with VK_TRUE in
GetPrintfParams for vk_type VkBool32


davidlunarg:
merged to master

PR #461, opened 2018-04-19
Fix vkreplay arg processing to reject negative values for unsigned settings.

### LINKED ISSUE #395: vkreplay -l <negative number> doesn't work right
Loops through -value -> -INT_MAX -> INT_MAX -> 0. A negative value for -l should be detected and an error generated. (Or a negative value could be used to indicate to loop foverever?)

karl-lunarg:
Fixes #395 

and addresses a similar problem with the -lsf and -lef settings.


karl-lunarg:
Updated PR to fix a clang-format violation.  I'll merge when CI finishes.

PR #462, opened 2018-04-19
vktrace: Search for matching physical device

joey-lunarg:
When enumerating physical devices in vkreplay, first check for a
matching device based on the vendor and hardware IDs. If no match is
found, map based on the indices.

Change-Id: I606886de30408380aba68a6ad3039ffa9a78420a

### DIFF
--- a/vktrace/vktrace_replay/vkreplay_vkreplay.cpp
+++ b/vktrace/vktrace_replay/vkreplay_vkreplay.cpp
@@ -546,10 +546,18 @@ VkResult vkReplay::manually_replay_vkEnumeratePhysicalDevices(packet_vkEnumerate
             vktrace_LogError("Skipping vkEnumeratePhysicalDevices() due to invalid remapped VkInstance.");
             return VK_ERROR_VALIDATION_FAILED_EXT;
         }
-        if (pPacket->pPhysicalDevices != NULL) pDevices = VKTRACE_NEW_ARRAY(VkPhysicalDevice, deviceCount);
+        if (pPacket->pPhysicalDevices != NULL) {
+            // If we are querying for the list instead of the count, use a previously acquired count
+            deviceCount = m_gpu_count;
+            pDevices = VKTRACE_NEW_ARRAY(VkPhysicalDevice, deviceCount);
+        }
         replayResult = m_vkFuncs.EnumeratePhysicalDevices(remappedInstance, &deviceCount, pDevices);
 
-        // TODO handle different number of physical devices in trace versus replay
+        if (pDevices == NULL) {
+            // If we are querying for the count, store it for later
+            m_gpu_count = deviceCount;
+        }
+
         if (deviceCount != *(pPacket->pPhysicalDeviceCount)) {
             vktrace_LogWarning("Number of physical devices mismatched in replay %u versus trace %u.", deviceCount,
                                *(pPacket->pPhysicalDeviceCount));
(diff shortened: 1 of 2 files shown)


vktrace/vktrace_replay/vkreplay_vkreplay.cpp
```diff
@@ -558,9 +558,27 @@ VkResult vkReplay::manually_replay_vkEnumeratePhysicalDevices(packet_vkEnumerate
         } else if (pDevices != NULL) {
             vktrace_LogVerbose("Enumerated %d physical devices in the system.", deviceCount);
         }
-        // TODO handle enumeration results in a different order from trace to replay
-        for (uint32_t i = 0; i < deviceCount; i++) {
-            if (pDevices != NULL) {
+        if (pDevices != NULL) {
+            for (uint32_t i = 0; i < deviceCount; i++) {
```
dustin-lunarg: This change is intended to handle the following case, which introduces a device count mismatch between trace and replay:
* The trace layer intercepted vkEnumeratePhysicalDevices before the Optimus layer and received: trace_devices = { intel_gpu, nvidia_gpu }
* vkreplay received a list of devices from vkEnumeratePhysicalDevices after it has been filtered by the Optimus layer: replay_devices = { nvidia_gpu }
* vkreplay needs to map trace_devices[1] to replay_devices[0]

To map each physical device from the trace file to a replay device, the loop counter should use *(pPacket->pPhysicalDeviceCount).  A mismatch between the trace and replay physical device counts will also need to be handled.

I think you want something like this:
```
for(i = 0; i < trace_device_count; i++)
    if ((i < replay_device_count) && (trace_device_id[i] == replay_device_id[i]))
        // Exact match
    else
        // Search for a match in the list of available physical devices
        for (j = 0; j < replay_device_count; j++)
                if (trace_device_id[i] == replay_device_id[j])
                    // Map i to j
```

vktrace/vktrace_replay/vkreplay_vkreplay.cpp
```diff
@@ -558,9 +558,27 @@ VkResult vkReplay::manually_replay_vkEnumeratePhysicalDevices(packet_vkEnumerate
...
+                // First, check if device on the same index has matching vendor and device ID
+                auto traceProps = tracePhysicalDeviceProperties[pPacket->pPhysicalDevices[i]];
```
dustin-lunarg: This may need to be changed to the list of GPU IDs from the trace file header.

vktrace/vktrace_replay/vkreplay_vkreplay.cpp
```diff
@@ -558,9 +558,27 @@ VkResult vkReplay::manually_replay_vkEnumeratePhysicalDevices(packet_vkEnumerate
...
+                auto replayProps = replayPhysicalDeviceProperties[pDevices[i]];
```
dustin-lunarg: This list can be generated above, just before the for loop.

vktrace/vktrace_replay/vkreplay_vkreplay.cpp
```diff
@@ -558,9 +558,27 @@ VkResult vkReplay::manually_replay_vkEnumeratePhysicalDevices(packet_vkEnumerate
...
+                    if (j == i) {
+                        continue;  // Already checked this
+                    }
+                    traceProps = tracePhysicalDeviceProperties[pPacket->pPhysicalDevices[j]];
+                    if (traceProps.vendorID == replayProps.vendorID && traceProps.deviceID == replayProps.deviceID) {
+                        m_objMapper.add_to_physicaldevices_map(pPacket->pPhysicalDevices[j], pDevices[i]);
+                        continue;
+                    }
+                }
+                // If all else fails, just map the indices
```
dustin-lunarg: This will need to handle a mismatch in trace and replay device count.

Issue #446, opened 2018-04-10
vkreplay: Is it possible for vkreplay to automatically reorient on Android?


justuswang:
A trace file traced from Android phone in landscape mode is not able to be replayed in landscape mode automatically on the same phone using vkreplay.apk.

Is it possible to automatically switch to landscape mode when replaying on Android and the trace's width is greater than height?

dustin-lunarg:
We may be able to do this by invoking Activity.setRequestedOrientation() through JNI after detecting that the width is greater than the height.  I should be able to look at this, but may not be able to do so immediately.

justuswang:
I have created a PR https://github.com/LunarG/VulkanTools/pull/477 to do the automatically orientation.  And it works for me when I was testing on an Android phone.  Please let me know if my PR has any side effect?

justuswang:
@dustin-lunarg , your change looks great!  I will close my PR.

dustin-lunarg:
This should be resolved by #478.

PR #477, opened 2018-04-26
vkreplay: GH446, Auto reorient to fit surface extent

justuswang:
This change is comparing content's orientation with surface's
orientation, if they are not match, vkreplay will rotate the content 90
degrees to make the content shown in expected orientation.

It is implemented by setting VK_SURFACE_TRANSFORM_ROTATE_90_BIT_KHR to
pCreateInfo->preTransform in vkCreateSwapchainKHR() when the content's
orientation does not match surface's orientation retrieved via
vkGetPhysicalDeviceSurfaceCapabilitiesKHR().

### DIFF
--- a/vktrace/vktrace_replay/vkreplay_vkreplay.cpp
+++ b/vktrace/vktrace_replay/vkreplay_vkreplay.cpp
@@ -2993,6 +2993,9 @@ VkResult vkReplay::manually_replay_vkGetPhysicalDeviceSurfaceCapabilitiesKHR(
 
     replayResult = m_vkFuncs.GetPhysicalDeviceSurfaceCapabilitiesKHR(remappedphysicalDevice, remappedSurfaceKHR,
                                                                      pPacket->pSurfaceCapabilities);
+
+    replaySurfaceCapabilities[remappedSurfaceKHR] = *(pPacket->pSurfaceCapabilities);
+
     return replayResult;
 }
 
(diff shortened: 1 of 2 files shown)


karl-lunarg:
I think my understanding is that the trace was created in landscape mode, but vkreplay always starts (on Android) in portrait mode, despite how the trace was made.

Although this "automatic" detection is nice, it may not play well with an application that intentionally set the ROTATE90 bit.

And since this seems like an Android-only problem, due to the default nature of vkreplay starting up in portrait mode, it seems that there should be some Android-only code in vkreplay that exposes a command line option that makes vkreplay start in landscape mode, which should also solve the problem.

Would this be a better approach?

dustin-lunarg:
@justuswang I implemented a different method for changing the screen orientation in #478.  It uses Activity.setRequestedOrientation() to change the orientation.  Can you test this change to see if it resolves the issue that you reported in #446?

justuswang:
@dustin-lunarg, your change looks great!  But when I was testing with it, the orientation is correct but the content was replayed in the left half of the screen instead of full screen.

It looks like: (Assuming following is an Android phone in landscape mode)
+-----------------------------------------------------+
|                                    |                                           |
|                                    |                                           |
|          content              |            black screen           |
|          being replayed  |                                           |
|                                    |                                           |
+-----------------------------------------------------+

@karl-lunarg, yes, I agree with you and seems @dustin-lunarg's change is able to solve your concern.
Looking forward to have https://github.com/LunarG/VulkanTools/pull/478 work as expected and I will close my PR then.

dustin-lunarg:
@justuswang The issue that you observed with #478 should be fixed now.  There was a mistake with the change to the Android manifest file.

dustin-lunarg:
@justuswang I was able to reproduce your issue with the initial version of #478, but have not been able to reproduce it with the latest version.  I did notice that vkreplay was ignoring system events after starting playback, so have added a change to perform event handling.  I have also added a call to ANativeWindow_setBuffersGeometry from vkDisplay::resize_window.

#478 appears to work correctly with Pixel XL, Pixel C, and Shield K1.  The changes to AndroidManifest.xml are required for the orientation change to work correctly (prevents the activity from restarting on screen rotation), so if you are using a manifest file other than the one provided in this repository, make sure to add the android:configChanges attribute.  

If you are still seeing failure on your device,  we could add a command line option to specify the replay orientation.

justuswang:
It works now.  Thanks!  I'm going to close my PR.

PR #478, opened 2018-04-27
vktrace: Adjust Android screen orientation for replay

dustin-lunarg:
Implements vkDisplay::resize_window for Android, to adjust the
screen orientation based on the requested window width and height.
Screen orientation is changed by invoking Activity.setRequestedOrientation
through JNI.

### DIFF
--- a/vktrace/vktrace_replay/vkreplay_vkdisplay.cpp
+++ b/vktrace/vktrace_replay/vkreplay_vkdisplay.cpp
@@ -386,6 +386,8 @@ void vkDisplay::resize_window(const unsigned int width, const unsigned int heigh
                 m_jni_vm->DetachCurrentThread();
             }
         }
+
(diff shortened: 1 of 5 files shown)


vktrace/vktrace_replay/vkreplay_vkdisplay.cpp
```diff
@@ -29,6 +29,10 @@
 #include <linux/input.h>
 #endif
 
+#if defined PLATFORM_LINUX && defined ANDROID
```
cnorthrop: I realize this is consistent with the lines above, but generally wrapping the values being checked in parentheses makes this much easier to read.  There are some examples below.

dustin-lunarg: Agreed about readability, but as noted the code was written this way to be consistent with the existing adjacent code.  I would like to do cleanup for style consistency in a separate PR.

PR #484, opened 2018-05-03
vkreplay: Implement --CompatibiltyMode

davidlunarg:
Change-Id: Icc8b08c179214eeb4080701fed4fdeb0649fc058


davidlunarg:
I'm not sure why the descriptions of the command line options occur 2 times in the code and the default values for the command lines options occur 3 times. Will attempt to fix this in a future pull request.

PR #485, opened 2018-05-03
android: Add trace replay APK script

joey-lunarg:
Change-Id: I0c958e3d2ff0d412f5bff93b3ad9d9aa42533a49

### DIFF
--- a/build-android/tracereplay_apk.sh
+++ b/build-android/tracereplay_apk.sh
@@ -1,0 +1,131 @@
+function printUsage {
+   echo "Supported parameters are:"
+   echo "    -p|--platform <platform> (optional)"
+   echo "    -s|--serial <target device serial number> (optional)"
+   echo "    -a|--apk <target apk file>"
+   echo "    -k|--package <target apk package>"
+   echo "    -v|--vktrace <vktrace executable path>"
+   echo
+   echo "i.e. ${0##*/} -p <platform> -f <test filter> -s <serial number>"
+   exit 1
+}
+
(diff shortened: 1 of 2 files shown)


cnorthrop:
The existing [build_vktracereplay.sh](https://github.com/LunarG/VulkanTools/blob/master/build-android/build_vktracereplay.sh) script will create the APK for you.  Can we add parameters to that instead of creating a new script?  Or break up the scripts into smaller pieces if it is doing too much?

joey-lunarg:
@cnorthrop Yeah I agree that the scripts are getting a little excessive, but the intent of build_vktracereplay.sh was to build trace/replay and not the traceable APK. The typical usage of the scripts is to run build_vktracereplay.sh, then run vktracereplay.sh on an existing APK with layers. We do have cube-with-layers, but I would like this functionality with other APKs for testing. Howabout this: I consolidate "vktracereplay.sh", "create_trace.sh", "replay_trace.sh" to avoid overlapping functionality (invoke the latter 2 from the 1st) and I add an option to be able to inject the layer object files into the APK in "create_trace.sh" instead of a new script. However this would overload the functionality of "create_trace.sh" a bit. I could also just make the name more clear of a new script, like "create_traceable_apk.sh". What do you think?

cnorthrop:
Ah, you are injecting those layers into a target APK, I see.  That's new functionality.  Yeah, I think having a clearer name would be fine.

joey-lunarg:
Not needed for now

PR #488, opened 2018-05-07
Screenshot: Create command pool for use in WritePPM

Tony-LunarG:
For thread safety as well as ensuring that the correct
queue family index is used, create a command pool in
WritePPM instead of picking one of the application's.

Change-Id: Ic6b9a76dcdcc8ef5b93938e07f0c0a6cdf539126

### DIFF
--- a/layersvt/screenshot.cpp
+++ b/layersvt/screenshot.cpp
@@ -1,7 +1,7 @@
 /*
  *
- * Copyright (C) 2015-2016 Valve Corporation
- * Copyright (C) 2015-2016 LunarG, Inc.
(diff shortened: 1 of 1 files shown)


layersvt/screenshot.cpp
```diff
@@ -167,8 +168,10 @@ typedef struct {
     std::list<VkCommandPool> commandPools;
```
davidlunarg: Since we now create our own command pool, the variable commandPools should be removed, as well as all references to it. Also remove the CreateCommandPools hook function and all references to it.

PR #492, opened 2018-05-09
vkreplay: Set replayResult to VK_SUCCESS

MarkY-LunarG:
The return result was initialized to an error code, but this
can cause problems for vkDestroyInstance which only allows success.


MarkY-LunarG:
I would argue that it's always better to have a definable default since someone might add a case statement in the future where it's not initialized.  If you remove the default value, you're relying on all compilers in that scenario to catch the case that the function is returning a non-initialized variable.  But I can make the change if you prefer it.

davidlunarg:
6 of one, half dozen of another. I don't think it's that big a deal which way we do it.

PR #493, opened 2018-05-09
vktrace: Handle pNexts in MemoryRequirements2 routines

TonyBarbour:
Add pNext processing to Get Image, Buffer and Image Sparse MemoryRequirements2 routines


scripts/vktrace_file_generator.py
```diff
@@ -859,6 +859,8 @@ def GenerateReplayGenSource(self):
                     replay_gen_source += '            VkPhysicalDeviceMemoryProperties memProperties = *(pPacket->pMemoryProperties);\n'
                 elif cmdname == 'GetImageMemoryRequirements':
                     replay_gen_source += '            VkMemoryRequirements memReqs = *(pPacket->pMemoryRequirements);\n'
+                elif 'SparseMemoryRequirements2' in cmdname:
+                    replay_gen_source += '            vktrace_interpret_pnext_pointers(pPacket->header, (void *)pPacket->pInfo);\n'
```
davidlunarg: vktrace_interpret_pnext_pointers needs to be called for all structs in generated replay entry points, which is beyond the scope of this pull request. I will create an issue in VulkanTools wrt these missing calls to vktrace_interpret_pnext_pointers.

PR #495, opened 2018-05-10
vktrace: Use consistent syntax for defined

dustin-lunarg:
Make all use of the preprocessor defined token use the 'defined(X)'
syntax, instead of 'defined X'.

This is a follow up to comments made on #478.

### DIFF
--- a/vktrace/vktrace_common/vktrace_common.h
+++ b/vktrace/vktrace_common/vktrace_common.h
@@ -69,7 +69,7 @@
 
 static const uint32_t  INVALID_BINDING_INDEX = UINT32_MAX;
 // Windows needs 64 bit versions of fseek and ftell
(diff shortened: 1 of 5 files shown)


davidlunarg:
There are some #ifdef lines in vkreplay_vkreplay.cpp that were not changed over to use parenthesis. Should probably also change them to read '#if defined(...)' to be consistent.

This PR only changed files in vktrace/vktrace_replay. There are lots of files in vktrace/* that need to be changed in the same manner. I'll make it your call if you want to change those in this PR, or we can make that change in a later PR.

dustin-lunarg:
@davidlunarg I don't see any occurrence of the '#if defined ' pattern in any of the other vktrace files.  It was only present in the vktrace/vktrace_replay files.

PR #497, opened 2018-05-11
vktrace: Change use of ROUNDUP_TO_4 so mem past orig src not access

davidlunarg:
Whenever ROUNDUP_TO_4 is used, need to make sure that it is used to
round up the size of packet elements, but is not used as the size of
the data to be copied from the caller's data to the packet. If the
rounded up size is used for the copy, the copy could potentially
access memory past the end of a buffer/string that may not be legal
to access.

This pull request addresses this problem.

Change-Id: If7b9160611509900c9eca0c34522bd3768cd8477


cforfang:
One comment somewhat related to this; many 64-bit platforms require 8-byte alignment for some Vulkan structs. With that in mind, shouldn't most of these `ROUNDUP_TO_4` really be `ROUNDUP_TO_8` to guarantee such structs end up stored at valid alignment in the trace-file?

(I had a look at the actual changes here as well; looks good to me.)

davidlunarg:
@justuswang : You originally discovered the need for using ROUND_TO_4 when running vktrace/replay on ARM. Are you aware of a need to use ROUND_TO_8 on ARM or any other architecture?

justuswang:
As I have said in PR #426, my change only makes sure the starting address of the additional data and the starting address of each packet in the trace file to be 8-byte aligned, which is the minimum change to fix the alignment fault on ARMv8 platform without any uncertain side effect.

Maybe it is better to have ROUND_TO_8 everywhere to guarantee every struct is aligned to 8 bytes not only the starting address but also the struct data itself.  Especially when there are multiple additional buffers with different alignment requirements. (e.g. the first additional buffer requires 4-byte alignment but the second one requires 8-byte alignment.)

It won't be a problem for now because ARMv8 platform supports unaligned access and the compiler won't generate vst1.64 instruction in that condition according to what I have observed.  But it is not guaranteed to work in the future.

davidlunarg:
I attempted to modify this pull request to use ROUNDUP_TO_8 so that all structures are aligned to 8 byte boundaries. The changes affected just about every api entry point. I got it mostly working, except for some cases for auto gen code, which I couldn't get right. I decided to abandon that effort for now since the reason to do ROUNDUP_TO_8  for some possible future CPU architecture. I would like to go ahead and merge this pull request, and I will create an issue identifying ROUNDUP_TO_8 as something that should be done in the future.

cforfang:
Sounds like a pragmatic solution to me, but I just want to highlight this isn't a problem that affects "_some possible future CPU architecture_"; it's a problem for x64 and AArch64 _right now_ since VKTrace isn't adhering to the alignment requirements.

This doesn't **guarantee** there will be many issues in practice since it depends entirely on the compiler code-gen, but again we have already seen GCC generate instructions that caused a failure already and have had to work around this.

So leaving it to the backlog is fine, just keep in mind this implies some degree of finger-crossing that it won't suddenly blow up depending on to what extent compilers chose to take advantage of the alignment requirements.

PR #499, opened 2018-05-15
layersvt: Enable GetPhysicalDeviceFeatures2

Whenning42:
Set up GetPhysicalDeviceFeatures2(KHR) to make calls to the
corresponding vulkan function instead of simply calling the non-extended
vkGetPhysicalDeviceFeatures.

### DIFF
--- a/layersvt/device_simulation.cpp
+++ b/layersvt/device_simulation.cpp
@@ -66,7 +66,7 @@ namespace {
 
 const uint32_t kVersionDevsimMajor = 1;
 const uint32_t kVersionDevsimMinor = 2;
-const uint32_t kVersionDevsimPatch = 5;
+const uint32_t kVersionDevsimPatch = 6;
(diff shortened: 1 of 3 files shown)


Whenning42:
The second commit adds GetPhysicalDeviceProperties2

layersvt/device_simulation.cpp
```diff
@@ -1247,6 +1247,11 @@ VKAPI_ATTR void VKAPI_CALL GetPhysicalDeviceProperties(VkPhysicalDevice physical
 
 VKAPI_ATTR void VKAPI_CALL GetPhysicalDeviceProperties2KHR(VkPhysicalDevice physicalDevice,
```
mikew-lunarg: This commit seems incomplete compared to the GetPhysicalDeviceFeatures2KHR replacement in the previous commit.  There is not a GetPhysicalDeviceProperties2 entry point, that would also be added to getprocaddress.

PR #501, opened 2018-05-15
vktrace: Check for loadable trace library

davidlunarg:
Modified vktrace when in non-server mode to checkout to see if the
trace layer library file is available and loadable, and generates an
error if is is not. This helps prevent the scenario in which vktrace
hangs or crashes when the trace layer can't be loaded.

Change-Id: I61f02f5aff9cc56b5382ef58b94351f6e9b2b928

### DIFF
--- a/vktrace/vktrace_trace/vktrace.cpp
+++ b/vktrace/vktrace_trace/vktrace.cpp
@@ -392,14 +392,44 @@ int main(int argc, char* argv[]) {
         } else {
             // We are not in server mode.
 
-            // Verify trace layer is available.
-            void* traceLayerLibHandle;
-            traceLayerLibHandle = vktrace_platform_open_library(VKTRACE_LIBRARY_NAME(VkLayer_vktrace_layer));
-            if (!traceLayerLibHandle) {
-                vktrace_LogError("Cannot find trace layer %s", VKTRACE_LIBRARY_NAME(VkLayer_vktrace_layer));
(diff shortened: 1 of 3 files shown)


vktrace/vktrace_trace/vktrace.cpp
```diff
@@ -390,6 +390,18 @@ int main(int argc, char* argv[]) {
             fflush(stdout);
             g_settings.arguments = NULL;
         } else {
+            // We are not in server mode.
+
+            // Verify trace layer is available.
+            void* traceLayerLibHandle;
+            traceLayerLibHandle = vktrace_platform_open_library(VKTRACE_LIBRARY_NAME(VkLayer_vktrace_layer));
```
dustin-lunarg: This will search the system library path, but VK_LAYER_PATH will also need to be searched for this case.  It may be necessary to create an instance to check for the presence of the layer.

davidlunarg:
I modified this pull request to use vkCreateInstance with the trace library enabled, then check to see of the call to vkCreateInstance succeeded.

The call to vkCreateInstance will fail and vktrace will print an error message if the json file for the trace layer is missing. However, if the json file is present but the trace layer shared library is missing, vkCreateInstance will indicate success with a return value of VK_SUCCESS. I think this is a bug in the loader, and until this bug is fixed in the loader, this pull request does not completely address the problem it was intended to fix.

PR #502, opened 2018-05-16
vkreplay: synchronize window resizes on Linux

erik-kz:
vkreplay currently uses a 0.05 second delay after
sending requests to the X server to resize and map
its window hoping that this will ensure it has been
processed before proceeding.

There is not guaranteed, however. Instead, it should
wait to receive a map notify event corresponding to
the request, pausing the replay meanwhile.


erik-kz:
Thanks David, formatting should be fixed now.

PR #503, opened 2018-05-17
vktrace: GH500, Fix deadlock in vktrace

justuswang:
This change fixes deadlock when tracing a multithread application by
making sure the locks in an API call are acquired in a consistent order.

### DIFF
--- a/vktrace/vktrace_layer/vktrace_lib_trace.cpp
+++ b/vktrace/vktrace_layer/vktrace_lib_trace.cpp
@@ -208,6 +208,9 @@ VKTRACER_EXPORT VKAPI_ATTR VkResult VKAPI_CALL __HOOKED_vkMapMemory(VkDevice dev
     vktrace_trace_packet_header* pHeader;
     packet_vkMapMemory* pPacket = NULL;
     VKAllocInfo* entry;
+#ifdef USE_PAGEGUARD_SPEEDUP
+    pageguardEnter();
+#endif
     CREATE_TRACE_PACKET(vkMapMemory, sizeof(void*));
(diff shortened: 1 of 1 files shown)


erik-kz:
Do you think it would be worthwhile to add an assertion to `pageguardEnter()` to check that we don't currently hold `s_trace_lock`?

justuswang:
It's a good idea to always check if we currently hold s_trace_lock in pageguardEnter().

But it will bring additional operations to vktrace (I haven't found a way to simply doing an assert on the lock state).  Here's what I can think about to check the lock state:

Additional steps to be added in pageguardEnter() to check if s_trace_lock is locked: (assuming s_trace_lock is visible in pageguardEnter())
* use pthread_mutex_trylock(s_trace_lock) to check lock state.
** it will return EBUSY if s_trace_lock is already locked
** it will return zero and lock s_trace_lock if it is not already locked.
*** use pthread_mutex_unlock(s_trace_lock) to unlock s_trace_lock

I personally prefer to not add above operations to pageguardEnter() for now if there's no other way to do the check for lock state.  There are only 6 pageguardEnter() in vktrace_lib_trace.cpp, so I think it is easy to make sure pageguardEnter() happens before s_trace_lock being locked when an API call will acquire both of those locks.

Please let me know if there's any simpler way to check the lock state?

davidlunarg:
While @erik-kz's suggestion to add an assert to make sure that the trace layer does not attempt to get the pageguard lock while holding the s_trace_lock is a good idea, @justuswang points out it is not simple to implement and might be an issue in only 6 places in the code. I will merge this pull request as submitted by @justuswang.

PR #507, opened 2018-05-22
tests: Adjust DevSim test after repo reorg

mikew-lunarg:
Tweaks to DevSim test script following VulkanTools using new KhronosGroup repos.
Test 1 (using vkjson) is commented out, since vkjson is not included in the submodules, and we want to move away from it anyway.


mikew-lunarg:
Pushed to master so CI debugging can proceed.

Issue #504, opened 2018-05-18
Update this repo to no longer depend on the deprecated LVL repo


Whenning42:
This repo uses the deprecated Vulkan-LoaderAndValidationLayers. I imagine this should be updated to use the repos LVL was split into.

mark-lunarg:
Yup, working on it now.

charles-lunarg:
Repo no longer depends on LVL

PR #509, opened 2018-05-23
vkreplay: Enable '-s' and '-sf' options on Andriod

justuswang:
This change is a follow up of #506 which makes '-s' and '-sf' options
of vkreplay work on Android platform.


dustin-lunarg:
The CI failures look to be related to issue #504.  A fix for that issue is in progress.

PR #510, opened 2018-05-23
vkreplay: Fix error when vkAllocateMemory is skipped

justuswang:
The problem is vkAllocateMemory will NOT do memory allocation if a
memory is not being bound to any image or buffer.  But the memory may
be used in vkMapMemory, vkUnmapMemory, vkFreeMemory,
vkFlushMappedMemoryRanges, vkInvalidateMappedMemoryRanges which will
cause error when calling
m_objMapper.m_devicememorys.find(NOT_ALLOCATED_MEMORY)->second without
checking the NOT_ALLOCATED_MEMORY exists in m_objMapper.m_devicememorys
or not.

This fix is to skip vkMapMemory, vkUnmapMemory, vkFreeMemory,
vkFlushMappedMemoryRanges, and vkInvalidateMappedMemoryRanges when such
memory is besing used.


vktrace/vktrace_replay/vkreplay_vkreplay.cpp
```diff
@@ -2313,6 +2317,11 @@ void vkReplay::manually_replay_vkFreeMemory(packet_vkFreeMemory *pPacket) {
         return;
     }
 
+    if (m_objMapper.m_devicememorys.find(pPacket->memory) == m_objMapper.m_devicememorys.end()) {
+        vktrace_LogError("Skipping vkFreeMemory() due to invalid remapped VkDeviceMemory.");
```
davidlunarg: Should this be vktrace_LogWarning? It's legal to allocate memory, never use it, and then free the memory, in which case we could generate a warning, not an error.

justuswang: Yes, vktrace_LogWarning should be used here instead of vktrace_LogError.  Will fix it in next commit.

PR #515, opened 2018-05-28
vktrace: Fix unlock uninitialized/unlocked mutex lock

justuswang:
This change fixes the bug mentioned in
https://github.com/LunarG/VulkanTools/commit/3d3c8e59f67a5d6262200433e3780d133ec7e234#commitcomment-29121242

This change adds an alternate to vktrace_delete_trace_packet() that
does not try to leave the critical section which only initialized and
entered in vktrace_initialize_trace_packet_utils() and
vktrace_create_trace_packet(). It will help to safely delete trace
packet read via vktrace_read_trace_packet() without having problem of
leaving uninitialized/already left critical section.

### DIFF
--- a/vktrace/vktrace_layer/vktrace_lib_trim.cpp
+++ b/vktrace/vktrace_layer/vktrace_lib_trim.cpp
@@ -2598,18 +2598,18 @@ void write_all_referenced_object_calls() {
     // Instances (& PhysicalDevices)
     for (auto obj = stateTracker.createdInstances.begin(); obj != stateTracker.createdInstances.end(); obj++) {
         vktrace_write_trace_packet(obj->second.ObjectInfo.Instance.pCreatePacket, vktrace_trace_get_trace_file());
-        vktrace_delete_trace_packet(&(obj->second.ObjectInfo.Instance.pCreatePacket));
+        vktrace_delete_trace_packet_no_lock(&(obj->second.ObjectInfo.Instance.pCreatePacket));
 
         if (obj->second.ObjectInfo.Instance.pEnumeratePhysicalDevicesCountPacket != NULL) {
             vktrace_write_trace_packet(obj->second.ObjectInfo.Instance.pEnumeratePhysicalDevicesCountPacket,
                                        vktrace_trace_get_trace_file());
-            vktrace_delete_trace_packet(&(obj->second.ObjectInfo.Instance.pEnumeratePhysicalDevicesCountPacket));
+            vktrace_delete_trace_packet_no_lock(&(obj->second.ObjectInfo.Instance.pEnumeratePhysicalDevicesCountPacket));
         }
 
         if (obj->second.ObjectInfo.Instance.pEnumeratePhysicalDevicesPacket != NULL) {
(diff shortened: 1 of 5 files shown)


vktrace/vktrace_layer/vktrace_lib_trim.cpp
```diff
@@ -854,7 +854,8 @@ void generateMapUnmap(bool makeCalls, VkDevice device, VkDeviceMemory memory, Vk
         // unmap for it.
         vktrace_delete_trace_packet(&pMapMemory);
     } else {
-        *ppMapMemoryPacket = pMapMemory;
+        *ppMapMemoryPacket = trim::copy_packet(pMapMemory);
+        vktrace_delete_trace_packet(&pMapMemory);
```
davidlunarg: This change seems like it is unrelated to the gist of this pull request. If it is unrelated, could you make this change in a separate pull request? Or maybe I'm wrong?

vktrace/vktrace_layer/vktrace_lib_pageguard.cpp
```diff
@@ -494,15 +494,14 @@ VkResult vkFlushMappedMemoryRangesWithoutAPICall(VkDevice device, uint32_t memor
     if (!g_trimEnabled) {
         // trim not enabled, send packet as usual
         FINISH_TRACE_PACKET();
-    } else if (g_trimIsPreTrim) {
-        vktrace_finalize_trace_packet(pHeader);
-    } else if (g_trimIsInTrim) {
-        // Currently tracing the frame, so need to track references & store packet to write post-tracing.
+    } else {
```
davidlunarg: The changes in file vktrace_lib_pageguard.cpp seem like they are unrelated to the gist of this pull request. If it is unrelated, could you make this change in a separate pull request? Or maybe I'm wrong?

vktrace/vktrace_layer/vktrace_lib_trim.cpp
justuswang: You are correct.  It is not fixing the issue of "unlock uninitialized mutex lock" but a fix to "lock on a locked mutex lock".  I forgot to update the description of the pull request when pushing the later two commits.  Shall I update the description of this pull request or create a new one?

I think it is better to have both fixes being merged together since both of them are needed to fix the mutex lock problem.

vktrace/vktrace_layer/vktrace_lib_pageguard.cpp
justuswang: You are correct.  It is not fixing the issue of "unlock uninitialized mutex lock" but a fix to the risk of "lock on a locked mutex lock".  I forgot to update the description of the pull request when pushing the later two commits.  Shall I update the description of this pull request or create a new one?

I think it is better to have both fixes being merged together since both of them are needed to fix the mutex lock problem.

davidlunarg: Please create a new pull request.

justuswang: Sure, https://github.com/LunarG/VulkanTools/pull/519 has been created.

PR #520, opened 2018-05-30
vktrace: GH498, sync back mapped memory on Linux

justuswang:
The problem is:

The default configuration of vktrace is not able to know there's a read
to a mapped memory on Linux and Android platforms. Which means it is not
going to sync data back from the real mapped memory (which was mapped
via vkMapMemory) to the copy of that memory for Vulkan application to
read.

This is the default behavior when
PAGEGUARD_ADD_PAGEGUARD_ON_REAL_MAPPED_MEMORY is disabled in
vktrace/vktrace_layer/vktrace_lib_pageguardcapture.h.

And it will cause problem when tracing an app which needs to read back
rendering results for environment probes.

The workaround in this commit:

Sync real mapped memory to the copy of that memory after
vkCmdCopyImageToBuffer being executed.
The real mapped memory is bound with the destination buffer.

The sync (memcpy) is added in __HOOKED_vkWaitForFences when a fence from
vkQueueSubmit is signaled and the command buffer submitted contains
vkCmdCopyImageToBuffer command.

This will at least make it possible for an app to read the correct
buffer data back. It won't solve memory read problem when a linear
image's memory is mapped to CPU memory and application wants to read
from that CPU memory. But seems it is expected for an application to
always read from buffer instead of read from image. So this workaround
should work for most applications which need to read buffer back.


davidlunarg:
I like this change, and the code seems correct.

However: The comment above is a good description of why the change needs to be made and how the problem is being (partially) solved. That information would be good to have in the code, since someone looking at the code in the future isn't likely to also read this pull request. Could add these descriptions as comments to the code you are adding?

justuswang:
Sure, will add those descriptions as comments.

davidlunarg:
Cherry-picked the 3 commits in this pull request to master

PR #528, opened 2018-06-01
tests: Update vktracereplay.sh and remove smoketest

joey-lunarg:
Change-Id: If1e2edb8544d24cd08c888bedb3aeab586780a62

### DIFF
--- a/tests/vktracereplay.sh
+++ b/tests/vktracereplay.sh
@@ -1,3 +1,5 @@
+# Note: For this script to run successfully, cube must be in PATH.
+
 #!/bin/bash
 #set -x
 if [ -t 1 ] ; then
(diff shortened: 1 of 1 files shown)


tests/vktracereplay.sh
```diff
@@ -47,13 +47,7 @@ function trace_replay {
...
 }
 
-trace_replay cube "" "--PMB false"
-# Test smoketest with pageguard
-trace_replay smoketest "" "--PMB true"
-# Test smoketest without pageguard, using push constants
-trace_replay smoketest "-p" "--PMB false"
-# Test smoketest without pageguard, using flush call
-trace_replay smoketest "--flush" "--PMB false"
+trace_replay ${VULKAN_SDK}/../examples/build cube "" "--PMB false"
```
cnorthrop: Let's not require the SDK to test vktrace.

Cube can be built out of Vulkan-Tools.

joey-lunarg: More changes to the organization of the repos is to come, and the current plan is to get rid of submodules. Vulkan-Tools will soon be removed from VulkanTools and dependencies will be found in the SDK if possible, and source dependencies will be referenced by environment variables.

cnorthrop: Yes, but the plan is for local overrides to take precedence over the SDK.  What you've added looks to be tied very closely to the SDK.

Haven't heard anything about more repo splits but that would be very good to know so we don't waste effort.

joey-lunarg: Okay, I think the best solution here is to reference cube through the path.

PR #533, opened 2018-06-09
vkreplay: Replace map with unordered_map

justuswang:
This change replaces std::map with std::unordered_map in vkreplay to
have better performance when looking up elements from objmapper.


davidlunarg:
Were you able to measure a difference in replay performance when you switched to using unordered maps?

PR #534, opened 2018-06-09
vkreplay: Make debug verbosity work on all builds

justuswang:
This change makes debug verbosity ("-v debug" option) work on both
_DEBUG build and non _DEBUG build.  It helps to make debug verbosity
work on Android build instead of crash when replaying with "-v debug"
option.


davidlunarg:
The intention of the "-v debug" option is for the use of the developers of vktrace/vkreplay to use to debug vktrace/vkreplay. They would probably debug vktrace/vkreplay with DEBUG build, so thus "-v debug" is only make available on DEBUG builds. I'm trying to understand why the "-v debug" option is needed for non-DEBUG builds. Is it something unique to Android?

If we change vktrace/replay to always allow "-v debug",  the documentation file vktrace.md would also have to change.

justuswang:
> Is it something unique to Android?
Yes, because I haven't found a way to make "-v debug" work on Android.
I have tried to add `LOCAL_CPPFLAGS += -D_DEBUG` to build-android/jni/Android.mk but it does not work.
And I have found the x64 "Release" build also have "-v debug" available because there's `set(CMAKE_CXX_FLAGS_DEBUG_LIST "-g -O0 -D_DEBUG")` in vktrace/CMakeLists.txt for not only Debug build but also Release build (applied for Windows and Linux builds but not Android build).
So I'm thinking of make it available for all the builds since it was not working on Android build only.

justuswang:
Tried with the latest Vulkan SDK on Windows and yes it does not list -v debug as a possible option.
Also tried to build a release build using the latest code of this repo and there's no "-v debug" option either.

Closing this PR and see if there's any other way to make "-v debug" work on Android platform.

PR #536, opened 2018-06-11
layersvt: Add index to draw call in api dump

justuswang:
This change adds index of draw call to api dump layer's output.
Draw call index will be shown in vkCmdDraw, vkCmdDrawIndexed,
vkCmdDrawIndirect, and vkCmdDrawIndexedIndirect calls in the output of
api dump layer.


charles-lunarg:
This behavior could be useful, but support for html and json outputs would be needed for feature parity.

charles-lunarg:
Closing due to inactivity.

PR #538, opened 2018-06-12
ci: Get AppVeyor working again

karl-lunarg:
Pull request needed to trigger AppVeyor on a branch.

Change-Id: If51027c95b735009a1e786e5c26e22513cefb8c9


karl-lunarg:
The Release build is OK on AppVeyor, but the Debug build is broken due to some problems that are addressed in #540.  These problems will be much easier to fix if this commit is made part of #540, so committing now.

At least before #540 hits, this commit changes the number of build configs to 2 and the Release/x64 build is working.

PR #551, opened 2018-06-21
vktrace: Use right arrow to skip frames. Add -p

joey-lunarg:
Adding "-p true" to vkreplay will now pause on the first frame.

Change-Id: I311c393fa44c9afab7c38810d6dd091d52056940

### DIFF
--- a/vktrace/vktrace_replay/vkreplay_main.cpp
+++ b/vktrace/vktrace_replay/vkreplay_main.cpp
@@ -41,7 +41,7 @@
 #include "vkreplay_window.h"
 #include "screenshot_parsing.h"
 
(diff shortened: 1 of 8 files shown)


vktrace/vktrace_replay/vkreplay_main.cpp
```diff
@@ -94,6 +94,13 @@ vktrace_SettingInfo g_settings_info[] = {
      {&replaySettings.compatibilityMode},
      TRUE,
      "Use compatibiltiy mode, i.e. convert memory indices to replay device indices, default is TRUE."},
+    {"p",
+     "Pause",
+     VKTRACE_SETTING_BOOL,
+     {&replaySettings.pause},
+     {&replaySettings.pause},
+     TRUE,
+     "Pause at the start of the replay."},
```
davidlunarg: This description of the new option should be consistent with the descritpion in the vktrace.md file

PR #556, opened 2018-06-25
vktrace: fix trim error on page guard handling

mizhen:
When starting trim, trim process need to read and save all image
and buffer data to trace file, that will trigger all page guard
on those memory if trim access shadow memory, and the status of
page guard for those memory will be different with capture without
trim. It causes corruption for some title if trim at some locations.
The change make trim use real memory pointer.

XCAP-1019
Change-Id: I9057ce04eb7ad68ee42d2f1ad9e25a09f1e42f46


vktrace/vktrace_layer/vktrace_lib_trace.cpp
```diff
@@ -247,7 +254,23 @@ VKTRACER_EXPORT VKAPI_ATTR VkResult VKAPI_CALL __HOOKED_vkMapMemory(VkDevice dev
...
+            // When starting trim, trim process need to read and save all image
+            // and buffer data to trace file, that will trigger all page guard
+            // on those memory if trim access shadow memory, and the status of
+            // page guard for those memory will be different with capture
+            // without trim. It causes corruption for some title if trim at
+            // some locations.
+            //
+            // So here we make trim to use real memory pointer to avoid change
+            // the pageguard status when PMB enabled.
+            pInfo->ObjectInfo.DeviceMemory.mappedAddress = pRealMappedData;
```
davidlunarg: Does mappedAddress get reset to the address of the shadow memory when we are done trimming and start tracing? It seems it would ned to be to properly track pmb changes after trimming is done. I couldn't find where that is done.

mizhen: The info in pInfo->ObjectInfo including DeviceMemory.mappedAddress is only used by trim module. the change is trim only, it doesn't need to reset after trim because PMB capture part has no change,  still use shadow memory pointer to track pmb change.

davidlunarg:
Merged to master

PR #565, opened 2018-07-03
vktrace: Fix error in trim 3d image handling

mizhen:
There's an error in trim saving data to staging buffer if the image is
a 3d image. It cause the trimmed trace file playback missing objects in
image. The change fix the problem.

XCAP-1019

Change-Id: I3f6bded038830aa890c2ac7ff8e248ebfb385130

### DIFF
--- a/vktrace/vktrace_layer/vktrace_lib_trim.cpp
+++ b/vktrace/vktrace_layer/vktrace_lib_trim.cpp
@@ -1009,7 +1009,7 @@ void snapshot_state_tracker() {
                         // buffer memory size.
 
                         copyRegion.bufferOffset = layout.offset;
-                        copyRegion.imageExtent.depth = 1;
+                        copyRegion.imageExtent.depth = imageIter->second.ObjectInfo.Image.extent.depth;
                         copyRegion.imageExtent.width = imageIter->second.ObjectInfo.Image.extent.width;
                         copyRegion.imageExtent.height = imageIter->second.ObjectInfo.Image.extent.height;
                         copyRegion.imageOffset.x = 0;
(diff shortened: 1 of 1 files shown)


vktrace/vktrace_layer/vktrace_lib_trim.cpp
```diff
@@ -1066,7 +1066,7 @@ void snapshot_state_tracker() {
                         copyRegion.bufferRowLength = 0;    //< tightly packed texels
                         copyRegion.bufferImageHeight = 0;  //< tightly packed texels
                         copyRegion.bufferOffset = lay.offset;
-                        copyRegion.imageExtent.depth = 1;
+                        copyRegion.imageExtent.depth = imageIter->second.ObjectInfo.Image.extent.depth >> i;
```
dustin-lunarg: Does this need to be 'max(1, imageIter->second.ObjectInfo.Image.extent.depth >> i)' to avoid setting copyRegion.imageExtent.depth to 0?  Same for width and height below.

mizhen: It doesn't need, these data are also depth, width and height of that image, trim get the data through capturing VkImageCreateInfo when target app call vkCreateImage, they are already verified at the image creation by the target running environment.

dustin-lunarg: This may be true for 3D images, but for 2D and 1D images, where depth is 1, the right shift will produce a depth of 0 for mip levels greater than 0.  This would also be the case for 1D images where height is equal to 1.

mizhen: Yes, you are right. Depth must be 1 for 2D image, depth and height must be 1 for 1D image ( https://www.khronos.org/registry/vulkan/specs/1.1-extensions/man/html/VkBufferImageCopy.html ). I'll submit another commit for it.

PR #573, opened 2018-07-12
vktrace: fix error in copying 1D/2D image

mizhen:
Add ImageType to track info, use this info when copying
1D/2D image in trim to fix an error which may cause depth
or height be wrongly set to zero.

XCAP-1019

Change-Id: I0034f5e0cf77668c3f74d0f11ec5f743f1976603


mizhen:
Original comment about the error is: https://github.com/LunarG/VulkanTools/pull/565#discussion_r201537416

dustin-lunarg:
This change looks good, but there are still cases where an image extent containing zero can be generated for the mip levels.  For example, the trim code that computes the values of copyRegion.imageExtent will produce the following extents for a 2D image with width = 2, height = 4, and mip levels = 3:

mip level 0 = 2x4x1
mip level 1 = 1x2x1
mip level 2 = 0x1x1

I would expect the extent for mip level 2 to be 1x1x1, which would require that the trim code clamp each dimension to 1.

PR #577, opened 2018-07-17
vkreplay: Fix issues replaying 32-bit large file

justuswang:
On 32-bit Linux/Android, fseek and ftell only works well for trace file
smaller than 2GB which will cause problem when geting file length and
using portability table on a large trace file.

This change fixes the issue by:

* x86 version:
** Use fseeko and ftello instead of fseek and ftell.
"-D_FILE_OFFSET_BITS=64" has been defined in vktrace's CMakeList.txt to
make those functions support large file.

* 32-bit Android version:
** Implement vktrace_fseek and vktrace_ftell using lseek64 which is
available in NDK.

For the NDK problem of using fseeko and ftello, refer to
https://github.com/android-ndk/ndk/issues/480

### DIFF
--- a/vktrace/vktrace_common/vktrace_common.h
+++ b/vktrace/vktrace_common/vktrace_common.h
@@ -72,9 +72,12 @@ static const uint32_t  INVALID_BINDING_INDEX = UINT32_MAX;
 #if defined(WIN32)
 #define Ftell _ftelli64
 #define Fseek _fseeki64
+#elif defined(ANDROID)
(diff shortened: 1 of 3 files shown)


davidlunarg:
The builds in this pull request failed because of a process problem - the Vulkan-Tools build was broken when this pull request was submitted. Now that the Vulkan-Tools build has been fixed, I have verified that this pull request in fact does build successfully.

I merged this change to master.

PR #596, opened 2018-07-31
vkreplay: Filter extensions in compatibility mode

justuswang:
This change checks and only uses available/supported instance/device
extensions in compatibility mode to make it possible for a trace file
to be replayed on a platform with less extensions.


davidlunarg:
@justuswang: Could you address the rebase conflicts and re-submit?

justuswang:
Sure, done.

PR #619, opened 2018-08-24
repo: Add known-good support

joey-lunarg:
Change-Id: I2ae79d5f0cdcd565b8728e0f960e0b1266f10df0

We have an almost circular dependency between VulkanTools and Vulkan-ValidationLayers. I believe since we do not recursively call update_deps.py, and the VT dependency for V-VL is travis-only, this should be fine.


scripts/known_good.json
```diff
@@ -0,0 +1,91 @@
...
+          "var_name" : "VULKAN_HEADERS_INSTALL_DIR",
+          "repo_name" : "Vulkan-Headers"
+        }
+      ],
+      "cmake_options" : [
+        "-DBUILD_TESTS=NO"
+      ]
+    },
+    {
+      "name" : "Vulkan-Tools",
```
karl-lunarg: I'm assuming VulkanTools needs Vulkan-Tools to get cube for use as a test program?

scripts/known_good.json
```diff
@@ -0,0 +1,91 @@
...
+        {
+          "var_name" : "VULKAN_LOADER_INSTALL_DIR",
+          "repo_name" : "Vulkan-Loader"
+        },
+        {
+          "var_name" : "VULKAN_TOOLS_INSTALL_DIR",
+          "repo_name" : "Vulkan-Tools"
+        }
+      ]
+    }
```
karl-lunarg: You might consider adding `-DBUILD_TESTS=NO` here (see how it is done in the Vulkan-Loader stanza if VulkanTools doesn't need to run the layer validation tests.  If VulkanTools is using the layer validation tests for something, then please consider if that is really needed.   Not building the validation tests will save a bunch of time and reduce dependencies.  It may not reduce dependencies enough to remove a repo from this list, but any reduction in dependencies is good.  It also eases the circular dependency issue a bit because it is these tests that need a component from VulkanTools.

karl-lunarg:
If this is working, then I'm good with it.  The circular dependency thing I don't think is too bad because it is more of a runtime dependency.  That is, if V-VL wants to run the layer validation tests with devsim, then it needs to build part of VulkanTools.  Tests don't run until after everything is built, so it doesn't matter if the pieces aren't all in place until the very end.

PR #623, opened 2018-08-30
layersvt: fix screen shot crash issue for some titles

tchiu8157:
layersvt: fix screen shot crash issue for some titles


Tony-LunarG:
Aren't there implementations that use separate queues for graphics and present?  The cube demo has code to handle that.  So in the Windows case, doesn't this code assume that if GetPhysicalDeviceWin32PresentationSupportKHR returns true, that we can send command buffers with copies or blits down the queue, when that might not be the case?  I think the code still needs to check for VK_QUEUE_GRAPHICS_BIT before accepting the queue.  Handling the case where there is a separate queue for graphics and present would be even better, but could be complicated.

tchiu8157:
Thanks for the note.  I'll give it a try and update accordingly.

PR #625, opened 2018-09-07
vktrace: add lock_guard for trim enabled capture

xooiamd:
This change adds lock/unlock for all hooked API only when trim is enabled.
When trim is enabled (g_trimEnabled is set to true),
calls like vkFlushMappedMemoryRanges, vkQueueSubmit and vkWaitForFences
are recorded to the trimmed trace file before the trim process generate
vkCreateDevice call, this causes remap error during playback
of the trimmed capture.
Lock/unlock added in this change will fix this race condition
and remap errors during playback.

VKTRACE-81

Change-Id: Ic0faaad2d45e858dc10d52be457e8ab40020ae78

### DIFF
--- a/scripts/vktrace_file_generator.py
+++ b/scripts/vktrace_file_generator.py
@@ -2463,6 +2463,7 @@ class VkTraceFileOutputGenerator(OutputGenerator):
                 ptr_packet_update_list = self.GetPacketPtrParamList(proto.members)
                 # End of function declaration portion, begin function body
                 trace_vk_src += ' {\n'
+                trace_vk_src += '    trim::TrimLockGuard<std::mutex> lock(g_mutex);\n'
                 if 'void' not in resulttype or '*' in resulttype:
                     trace_vk_src += '    %s result;\n' % resulttype
                     return_txt = 'result = '
--- a/vktrace/vktrace_layer/vktrace_lib_trace.cpp
+++ b/vktrace/vktrace_layer/vktrace_lib_trace.cpp
@@ -50,6 +50,8 @@
 
 #include "vk_struct_size_helper.h"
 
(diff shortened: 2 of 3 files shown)


davidlunarg:
There is a mutex that is acquired inside of CREATE_TRACE_PACKET and released in FINISH_TRACE_PACKET. CREATE_TRACE_PACKET and FINISH_TRACE_PACKET are called by most (all?) entrypoints. It seems this mutex would be sufficient. Or am I missing something?

xooiamd:
hi @davidlunarg , I just checked the CREATE_TRACE_PACKET (vktrace_create_trace_packet function [here](https://github.com/LunarG/VulkanTools/blob/master/vktrace/vktrace_common/vktrace_trace_packet_utils.c)) and FINISH_TRACE_PACKET (vktrace_finalize_trace_packet, vktrace_write_trace_packet, vktrace_delete_trace_packet in the same source file) functions and can't find the mutex, am I missing something? And I noticed that the FINISH_TRACE_PACKET seems to be called only when trim is not enabled, this pull request is mutex lock/unlock only when trim is enabled with the customized lockguard class. Thanks.

davidlunarg:
@xooiamd: Most entry points call CREATE_TRACE_PACKET, which calls vktrace_create_trace_packet, which calls vktrace_enter_critical_section. CREATE_TRACE_PACKET is called whether trimming is enabled or disabled.

When trimming is not enabled, the critical section is typically exited via a call to vktrace_leave_critical_ section in vktrace_delete_trace_packet.

If trimming is enabled, the critical section is typically exited when vktrace_delete_packet is directly called, or when trim::write_packet is called, which calls vktrace_delete_trace_packet.

PR #627, opened 2018-09-20
vktrace: Handle vkInvalidateMappedMemory call with range_size == VK_WHOLE_SIZE

jyangsh:
When the parameter pRange->size == VK_WHOLE_SIZE, it means to use the range from offset to the end of the current mapping of the allocation, but not the value(~0ULL) of the size.


davidlunarg:
The CI travis tests failed due to a profess problem which has since been fixed. Could you rebase to master or otherwise make a trivial change so as to force CI tests to run again?

jyangsh:
Rebased

PR #629, opened 2018-09-21
Reorganize VIA

MarkY-LunarG:
Reorganize VIA to separate OS-specific behavior into separate files/classes.

Also, add support (with @jeremy-lunarg) to determine if the new Linux install SDK target is present on the system.


MarkY-LunarG:
All changes ready, I should just be able to merge it, but adding @jeremy-lunarg and @lenny-lunarg for a quick review.

PR #630, opened 2018-09-21
vktrace: add process in command buffer recreation of trim

mizhen:
If trim starting at location where binding pipeline be destroyed and
command buffer still alive, trimmed trace file may include invalid
command calls in target command buffer recreation. Some of these
calls like vkCmdDrawIndexed cause crash on some hardware. The change
fix the problem.

VKTRACE-13

Change-Id: Ia6b8eb3abf059b77138b0b8542cb82b4f5cd4287


davidlunarg:
cherry-picked to master

PR #636, opened 2018-10-01
vkreplay: Removed skip of vkAllocMem replay when not bound

davidlunarg:
Removed skipping call to vkAllocMemory during replay when the
allocate memory is not bound to an image or buffer -- the memory
may still be mapped, written to, umapped, and freed, and those
operations would fail.

If the memory is not bound to an image or buffer, we assume it is
later mapped, so we make sure that the memory we allocate is
HOST_VISIBLE and HOST_COHERENT.

Change-Id: I6661b67d0515b5bbae723c70f1dd0d7c8fb359be


davidlunarg:
The changes seem bigger than they are - indentation changed on several chunks of code.

davidlunarg:
cherry-picked to master

PR #638, opened 2018-10-03
vktrace: append cxx flags to support cross compilation

ArsalanHAwan:
To support cross compilation, we need to set the variable
CMAKE_CXX_FLAGS_LIST such that the CXX flags set by user
for cross compilation are also included here.

Signed-off-by: Arsalan H. Awan <Arsalan_Awan@mentor.com>


ArsalanHAwan:
What if a particular SDK branch or more than one SDK branches do not implement the above commit, or this commit + a few additional commits that add the missing flags to other components of the VulkanTools, to allow users to add CXX flags for cross-compilation according to their toolchain? Can I open a PR against a specific SDK branch for that matter?

davidlunarg:
> What if a particular SDK branch or more than one SDK branches do not implement the above commit, or this commit + a few additional commits that add the missing flags to other components of the VulkanTools, to allow users to add CXX flags for cross-compilation according to their toolchain? Can I open a PR against a specific SDK branch for that matter?

We rarely make changes to an SDK branch, choosing instead to put changes on master so that they will be in the next SDK. The few times we have changed an SDK branch was because there was a serious issue in the SDK, and the change was followed up with a re-release of the SDK. I suggest that submitted PRs be for the master branch.

ArsalanHAwan:
> > What if a particular SDK branch or more than one SDK branches do not implement the above commit, or this commit + a few additional commits that add the missing flags to other components of the VulkanTools, to allow users to add CXX flags for cross-compilation according to their toolchain? Can I open a PR against a specific SDK branch for that matter?
> 
> We rarely make changes to an SDK branch, choosing instead to put changes on master so that they will be in the next SDK. The few times we have changed an SDK branch was because there was a serious issue in the SDK, and the change was followed up with a re-release of the SDK. I suggest that submitted PRs be for the master branch.

I understand. Thank you very much!

PR #642, opened 2018-10-11
vktrace: add mutex to protect fence,buffer and commandbuffer maps

mizhen:
Some title crash during capture because of multithreads simultaneous access
on fence, buffer and commandbuffer maps, the change add mutex to protect
these maps.

VKTRACE-86
Change-Id: I15f775c2193e5c780adc4c005a45f0e94d0ba2ad


davidlunarg:
There is a mutex that is acquired inside of CREATE_TRACE_PACKET and released in FINISH_TRACE_PACKET. CREATE_TRACE_PACKET and FINISH_TRACE_PACKET is called by most (all?) entrypoints. It seems this mutex would be sufficient. Or am I missing something?

PR #645, opened 2018-10-23
vktrace: Three commits about removing/adding lock to fix some capture/trim problems

mizhen:
The three commits in the pull request is used to fix several capture and trim problems for some titles.

### DIFF
--- a/vktrace/vktrace_common/vktrace_trace_packet_utils.c
+++ b/vktrace/vktrace_common/vktrace_trace_packet_utils.c
@@ -205,12 +205,6 @@ vktrace_trace_packet_header* vktrace_create_trace_packet(uint8_t tracer_id, uint
     return pHeader;
 }
 
-// Delete packet after vktrace_create_trace_packet being called.
-void vktrace_delete_trace_packet(vktrace_trace_packet_header** ppHeader) {
-    vktrace_delete_trace_packet_no_lock(ppHeader);
-
-}
-
 void* vktrace_trace_packet_get_new_buffer_address(vktrace_trace_packet_header* pHeader, uint64_t byteCount) {
     void* pBufferStart;
     assert(byteCount > 0);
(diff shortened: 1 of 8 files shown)


vktrace/vktrace_common/vktrace_trace_packet_utils.c
```diff
@@ -213,7 +209,6 @@ vktrace_trace_packet_header* vktrace_create_trace_packet(uint8_t tracer_id, uint
 void vktrace_delete_trace_packet(vktrace_trace_packet_header** ppHeader) {
     vktrace_delete_trace_packet_no_lock(ppHeader);
```
justuswang: Since the vktrace_delete_trace_packet is exactly the same as vktrace_delete_trace_packet_no_lock now, can we remove vktrace_delete_trace_packet_no_lock and rename all its references to vktrace_delete_trace_packet?

mizhen:
The first commit is added because some title still hang in our test when capturing with trim if using s_trace_lock.

vktrace/vktrace_common/vktrace_trace_packet_utils.c
mizhen: Yes, vktrace_delete_trace_packet_no_lock can be removed after the change.

davidlunarg:
But how does replacing one mutex with three mutexes change behavior?

mizhen:
There are some calls to function create_trace_packet in trim, trim use that to generate recreation commands for playback. With the s_trace_lock, the delete_trace_packet must follow any create_trace_packet, otherwise deadlock will happen.

davidlunarg:
Looks good. Please address the build failure on Linux and remove vktrace_delete_trace_packet_no_lock and rename all its references to vktrace_delete_trace_packet.

davidlunarg:
There's a clang formatting error that still needs to be addressed.

davidlunarg:
> There's a question with the CI build, it show the following error:
> 
> /home/travis/build/LunarG/VulkanTools/vktrace/vktrace_layer/vktrace_lib_trim.cpp:4505:9: error:
> use of undeclared identifier 'vktrace_delete_trace_packet_no_lock'; did
> you mean 'vktrace_delete_trace_packet'?
> vktrace_delete_trace_packet_no_lock(&(*iter));
> ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> check vktrace_lib_trim.cpp in the commit, there's no line 4505. Is Travis CI build based on a merged version?

I get the same error when I pull your repo/branch and try to build. And your commits are not the last commits, so, yes, I think CI uses the merged version.

mizhen:
The pull request is replaced with https://github.com/LunarG/VulkanTools/pull/669 and another pull request (will be created later) which include the needed change in reviewer's comment.

PR #653, opened 2018-10-30
vktrace: Add option for post process trim

justuswang:
This PR adds a environment variable VKTRACE_TRIM_POST_PROCESS and a vktrace option "-tpp" to enable/disable trim post processing.
Disable post processing means "write all referenced object calls" will happen in the trim start frame.
Enable post processing means "write all referenced object calls" will happen in the trim end frame and will NOT save objects NOT referenced in-trim.

- When VKTRACE_TRIM_POST_PROCESS is set to 1, trim post processing is enabled and the "write all referenced object calls" will happen in trim end frame.
- When VKTRACE_TRIM_POST_PROCESS is set to 0, trim post processing is disabled and the "write all referenced object calls" will happen in trim start frame.
- When VKTRACE_TRIM_POST_PROCESS is NOT set or set with other value, the "write all referenced object calls" will happen in trim start frame.

(The same as the behavior without this change.)

This change helps to reduce the trimmed trace file size.
It saves around 30% file size in average in our testing but it depends on the number of objects (especially shader modules, images and buffers) NOT referenced in-trim.

### DIFF
--- a/scripts/vktrace_file_generator.py
+++ b/scripts/vktrace_file_generator.py
@@ -1563,14 +1563,14 @@ class VkTraceFileOutputGenerator(OutputGenerator):
     # during the trim frames.
     def GenerateTrimRecordingInstructions(self, proto):
         trim_instructions = []
-        if 'CmdExecuteCommands' is proto.name:
+        if 'vkCmdExecuteCommands' == proto.name:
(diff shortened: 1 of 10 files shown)


davidlunarg:
cherry-picked to master

PR #658, opened 2018-11-01
vktrace: Keep the order of instance and swapchain creation in trim

justuswang:
This change records and keeps the creation order of instances and
swapchains in trim to make the trimmed trace file work when there are
multiple instances and swapchains.

It solves remap failures which are caused by the incorrect order of
instance creation or swapchain creation when replaying a trimmed trace
file.


davidlunarg:
@mizhen : Could you review this pull request?

PR #660, opened 2018-11-02
vkreplay: Fix vkGetDisplayPlaneSupportedDisplaysKHR

justuswang:
This change adds the missing VKTRACE_DELETE to the memory allocated
using VKTRACE_NEW_ARRAY.

It also adds the returned VkDisplayKHR to VkDisplayKHR map.

### DIFF
--- a/vktrace/vktrace_replay/vkreplay_vkreplay.cpp
+++ b/vktrace/vktrace_replay/vkreplay_vkreplay.cpp
@@ -4454,6 +4454,7 @@ VkResult vkReplay::manually_replay_vkBindImageMemory2KHR(packet_vkBindImageMemor
 }
 
 VkResult vkReplay::manually_replay_vkGetDisplayPlaneSupportedDisplaysKHR(packet_vkGetDisplayPlaneSupportedDisplaysKHR *pPacket) {
+    VkResult replayResult = VK_ERROR_VALIDATION_FAILED_EXT;
     VkPhysicalDevice remappedphysicalDevice = m_objMapper.remap_physicaldevices(pPacket->physicalDevice);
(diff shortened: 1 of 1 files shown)


vktrace/vktrace_replay/vkreplay_vkreplay.cpp
```diff
@@ -4462,28 +4463,26 @@ VkResult vkReplay::manually_replay_vkGetDisplayPlaneSupportedDisplaysKHR(packet_
...
-
-    VkDisplayKHR *displays = pPacket->pDisplays;
-
-    auto result = m_vkFuncs.GetDisplayPlaneSupportedDisplaysKHR(remappedphysicalDevice, pPacket->planeIndex, pPacket->pDisplayCount,
-                                                                remapped_displays);
-
-    if (memcmp(displays, pPacket->pDisplays, sizeof(VkDisplayKHR) * (*pPacket->pDisplayCount)) != 0) {
-        vktrace_LogError("Display Plane supported displays differ. Displays may not match as expected.");
+    if (pPacket->pDisplays != NULL) {
+        if (memcmp(pDisplays, pPacket->pDisplays, sizeof(VkDisplayKHR) * (*pPacket->pDisplayCount)) != 0) {
```
davidlunarg: I don't understand why this memcmp should be done. The values in pDisplays will be values from the replay device, the values in pPacket->pDisplays will be from the trace device. It seems an error would be generated a vast majority of the time when this code is executed. Have you seen this code execute without generating an error?

The old code also did the compare, but it looks like it was comparing the values returned to themselves.

justuswang: This piece of code is inherited from the old code.  Yes, you are correct, the memcmp should be removed since the values from display device is not expected to be the same as the values on trace device.

PR #661, opened 2018-11-02
vkreplay: back out previous commit

davidlunarg:
The prior commit, ad30962a4b4dad044adfce38643f14ff4a797aad, caused
a failure in LunarG's internal CI tests on Windows.

The failure was in a replay/trace/replay of Dota2, and in a replay
of an older trace of vkQuake.

The errors introduced by the previous commit are:

   vkreplay error: Return value VK_ERROR_OUT_OF_HOST_MEMORY from API call (vkQueuePresentKHR) does not match return value from trace file VK_SUCCESS.
   vkreplay error: Failed to replay packet_id 165, with global_packet_index 2686.
   vkreplay error: Return value VK_ERROR_VALIDATION_FAILED_EXT from API call (vkAcquireNextImageKHR) does not match return value from trace file VK_SUCCESS.
   vkreplay error: Failed to replay packet_id 164, with global_packet_index 2689.
   vkreplay error: API call vkQueueSubmit returned VK_ERROR_DEVICE_LOST. vkreplay cannot continue, exiting.

Backing out this commit for now until we can isolate the issue.

Change-Id: Ib8d45807e5f84dd02338bac1c522653a54458317

### DIFF
--- a/scripts/vktrace_file_generator.py
+++ b/scripts/vktrace_file_generator.py
@@ -789,7 +789,7 @@ class VkTraceFileOutputGenerator(OutputGenerator):
                     else:
                         replay_gen_source += '            %s local_%s;\n' % (params[-1].type.strip('*').replace('const ', ''), params[-1].name)
                 elif cmdname == 'ResetFences':
-                    replay_gen_source += '            VkFence* fences = (VkFence *)pPacket->pFences;\n'
+                    replay_gen_source += '            VkFence* fences = VKTRACE_NEW_ARRAY(VkFence, pPacket->fenceCount);\n'
                     replay_gen_source += '            for (uint32_t i = 0; i < pPacket->fenceCount; i++) {\n'
                     replay_gen_source += '                fences[i] = m_objMapper.remap_fences(pPacket->%s[i]);\n' % (params[-1].name)
                     replay_gen_source += '                if (fences[i] == VK_NULL_HANDLE) {\n'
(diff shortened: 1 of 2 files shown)


justuswang:
Hi @davidlunarg , is it possible for me to get the trace file you are using which is impacted by my commit?

I think using api_dump layer will able to help identify which piece of the code change causes the issue.
1. Replay the trace file without my commit and save api_dump to file_original (with "lunarg_api_dump.no_addr = TRUE" in the layer setting file)
2. Replay the trace file with my commit and save api dump to file_new (with "lunarg_api_dump.no_addr = TRUE" in the layer setting file)

If there's any difference in the file_original and file_new, that will be where the issue happens.

davidlunarg:
@justuswang: I tried dumping the api calls during replay before and after your change. I didn't see a difference in output prior to the VK_ERROR_OUT_OF_HOST_MEMORY error.

Here's a link to the trace file:
https://drive.google.com/file/d/1NiBP9ZqiuIdB4yAN06jp8-PKJRNvynA0/view?usp=sharing

justuswang:
@davidlunarg  I have found the problem is caused by the pWaitSemaphores in vkQueuePresentKHR is NULL in the trace file and I forgot to remove the checking of that pointer from the source code.
I have fixed it based on my pull request #652 using following patch.

```
$ git diff
diff --git a/vktrace/vktrace_replay/vkreplay_vkreplay.cpp b/vktrace/vktrace_replay/vkreplay_vkreplay.cpp
index a13f123..44ea34f 100644
--- a/vktrace/vktrace_replay/vkreplay_vkreplay.cpp
+++ b/vktrace/vktrace_replay/vkreplay_vkreplay.cpp
@@ -3182,7 +3182,7 @@ VkResult vkReplay::manually_replay_vkQueuePresentKHR(packet_vkQueuePresentKHR *p
         pResults = VKTRACE_NEW_ARRAY(VkResult, pPacket->pPresentInfo->swapchainCount);
     }

-    if (pRemappedSwapchains == NULL || pRemappedWaitSems == NULL || pResults == NULL) {
+    if (pResults == NULL) {
         replayResult = VK_ERROR_OUT_OF_HOST_MEMORY;
     }
```

PR #669, opened 2018-11-16
Fix capture and trim problems for some titles.

mizhen:
It Include the following changes: fixes for some titles capture and trim problem including remove s_trace_lock, delete vktrace_delete_trace_packet_no_lock and rename all references of the function to vktrace_delete_trace_packet, add lock to protect fence,buffer and command buffer maps.


mizhen:
The pull request is part of replacement of pull request #645

Issue #473, opened 2018-04-25
vktrace: multi-thread API calls are re-ordered in the trace


justuswang:
When tracing a mult-thread Vulkan application, I have found that some API-calls are re-ordered in the trace file because current vktrace is recording API calls in completion order instead of issue order.  Packet with smaller global_packet_index sometimes is not stored before a packet with larger global_packet_index.

The observation is:

Thread 1: (start create image#1 ... end) -> (start destroy image#1 .............. end) --------> (other API)
Thread 2: ---------------------------------------> (start create image#1 ... end) ------> (start destroy image#1 ... end)

There's a vkDestroyImage being called after image#1 (a VkImage handle which does not mean identical image in thread#1 and thread#2) was destroyed on a different thread.

vkDestroyImage in thread#1 is called before vkCreateImage being called in thread#2 but completed after vkCreateImage in thread#2 being completed.

vkDestroyImage in thread#1 takes longer time than vkCreatImage in thread#2 which makes it being saved to trace file as a trace packet after the trace packet for vkCreateImage in thread#1.

The issue happens when tracing a multi-thread Vulkan application, every API call will be saved to trace file in their completion order instead of issue order which will cause unexpected overlap.

Is this a bug of vktrace?  Does vktrace need to have a mutex lock in every API call?  e.g. Add vktrace_enter_critical_section at the beginning of an API call and vktrace_leave_critical_section at the end of the API call.

davidlunarg:
Yes, this is a bug in vktrace. The api calls should be added to the trace file in the order they are made. A mutex would solve the issue, but might negatively affect performance. Another solution might be to break up the packets into a "call packet" and a "result packet", in which the call packet is added to the trace file before the trace layer calls down to the next layer, and the result packet would be added after the call to the next layer returns. We're currently looking at other changes to vktrace in which such a strategy might be used. It will be a while before we make those changes, so maybe we want to implement the mutex solutions in the meantime?

justuswang:
> We're currently looking at other changes to vktrace in which such a strategy might be used. It will be a while before we make those changes, so maybe we want to implement the mutex solutions in the meantime?

Since those changes won't happen in a short time, shall I create a pull request to implement the mutex to fix it for now?  Though it might negatively affect performance.  I think it's better to have a trace file with correct order than have a better performance when tracing an app.  And I prefer to have the fix ASAP.

davidlunarg:
[...]

PR #675, opened 2018-11-23
Add mutex lock for all manual and autogenerated hooked API when trim capture is enabled.

mizhen:
Add mutex lock for all manual and autogenerated hooked API when trim capture is enabled.
declare extern mutex variable in vktrace_lib_trim header file.

### DIFF
--- a/scripts/vktrace_file_generator.py
+++ b/scripts/vktrace_file_generator.py
@@ -2574,6 +2574,7 @@ class VkTraceFileOutputGenerator(OutputGenerator):
                 ptr_packet_update_list = self.GetPacketPtrParamList(proto.members)
                 # End of function declaration portion, begin function body
                 trace_vk_src += ' {\n'
+                trace_vk_src += '    trim::TrimLockGuard<std::mutex> lock(g_mutex);\n'
                 if 'void' not in resulttype or '*' in resulttype:
                     trace_vk_src += '    %s result;\n' % resulttype
                     return_txt = 'result = '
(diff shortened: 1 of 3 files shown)


mizhen:
The pull request is second part of replacement of pull request #645, the first part of the replacement is https://github.com/LunarG/VulkanTools/pull/669

justuswang:
@mizhen, is it possible to always have the mutex lock no matter trim capture is enabled or not?
Otherwise issue #473 "multi-thread API calls are re-ordered in trace file" will happen again.

mizhen:
Yes. If considering enable the lock when capturing without trim, an environment variable might be needed as a switch.

davidlunarg:
It seems it would be better not to use an environment variable to address a problem such as reordering of api calls.

xooiamd:
Hi, there is some performance drop if always have mutex lock, which is why we enable it only during trim capture. But we could add an option in vktrace to turn it  on/off. Any comments/advice? Thanks

davidlunarg:
It looks like the current set of commits enable the use of a mutex for all api calls. Is this correct?

To enable the use of a mutex,  an option can be added to vktrace.

xooiamd:
hi, the current commit is using a customized mutex lock guard class (named TrimLockGuard) which is declared in vktrace_lib_trim.h file. This class will only lock/unlock mutex when g_trimEnabled is set to true. Although all api calls has TrimLockGuard, but the use of mutex is only enabled by g_trimEnabled. 
Yes, we can add an option to enable the use of mutex.

mizhen:
The pull request is replaced by https://github.com/LunarG/VulkanTools/pull/718

PR #680, opened 2018-12-04
Build script

xykivo:
Hi

This pull request contains
1. Added a build script to simplify dependency updates + builds on non Android platforms.
2. Changes to the root cmake file so it will auto update dependencies when a cmake build is called.

Thanks
 Dror


davidlunarg:
Closing this PR due to inactivity. If desired, this change can be submitted again as a new PR.

PR #681, opened 2018-12-07
Update Android known goods and Android build

Tony-LunarG:
Updating known goods to avoid the gnustl_static flag that is no longer supported.  As a result, I needed to update V-VL known good, and as a result of that, I needed to port V-VL Android build changes to this repo


cnorthrop:
Do we need to maintain the validation layer build in this repo?  Or does that facilitate the other layers we do build here?

Tony-LunarG:
I think it's needed to build the trace layer.  There is a longer term task to untangle this repo from the others, this is a short term fix to get LunarG's internal CI running and passing again.

PR #683, opened 2018-12-10
VKVIA: CMakeLists-SDK, Allow make install for via in the SDK

johnzupin:
Allow make install via in the SDK using the build_tools.sh script


MarkY-LunarG:
I know nothing about install.  @jeremy-lunarg might be better.

PR #684, opened 2018-12-10
build: Allow make install on Unix for vkconfig

johnzupin:
Allow for `make install` for vkconfig.

### DIFF
--- a/vkconfig/CMakeLists.txt
+++ b/vkconfig/CMakeLists.txt
@@ -33,4 +33,8 @@ else()
     else()
         include(macOS/vkconfig.cmake)
     endif()
+
+    if(UNIX)
+        install(TARGETS vkconfig DESTINATION ${CMAKE_INSTALL_BINDIR})
+    endif()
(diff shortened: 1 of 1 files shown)


johnzupin:
This pull request allows the build_tools.sh script in the SDK to do a `make install` instead of copying vkconfig to the bin dir.

vkconfig/CMakeLists.txt
```diff
@@ -2,6 +2,10 @@
 cmake_minimum_required(VERSION 3.4)
 project(vkconfig)
 
+if (NOT DEFINED CMAKE_INSTALL_BINDIR)
+  include(GNUInstallDirs)
+endif()
+
 set(CMAKE_CXX_STANDARD 11)
```
johnzupin: Maybe this part isn't needed as I am guarenteed to define CMAKE_INSTALL_PREFIX and CMAKE_INSTALL_BINDIR from the commandline.

PR #692, opened 2019-01-03
vkreplay: Optimal image memory allocation enhancement in compatibility mode

justuswang:
This change makes vkreplay have better compatibility when the replay
platform has different memory requirement for optimal tiling image
than the trace platform which cannot be calculated using alignment.

vkreplay will skip vkAllocateMemory if the memory allocation is only
for optimal tiling image(s) and do memory allocation in
vkBindImageMemory (and memory free in vkDestroyImage) when
CompatibilityMode is enabled.

### DIFF
--- a/vktrace/vktrace_replay/vkreplay_vkreplay.cpp
+++ b/vktrace/vktrace_replay/vkreplay_vkreplay.cpp
@@ -512,12 +512,12 @@ VkResult vkReplay::manually_replay_vkCreateDevice(packet_vkCreateDevice *pPacket
     VkExtensionProperties *extensions = NULL;
     if (g_pReplaySettings->compatibilityMode) {
         if (VK_SUCCESS != m_vkFuncs.EnumerateDeviceExtensionProperties(remappedPhysicalDevice, NULL, &extensionCount, NULL)) {
-            vktrace_LogError("vkEnumerateInstanceExtensionProperties failed to get extension count!");
+            vktrace_LogError("vkEnumerateDeviceExtensionProperties failed to get extension count!");
         } else {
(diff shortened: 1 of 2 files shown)


vktrace/vktrace_replay/vkreplay_vkreplay.cpp
```diff
@@ -1885,157 +1898,148 @@ void vkReplay::modifyMemoryTypeIndexInAllocateMemoryPacket(VkDevice remappedDevi
...
-        for (i = 0; i < replayMemoryProperties[replayPhysicalDevice].memoryTypeCount; i++) {
-            if (traceMemoryProperties[tracePhysicalDevice].memoryTypes[pPacket->pAllocateInfo->memoryTypeIndex].propertyFlags ==
-                replayMemoryProperties[replayPhysicalDevice].memoryTypes[i].propertyFlags) {
-                *((uint32_t *)&pPacket->pAllocateInfo->memoryTypeIndex) = i;
-                break;
-            }
-        }
+        // This isn't an error - the memory is allocated but never used.
+        // So just use the index from the trace file and continue.
+        goto out;
```
davidlunarg: What if the index in the trace file is not valid on the replay device? The bad index will be used in the vkAllocateMemory and generate an error. Shouldn't an attempt be made to find a valid index?

justuswang: This is an error of the comments.  The memory allocated but never used will be skipped vkAM.
This helps to skip the vkAM calls for optimal images which will have their own vkAM call in replaying vkBindImageMemory().

I will submit a commit to correct the comments.

PR #695, opened 2019-01-03
Add multi-client support for vktrace socket handling

mizhen:
Enable vktrace support multi-process capturing at same time. Adding this
support is to fix the crash problem caused by the case that one title has
several processes running.


davidlunarg:
I haven't looked at the code in detail yet, but I tried out this change on Linux.

If I start vktrace in server mode, then run two Vulkan programs simultaneously with the trace layer enabled, two files are created: vktrace_out.vktrace and vktrace_out.vktrace-1. It seems that both files should have the same extension (.vktrace), and that the second file should be named vktrace_out-1.vktrace.

If I start vktrace in server mode, run cube from another term window with the trace layer enabled and let it run a while, then kill the cube program, the vktrace server dies with the error message:
"vktrace error: Host: Failed binding socket err=98. Segmentation fault (core dumped)"

davidlunarg:
This pull request introduces three new compile warnings:

VulkanTools/vktrace/vktrace_trace/vktrace.cpp:569:27: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
VulkanTools/vktrace/vktrace_trace/vktrace.cpp:580:68: warning: NULL used in arithmetic [-Wpointer-arith]
VulkanTools/vktrace/vktrace_trace/vktrace_process.cpp:68:23: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]

davidlunarg:
Started vktrace in server mode in background:
   $ vktrace &
Enabled trace layer and ran cube in background:
   $ export VK_INSTANCE_LAYERS=VK_LAYER_LUNARG_vktrace
   $ vkcube &
vkcube start to trace. Tried starting a second cube, get an error:
   $ vkcube
   vktrace error: Client: Couldn't find any connections.

I would have expected the second vkcube to trace successfully.
I am still looking at this, making sure it is not a user error.

davidlunarg:
I'm puzzled on the source of the CI errors. I don't think they are due to this pull request.

PR #696, opened 2019-01-04
vktrace: Add a new tool vktracedump

justuswang:
This change adds an offline api dump tool vktracedump to dump api from
existing Vulkan trace file.

It utilizes the auto generated source code of API dump layer and depends on #535 to make "-ds" option (dump shader to files) work as expected.


davidlunarg:
I like this new tool, and would like to eventually add it to VulkanTools.

Before merging this pull request, let's address the following:

1. I wonder if there might be a better name for the tool than vkparse. The name seems too generic - it doesn't indicate what the tool parses. How about vktraceparse? vktracedump? Maybe other folks reading this might have further suggestions for a name.

1. Add the capability to send output to stdout. This might be useful when processing large traces so that a large intermediate file is not created.

1. The resolution of a window can change, and the app may have created multiple windows, so the resolution printed might be misleading. Maybe it is better to omit the resolution?

1.  It needs to build and run on all platforms. - the Windows build failed for this pull request.

I haven't looked at the actual code changes yet, I'll do that after these higher level issue are addressed.

davidlunarg:
The help text seems to indicate that the -ds option dumps shader code to a separate file, but it is dumped to the same file as other output.

justuswang:
> The help text seems to indicate that the -ds option dumps shader code to a separate file, but it is dumped to the same file as other output.

This feature (dump shaders to separate files) depends on #535.  Because it is using API dump layer's source code to do "full API dump" and the change to dump shaders to separate files needs to be applied to API dump layer.  Not sure if I should include the code change of #535 in this pull request?

davidlunarg:
No, don't include #535 in this pull request. They should be kept separate. Let's address the other issues, and in parallel, I'll grease the wheels to get #535 reviewed and merged. If for some reason #535 is not approved, this pull request will have to be modified to not depend on #535.

justuswang:
> 1. I wonder if there might be a better name for the tool than vkparse. The name seems too generic - it doesn't indicate what the tool parses. How about vktraceparse? vktracedump? Maybe other folks reading this might have further suggestions for a name.

vktraceparser sounds good to me.  Changed to vktraceparser for now until getting any further suggestions.

> 2. Add the capability to send output to stdout. This might be useful when processing large traces so that a large intermediate file is not created.

Done.  "-s stdout" will send output of simple dump to stdout and "-f stdout" will send full dump to stdout.

> 3. The resolution of a window can change, and the app may have created multiple windows, so the resolution printed might be misleading. Maybe it is better to omit the resolution?

Done. Yes, it is better to omit the resolution to avoid confusion.

> 4. It needs to build and run on all platforms. - the Windows build failed for this pull request.

Done.  Windows build is fixed.

davidlunarg:
I have discussed the name of this new tool within the vktrace team and we decided it should be vktracedump.

Copyright dates need to be 2019. Any file that is modified that has a copyright date needs to add 2019.

The vktrace.md file needs to have a description of the command line options, like the descriptions for vktrace and vkreplay.

davidlunarg:
I merge master into this pull request branch on my local system. Note that master includes pull request #535. I noted several problems:

"vktracedump -o c.vktrace -f file" sends output to stdout
"vktracedump -o c.vktrace -na -f file" sends output to stdout, and the -na option seems to have no effect.
"vktracedump -o c.vktrace -dh -f file" send output to stdout, and the -dh option seems to have no effect.

justuswang:
The problems are caused by legacy dump configuration file not removed (if the vktracedump was terminated by Ctrl-C).
I have uploaded a new commit to address those problems by removing legacy configuration file before doing dump.

davidlunarg:
The command:

vktracedump -o c.vktrace  -f stdout -dh >/tmp/t.html

Results in strangely formatted text when the html file is viewed in a browser. (This text contains the file version, file type, etc from the trace file header.) It would be nice if this information was incorporated into the html file.

Frame numbers in the html output are formatted as hex numbers. It seems decimal might be better. Is this because api_dump also outputs frame numbers in hex? Probably best to stay consistent with api_dump.

justuswang:
> Results in strangely formatted text when the html file is viewed in a browser. (This text contains the file version, file type, etc from the trace file header.) It would be nice if this information was incorporated into the html file.
> 
> Frame numbers in the html output are formatted as hex numbers. It seems decimal might be better. Is this because api_dump also outputs frame numbers in hex? Probably best to stay consistent with api_dump.

Maybe it is better to hide the trace file header info (file version, file type, etc) when the output of trace dump is sending to stdout to make the stdout consistent with the output of api dump layer?
The change to hide those info from stdout fixes both problems.  Please have a try with the latest change.

PR #703, opened 2019-01-18
vkreplay: Another try to preload trace file to replay

justuswang:
This is an experimental change to make the vkreplay run faster.

This change adds a new option "-pltf" to help user to pre-load trace
file to memory when running vkreplay. (with "-pltf true")

It helps to replay a trace file faster especially on a system with slow
file I/O.


justuswang:
It is not necessary to merge this change because it may NOT have any performance improvement on a system where memory may be swapped out to disk.

It brings around 5% performance improvement on my test platform (Android OS development board).  But it does not have that much improvement when testing on a Ubuntu PC.

davidlunarg:
Closing without merging as requested by submitter.

PR #710, opened 2019-01-24
Bob 119 fix png warning

lunarpapillo:


mikew-lunarg:
The title of this PR seems quite unrelated to the changes.
This PR appears to be documentation tweaks.

lunarpapillo:
> The title of this PR seems quite unrelated to the changes.
> This PR appears to be documentation tweaks.

You're not wrong... the one binary file is the one that fixes the issue.  The rest are documentation tweaks that I thought had been merged into master previously - I may have been mistaken, and the earlier changes might not have been merged yet.

I think they're still all good, and should be in master, even if the process to get them there wasn't as tidy as I would have liked.

PR #713, opened 2019-01-24
vkreplay: Always use playable device features

justuswang:
This change makes vkreplay always use playable device features in
compatibility mode.


vktrace/vktrace_replay/vkreplay_vkreplay.cpp
```diff
@@ -696,6 +697,11 @@ VkResult vkReplay::manually_replay_vkCreateDevice(packet_vkCreateDevice *pPacket
                 extensionCount = 0;
             }
         }
+        if (pCreateInfo->pEnabledFeatures) {
+            m_vkFuncs.GetPhysicalDeviceFeatures(remappedPhysicalDevice, &features);
+            // Use replayable features instead of traced features
+            pCreateInfo->pEnabledFeatures = &features;
```
davidlunarg: Good find. Replay shouldn't attempt to enable features from the trace device that are not available on the replay device.

But shouldn't only those features that are available on the trace device be enabled on the replay device? In other words, should the features enabled on replay be the "intersection" of the features requested on the trace device and the features available on the replay device? It seems like this should be the case. What do others think?

PR #716, opened 2019-01-25
vkreplay: Skip replaying failed api calls in trace

justuswang:
This change makes vkreplay to skip replaying the API calls with failed
results recorded in trace file.  The expected to fail API calls in
trace file will not be replayed in case those API calls may success on
replay platform.

### DIFF
--- a/scripts/vktrace_file_generator.py
+++ b/scripts/vktrace_file_generator.py
@@ -702,6 +702,17 @@ class VkTraceFileOutputGenerator(OutputGenerator):
             if protect == "VK_USE_PLATFORM_XLIB_XRANDR_EXT":
                 replay_gen_source += '#endif // %s\n' % protect
         replay_gen_source += '}\n\n'
+        replay_gen_source += 'bool vkReplay::callFailedDuringTrace(VkResult result, uint16_t packet_id) {\n'
+        replay_gen_source += '    if (result && g_pReplaySettings->compatibilityMode && m_pFileHeader->portability_table_valid && !platformMatch() &&\n'
+        replay_gen_source += '        result != VK_SUCCESS && result != VK_NOT_READY && result != VK_TIMEOUT && result != VK_EVENT_SET &&\n'
+        replay_gen_source += '        result != VK_EVENT_RESET && result != VK_INCOMPLETE && result != VK_SUBOPTIMAL_KHR) {\n'
+        replay_gen_source += '        vktrace_LogVerbose("Skip %s which has failed result in trace file!",\n'
+        replay_gen_source += '                           vktrace_vk_packet_id_name((VKTRACE_TRACE_PACKET_ID_VK)packet_id));\n'
+        replay_gen_source += '        return true;\n'
+        replay_gen_source += '    } else {\n'
+        replay_gen_source += '        return false;\n'
+        replay_gen_source += '    }\n'
+        replay_gen_source += '}\n\n'
         replay_gen_source += 'vktrace_replay::VKTRACE_REPLAY_RESULT vkReplay::replay(vktrace_trace_packet_header *packet) { \n'
         replay_gen_source += '    vktrace_replay::VKTRACE_REPLAY_RESULT returnValue = vktrace_replay::VKTRACE_REPLAY_SUCCESS;\n'
         replay_gen_source += '    VkResult replayResult = VK_SUCCESS;\n'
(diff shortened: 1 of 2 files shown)


justuswang:
Yes, it makes sense to only skip the failed API calls in portability mode.  A new commit has been uploaded to address this.

scripts/vktrace_file_generator.py
```diff
@@ -734,6 +734,14 @@ def GenerateReplayGenSource(self):
...
             replay_gen_source += '            packet_vk%s* pPacket = (packet_vk%s*)(packet->pBody);\n' % (cmdname, cmdname)
+            if resulttype is not None and resulttype.text != 'void' and resulttype.text != 'PFN_vkVoidFunction':
+                replay_gen_source += '            if (pPacket->result &&\n'
+                replay_gen_source += '                g_pReplaySettings->compatibilityMode && m_pFileHeader->portability_table_valid && !platformMatch() &&\n'
+                replay_gen_source += '                pPacket->result != VK_SUCCESS && pPacket->result != VK_NOT_READY && pPacket->result != VK_TIMEOUT && pPacket->result != VK_EVENT_SET &&\n'
+                replay_gen_source += '                pPacket->result != VK_EVENT_RESET && pPacket->result != VK_INCOMPLETE && pPacket->result != VK_SUBOPTIMAL_KHR) {\n'
+                replay_gen_source += '                vktrace_LogVerbose("Skip %s which has failed result in trace file!", vktrace_vk_packet_id_name((VKTRACE_TRACE_PACKET_ID_VK)packet->packet_id));\n'
+                replay_gen_source += '                break;\n'
+                replay_gen_source += '            }\n'
```
davidlunarg: This complex expression is repeated many times in the generated code. Might it be a good idea to make it a function? 
The generated code look something like:

`if (callFailedDuringTrace()) break;`

The function callFailedDuringTrace would check for the condition in which the api call should be skipped and call vktrace_LogVerbose if it is skipped.

justuswang: Maybe the vktrace_LogVerbose should not be included in the new function and the new function should only check a result is success code or error code, no matter the result is trace result or replay result?

Here's my proposal:
New function:
```
bool isSuccessReturnCode(VkResult result) {
    return (result == VK_SUCCESS || result == VK_NOT_READY || result == VK_TIMEOUT || result == VK_EVENT_SET || result == VK_EVENT_RESET || result == VK_INCOMPLETE || result == VK_SUBOPTIMAL_KHR);
}
```

and the generated code will look like:
```
if (pPacket->result && 
    g_pReplaySettings->compatibilityMode && m_pFileHeader->portability_table_valid && 
!platformMatch() &&
    !isSuccessReturnCode(pPacket->result) {
    vktrace_LogVerbose("Skip %s which has failed result in trace file!", vktrace_vk_packet_id_name((VKTRACE_TRACE_PACKET_ID_VK)packet->packet_id));\n'
    break;
}
```

Maybe 

> g_pReplaySettings->compatibilityMode && m_pFileHeader->portability_table_valid && 
> !platformMatch()

 should also be made a function?

davidlunarg: My motivation for suggesting an if statement with a function call and a break is to reduce the code size, both the generated C code and the resulting object code. This sequence is repeated for every API call, a compiler might identify the identical code in each case of the switch statement, but we can't be sure that it will. I think the function callFailedDuringTrace that I am proposing would need to be passed a pointer to the packet.

The code:
`g_pReplaySettings->compatibilityMode && m_pFileHeader->portability_table_valid &&
!platformMatch()`
could be made into a function for readability of this file and other files that use that sequence, but it is not necessarily within the scope of this pull request.

justuswang: Seems it will need to parse each type of packet to get the "result" value if passing a pointer to the packet to callFalledDuringTrace() because the pPacket->result's location in each packet varies among different packet types.  Each type of packet needs to have `packet_vk%s* pPacket = (packet_vk%s*)(packet->pBody);` to make it possible to use pPacket->result.

I have made a change to pass result value and packet id instead of a pointer to packet to callFailedDuringTrace(), please take a look.

davidlunarg: You're right about the result not being in a consistent offset within the packet, so the result and packet id need to be passed as parameters. Change looks good.

PR #726, opened 2019-02-14
vkreplay: search full portability table

davidlunarg:
Previously used to keep the starting position for portability table
searches in a static variable. This would break looping, because
after the 1st loop, the starting positon was at the end of the
portablity table, so vkAllocateMemory packets were not being
found in the portability table. Changed the code to always search
from the beginning.

Change-Id: Iebaaeec568f2a3d80a73c3d8fda168b0c76decf2


davidlunarg:
@cforfang : I like you suggestion. Thanks. I will implement that.

PR #731, opened 2019-02-20
tests: disable devsim test that depends on vulkaninfo output

davidlunarg:
Doing this temporarily, until either golden image can be updated or
vulkaninfo is added to VulkanTools/tests so as to avoid changes to
Vulkan-Tools/vulkaninfo affecting VulkanTools tests.

Change-Id: Ic46e7909e729afc071d4730b3dfa704a7f756eae


davidlunarg:
> I presume this temporary change is just until we can verify the different output, and establish a new golden .json?

Yep.

PR #736, opened 2019-02-25
Fixed undefined symbols PRIu64/PRIu32 and reference to clock_gettime

francoisbertelatschrodinger:
On CentOS6.8 with a local build of g++ 6.4.0 (the default g++ version on CentOS6.8 is 4.4.7), the build fails with undefined symbol PRIu64 and PRIu32 at compile time and undefined reference to 'clock_gettime' at link time.

Fixed undefined symbols PRIu64/PRIu32 by including `<cinttypes>` instead of
`<inttypes.h>`. (a less elegant solution would be to add `#define __STDC_FORMAT_MACROS` prior to include `<inttypes.h>`). `layersvt/device_simulation.cpp` is a C++ file and as such should include C header files that way anyway (note that this file includes other C header files the wrong way: `<stdarg.h>`, `<stdio.h>`, `<stdlib.h>` and `<string.h>` instead of `<cstdarg>`, `<cstdio>`, `<cstdlib>` and `<cstring>` respectively. I have not fixed these inclusions in the current patch as I want to keep it minimal to address the actual build issue)

Fixed undefined reference to clock_gettime by linking rt library
This fix is similar to https://github.com/KhronosGroup/SPIRV-Tools/pull/2409

### DIFF
--- a/layersvt/device_simulation.cpp
+++ b/layersvt/device_simulation.cpp
@@ -40,7 +40,7 @@
 #include <stdarg.h>
 #include <stdio.h>
 #include <stdlib.h>
-#include <inttypes.h>
+#include <cinttypes>
 #include <string.h>
 
 #include <functional>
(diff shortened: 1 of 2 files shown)


francoisbertelatschrodinger:
In case you have not been notified automatically, I signed the CLA on Feb 28 2019.

Issue #733, opened 2019-02-21
vkreplay prints bad frame range if number of frames is 0


davidlunarg:
If a trace file has 0 frames, i.e. zero calls to vkQueuePresentKHR, the framerange printed by vkreplay is incorrect:

$ ./vkreplay.exe  -t x.vktrace
vkreplay info: 0.000000 fps, 0.452336 seconds, 0 frame, 1 loop, framerange 0--1

Should probably just skip printing the framerange for this case.

davidlunarg:
Fix in pull request #738

PR #738, opened 2019-02-25
vkreplay: remove print of framerange when trace has 0 frames

davidlunarg:
Change-Id: I1d735aa33ef5a3bfd119513fa08a7b71719c631e


davidlunarg:
This pull request address issue #733

davidlunarg:
> Change looks good, excluding the formatting issue that failed the Travis build. Is there a need to print anything if the trace has no frames?

A Vulkan program doesn't *have* to have a call to vkQueuePresent for it to be valid. For example, vulkaninfo. And maybe someone is interested in how long it took to do the api calls in vulkaninfo.

PR #741, opened 2019-02-27
api_dump: Add Environment variable support

MarkY-LunarG:
Sometimes, it's easier to modify settings using
environment variables instead of using the settings
file.  So, add support for some of the primary
API Dump layer settings using environment variables.
Those settings environment variables are:

 * APIDUMP_LOG_FILENAME  - Set the output filename.  If set, also
                           automatically retargets the output to this
                           file.
 * APIDUMP_OUTPUT_FORMAT - Set the output format to either "html" or
                           "text".
 * APIDUMP_DETAILED      - Boolean to enable the detailed output
 * APIDUMP_FLUSH         - Boolean to disable flushing after every write
 * APIDUMP_NO_ADDR       - Boolean to disable address outputs in the
                           API Dump output file.

Change-Id: I110609c48fb615d0eeae623c7de3176c65bd935d

### DIFF
--- a/layersvt/api_dump.h
+++ b/layersvt/api_dump.h
@@ -1,6 +1,6 @@
-/* Copyright (c) 2015-2016 The Khronos Group Inc.
- * Copyright (c) 2015-2016 Valve Corporation
- * Copyright (c) 2015-2016 LunarG, Inc.
+/* Copyright (c) 2015-2019 The Khronos Group Inc.
+ * Copyright (c) 2015-2019 Valve Corporation
+ * Copyright (c) 2015-2019 LunarG, Inc.
  * Copyright (C) 2015-2016 Google Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -47,6 +47,13 @@
 
 #define MAX_STRING_LENGTH 1024
 
+// Defines for utilized environment variables.
+#define API_DUMP_ENV_VAR_LOG_FILE "VK_APIDUMP_LOG_FILENAME"
+#define API_DUMP_ENV_VAR_OUTPUT_FMT "VK_APIDUMP_OUTPUT_FORMAT"
+#define API_DUMP_ENV_VAR_DETAILED_OUTPUT "VK_APIDUMP_DETAILED"
+#define API_DUMP_ENV_VAR_NO_ADDRESSES "VK_APIDUMP_NO_ADDR"
+#define API_DUMP_ENV_VAR_FLUSH_FILE "VK_APIDUMP_FLUSH"
+
 enum class ApiDumpFormat {
     Text,
     Html,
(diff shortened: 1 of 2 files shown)


KarenGhavam-lunarG:
Or is this something we would want to add to vkconfig? To get it to be the "control panel" of use?

MarkY-LunarG:
No.  This is entirely to support Android.  The functionality is already there for Windows/Linux.

MarkY-LunarG:
Updating commit message to clarify Android target.  I thought it was already there but based on @KarenGhavam-lunarG confusion it seems clear that I didn't have any mention of Android in it.

KarenGhavam-lunarG:
@MarkY-LunarG Pardon my limited understanding. Isn't there a configuration file somewhere where you can place settings for a layer that then get passed on to the layer?

KarenGhavam-lunarG:
Never mind my previous comment. I now understand that on Android it is difficult to use the layer settings file.

davidlunarg:
Might we want to change the names of the environment variables to begin with "VK"? For example change APIDUMP_LOG_FILENAME to VK_APIDUMP_LOG_FILENAME. There might be other APIs that the user is trying to dump, this would avoid potential conflicts.

Screenshot and trace layers use env variables that begin with "VK".

layersvt/api_dump.h
```diff
@@ -55,30 +55,69 @@ enum class ApiDumpFormat {
...
+                filename_string = filename_option;
             } else {
-                output_stream.open("vk_apidump.txt", std::ofstream::out | std::ostream::trunc);
+                filename_string = "vk_apidump.txt";
+            }
+        }
+        // If an environment variable is set, always output to that filename instead,
+        // whether or not the settings file enables the option.  Just assume a non-empty
+        // string is asking for the file output to the given name.
+        std::string env_value = GetPlatformEnvVar("APIDUMP_LOG_FILENAME");
```
davidlunarg: Might want to define the env variable names in one place, in a single header file. It's easier to find the list of env variables that this layer uses. And we should avoid the use of literal constants in the code. (Not that the rest of this code avoids literal constants. :-)

layersvt/api_dump.h
```diff
@@ -55,30 +55,69 @@ enum class ApiDumpFormat {
...
+        if (!env_value.empty()) {
+            filename_string = env_value;
+        }
+        // If one of the above has set a filename, open the file as an output stream.
+        if (!filename_string.empty()) {
+            use_cout = false;
+            output_stream.open(filename_string, std::ofstream::out | std::ostream::trunc);
+            size_t last_slash_idx = filename_string.find_last_of("\\/");
```
davidlunarg: What about  "\\" used to in Windows path names?

MarkY-LunarG: This looks for the last of either "\" or "/"  the "\\" translates to "\".

davidlunarg: Doh! Misread the code.

layersvt/api_dump.h
MarkY-LunarG: I'm not sure how you intend to not have literals.  Even if you #define it, it still ends up as a literal.  Plus, we already have these values for the file options.  Also, this is open-source so it's not like we're trying to hide any strings.My argument is I don't think moving them makes them any easier to find since you could just search for `GetPlatformEnvVar` or look at the documentation to find out what they are.

MarkY-LunarG: Changed

MarkY-LunarG:
Environment variables now start with "VK_".  FYI, all Khronos layers don't use VK_ before the environment variable name.  But I agree with the potential to avoid conflict.

Issue #745, opened 2019-03-01
vktrace should exit when not able to spawn process


davidlunarg:
vktrace prints an error message and hangs when it is not able to spawn the trace process:

$ vktrace -p nonExistantFileName 
vktrace error: Failed to spawn process
No such file or directory

The vktrace program seems to hang at this point. (Actually it goes into server mode, listening to the socket for a connection from any program using the trace layer, which might cause some confusion.) It should exit when it is not able to spawn the process.

davidlunarg:
Fixed in commit #752

PR #752, opened 2019-03-08
vktrace: handle error when exec of program fails

davidlunarg:
Kill all processes in the process group when the program
specified by the -p option cannot be executed. This prevents
the vktrace command from simply hanging when the program file
doesn't exist.

Change-Id: Iffc5cd7f864c5577b97e856d75dddc4a274077ba

### DIFF
--- a/vktrace/vktrace_common/vktrace_process.c
+++ b/vktrace/vktrace_common/vktrace_process.c
@@ -100,7 +100,11 @@ BOOL vktrace_process_spawn(vktrace_process_info* pInfo) {
         if (execv(fullExePath, args) < 0) {
             vktrace_LogError("Failed to spawn process.");
             perror(NULL);
-            exit(1);
+            // Exit by killing the whole process group.
+            // This kills both our process and our parent. We do this because
+            // we don't want to leave the parent hanging, waiting for data
+            // to arrive from the child.
+            kill(0, SIGINT);
         }
     }
 #endif


davidlunarg:
This pull request addresses issue #745.

vktrace/vktrace_common/vktrace_process.c
```diff
@@ -100,7 +100,11 @@ BOOL vktrace_process_spawn(vktrace_process_info* pInfo) {
         if (execv(fullExePath, args) < 0) {
             vktrace_LogError("Failed to spawn process.");
             perror(NULL);
-            exit(1);
+            // Exit by killing the whole process group.
+            // This kills both our process and our parent. We do this because
+            // we don't want to leave the parent hanging, waiting for data
+            // to arrive from the child.
+            kill(-getpgrp(), SIGINT);
```
mikew-lunarg: I'm not up on the details.  From the [kill man page](http://man7.org/linux/man-pages/man2/kill.2.html) it's not clear to me if pid should be negative or 0.  If 0 works, would that be clearer?

davidlunarg: 0 would also work. If the first arg is 0, kill will send the signal to all processes in the current process' group. If the first arg negative, kill will send the signal to all processes specified by the negative of the first arg. Either way would specify the same process group, and 0 would be cleaner. I'll change it.

mikew-lunarg: Right, that was my thinking.  Negative pid is best for killing some arbitrary group, whereas 0 is for killing your own group, which is the intent here.  Thanks

Issue #743, opened 2019-02-28
Replay of 32 bit trace file with 64 bit vkreplay gets seg fault on Windows


davidlunarg:

davidlunarg:
This issue has been fixed in above pull request.

PR #754, opened 2019-03-11
vkreplay: don't allow 32-bit trace replay on 64-bit replayer

davidlunarg:
Code was already in place to enforce this, but the recently
added code to pre-read portability packets crashes when
cross-playing, so the code to enforce it was not executed.
Moved the code to check for 32/64 to earlier in the replay.

Also added enforcing of not allowing replay if the Endianess of
the trace file and replay system do not match.

Change-Id: I018a806ac9cea68ba35c572e4a4177170cf4a73b


davidlunarg:
This pull request addresses issue #743

PR #758, opened 2019-03-13
vktrace: CreateImage and DestroyImage should use same allocator.

xooiamd:
This minor fix avoids some game title crash during trim.

Change-Id: I92903eb94a67b2c2fbe4906bb3e7603cb166db02


davidlunarg:
I'm trying to understand how setting the pAllocator arg of DestroyImage to NULL makes CreateImage and DestroyImage use the same allocator. Could you explain this?

xooiamd:
Just a few lines before this change at line #132  there is a createImage call with NULL pAllocator: 
VkResult result = mdd(device)->devTable.CreateImage(device, &imageCreateInfo, NULL, &image);

davidlunarg:
OK, I see where the image is created. Since pAllocator is no longer used, it should be removed. Otherwise, compile warnings may be generated. It should also be removed from the calling function.

The CI failure looks like a process problem not due to this code change.

PR #759, opened 2019-03-13
vkreplay: Some title hang during trim playback

xooiamd:
which is caused by infinite loop condition of
vkGetQueryPoolResults, the non-error return code
include VK_NOT_READY, so this change add VK_NOT_READY
to the loop condition to prevent the hang.
Also, add parentheses around the conditions for clear
logic.

Change-Id: Id311ab97c051a29f4da501d48c72a36cb135b164

### DIFF
--- a/scripts/vktrace_file_generator.py
+++ b/scripts/vktrace_file_generator.py
@@ -665,7 +665,7 @@ class VkTraceFileOutputGenerator(OutputGenerator):
         # Special cases for functions that use do-while loops
         do_while_dict = {'GetFenceStatus': 'replayResult != pPacket->result  && pPacket->result == VK_SUCCESS',
                          'GetEventStatus': '(pPacket->result == VK_EVENT_SET || pPacket->result == VK_EVENT_RESET) && replayResult != pPacket->result',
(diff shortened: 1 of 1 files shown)


davidlunarg:
After I merged this pull request, we discovered that our internal CI tests for VulkanTools started to fail. We narrowed it down to the commit for this pull request. The problem is that the replay of a trace of the Sasha Willems pipelinestatistics demo generates numerous errors.

I don't know if the failure is due to a problem in this commit, the program itself, the trace file, or a bug that has been in vktrace/replay all along but was uncovered by this commit. Until this failure can be addressed, I am reverting the commit for this pull request.

Here is a link to the trace file that shows the error. It was created on a Windows 10 system with nVidia graphics.

https://drive.google.com/file/d/1WvIFEqcFjm14os6yEejCDPCyWEc_CTcD/view?usp=sharing

xooiamd:
hi @davidlunarg , what is the error that is shown during the replay of the trace? is it VK_NOT_READY does not match VK_SUCCESS? 
Thanks,

davidlunarg:
The output is 1096 lines containing:

vkreplay error: Return value VK_NOT_READY from API call (vkGetQueryPoolResults) does not match return value from trace file VK_SUCCESS.

PR #763, opened 2019-03-18
Fixed comparison of threadID against wrong initial value. UINT32 -> UINT64

charles-lunarg:
Causes the output threadID to be UINT64_MAX in some cases.

Change-Id: I9ab7397c7f1423c9ff8826ad7ee4543b075cb160


davidlunarg:
@justuswang : Could you please review this change? You are the original author.

PR #767, opened 2019-03-20
vkreplay: fix an error in pNext handling

mizhen:
pNext of VkSamplerCreateInfo may point to a struct of extension other
than sampler YCBCR conversion, it cause playback hang problem. The change
fix the issue.

VKTRACE-97

Change-Id: I603548bbaddcf42877c34a8a98f7d2dec7920e65


mizhen:
The existing code put "sci = (VkSamplerYcbcrConversionInfo *)sci->pNext;" under the condition "if (sci->sType == VK_STRUCTURE_TYPE_SAMPLER_YCBCR_CONVERSION_INFO)", some title use other extension, so the playback will hang because sci will never get update.

davidlunarg:
> The existing code put "sci = (VkSamplerYcbcrConversionInfo *)sci->pNext;" under the condition "if (sci->sType == VK_STRUCTURE_TYPE_SAMPLER_YCBCR_CONVERSION_INFO)", some title use other extension, so the playback will hang because sci will never get update.

Got it. It was moved to after the closing brace.

PR #768, opened 2019-03-20
api_dump: Only output frames within a given range

charles-lunarg:
Range is given either by a list of comma separated values or a
frame start, count, and optional interval separated by dashes.
A count of 0 will keep outputting frames after the start frame
is encountered. Enables selective logging, useful to ignore
irrelevant data.

Modified:
- `layersvt/api_dump_layer.md`
- `layersvt/api_dump.h`
- `scripts/api_dump_generator.py`
- `vkconfig/layer_info.json`

Change-Id: Icd9f22644e6e880a600fea10e73ac89073729681

### DIFF
--- a/layersvt/api_dump.h
+++ b/layersvt/api_dump.h
@@ -18,6 +18,7 @@
  * Author: Lenny Komow <lenny@lunarg.com>
  * Author: Shannon McPherson <shannon@lunarg.com>
  * Author: David Pinedo <david@lunarg.com>
+ * Author: Charles Giessen <charles@lunarg.com>
  */
 
 #pragma once
@@ -46,6 +47,13 @@
 #include <unordered_set>
 #include <utility>
 
+#ifdef ANDROID
+
+#include <android/log.h>
+#include <sys/system_properties.h>
+
+#endif  // ANDROID
+
 #define MAX_STRING_LENGTH 1024
 
 // Defines for utilized environment variables.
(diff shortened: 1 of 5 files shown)


layersvt/api_dump.h
```diff
@@ -53,12 +61,22 @@
...
 #define API_DUMP_ENV_VAR_FLUSH_FILE "VK_APIDUMP_FLUSH"
+#define API_DUMP_ENV_VAR_OUTPUT_RANGE "VK_APIDUMP_RANGE"
 
 enum class ApiDumpFormat {
     Text,
     Html,
 };
 
+static const int FRAME_RANGE_UNLIMITED = 0;
+static const int FRAME_RANGE_INTERVAL_DEFAULT = 1;
```
MarkY-LunarG: Your constants are `const int`, but your range data below is `uint64_t`.  Some compilers will throw errors at the signed/unsigned and size discrepancy.  The above should likely be uint64_t.

layersvt/api_dump_layer.md
```diff
@@ -105,6 +105,8 @@ Detailed Output | `VK_APIDUMP_DETAILED` | `lunarg_api_dump.detailed` | true | Ge
 No Addresses/Handles | `VK_APIDUMP_NO_ADDR` | `lunarg_api_dump.no_addr` | false | Generate output without addresses or handles (which can vary run to run. Instead use the placeholder value "address".
 Flush After Every Command | `VK_APIDUMP_FLUSH` | `lunarg_api_dump.flush` | true | Flush after every API command's output
 Flush After Every Command | `VK_APIDUMP_OUTPUT_FORMAT` | `lunarg_api_dump.output_format` | `text` | Output the API Dump information as either a plain text file (`text`) or as an HTML-formated file (`html`).
+Selective Output Range | `VK_APIDUMP_RANGE` | `lunarg_api_dump.frame_range_format` | `false` | Only output data that belong to frames within the specified range. The range is given by a comma separated list of frames to output or a range of frames with a start, count, and optional interval separated by a dash. A count of 0 will output every frame after the start of the range.
+
```
MarkY-LunarG: You should give a string example. Like "For example, the string `5,15,3` will start outputting at frame 5, stop outputting at frame 15, and dump every 3 frames.".

Also, you should give an example for the single frames capability.

layersvt/api_dump.h
```diff
@@ -367,6 +432,102 @@ class ApiDumpSettings {
...
+    inline void printErrorMsg(const char *msg) {
+#ifdef ANDROID
+        __android_log_print(ANDROID_LOG_DEBUG, "api_dump", "%s", msg);
+#else
+        fprintf(stderr, "%s", msg);
+#endif
+    }
+
+    inline bool parseConditionalFrameRange(std::string range_str) {
+        int current_char = 0;
```
MarkY-LunarG: This should probably be `size_t` because some compilers may throw a fit on line 469, 480, etc because int and size_t could be different sizes.

layersvt/api_dump.h
cdgiessen: Good catch, fixed that.

layersvt/api_dump_layer.md
cdgiessen: Added examples for both, and in combination.

layersvt/api_dump.h
cdgiessen: Thanks, added.

layersvt/api_dump.h
charles-lunarg: Fixed

layersvt/api_dump_layer.md
charles-lunarg: Added examples for both and even includes the combination of them.

layersvt/api_dump.h
charles-lunarg: fixed

davidlunarg:
The frame range option should be added to layersvt/vk_layers_settings.txt

PR #779, opened 2019-03-29
Change devsim apiversion error msg to debug msg

mark-lunarg:
This is not an error for the the main devsim customer (LVTs) but the informational message prints "ERROR" in the LVT logs, adding some unnecessary confusion.

### DIFF
--- a/layersvt/device_simulation.cpp
+++ b/layersvt/device_simulation.cpp
@@ -1180,7 +1180,7 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateInstance(const VkInstanceCreateInfo *pCreat
     const VkApplicationInfo *app_info = pCreateInfo->pApplicationInfo;
     const uint32_t requested_version = (app_info && app_info->apiVersion) ? app_info->apiVersion : VK_API_VERSION_1_0;
     if (requested_version != VK_API_VERSION_1_0) {
-        ErrorPrintf("%s currently supports only VK_API_VERSION_1_0\n", kOurLayerName);
+        DebugPrintf("%s currently supports only VK_API_VERSION_1_0\n", kOurLayerName);
     }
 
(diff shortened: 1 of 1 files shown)


layersvt/device_simulation.cpp
```diff
@@ -1180,7 +1180,7 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateInstance(const VkInstanceCreateInfo *pCreat
     const VkApplicationInfo *app_info = pCreateInfo->pApplicationInfo;
     const uint32_t requested_version = (app_info && app_info->apiVersion) ? app_info->apiVersion : VK_API_VERSION_1_0;
     if (requested_version != VK_API_VERSION_1_0) {
-        ErrorPrintf("%s currently supports only VK_API_VERSION_1_0\n", kOurLayerName);
+        DebugPrintf("%s currently supports only VK_API_VERSION_1_0\n", kOurLayerName);
```
mikew-lunarg: Note that the difference between DebugPrintf() and ErrorPrintf(), other than the text, is the latter terminates the process:
https://github.com/LunarG/VulkanTools/blob/315d246c4d98fd4d40d1718d8eda353192b84c3a/layersvt/device_simulation.cpp#L413
Is that behavior expected by the upstream?

mark-lunarg: I thought that errorprintf only failed if you had the fail-on-error env/config thingy set.  So when this error shows up in an LVT, is devsim giving up and jumping out of the way or does it sit there and still do its thing?

mikew-lunarg: You are correct, the exit() only occurs if enabled with the envar.  Without the envar, it will continue like the debug version.

mark-lunarg: Thanks Mike!

PR #784, opened 2019-04-02
api_dump: Add CMake code for api_dump on MacOS

jeremyk-lunarg:
Added code, provided by Mike Schuchardt, to CMakelist.txt
files to provide build instructions for api_dump on MacOS.

### DIFF
--- a/layersvt/CMakeLists.txt
+++ b/layersvt/CMakeLists.txt
@@ -27,6 +27,15 @@ elseif(CMAKE_SYSTEM_NAME STREQUAL "Linux")
        # TODO Add Wayland Support
        # add_definitions(-DVK_USE_PLATFORM_WAYLAND_KHR)
     endif()
+elseif(CMAKE_SYSTEM_NAME STREQUAL "Darwin")
+    if (BUILD_MOLTENVK_SUPPORT)
+        add_definitions(-DVK_USE_PLATFORM_MACOS_MVK)
+    endif()
+
+    if (BUILD_METAL_SUPPORT)
+       # TODO Add Metal Support
+       add_definitions(-DVK_USE_PLATFORM_METAL_EXT)
+    endif()
 else()
     message(FATAL_ERROR "Unsupported Platform!")
 endif()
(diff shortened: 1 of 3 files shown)


layersvt/CMakeLists.txt
```diff
@@ -62,18 +77,31 @@ else()
     add_library(VkLayer_${target} SHARED ${ARGN})
     target_link_Libraries(VkLayer_${target} ${VkLayer_utils_LIBRARY})
     add_dependencies(VkLayer_${target} generate_api_cpp generate_api_h generate_api_html_h)
-    set_target_properties(VkLayer_${target} PROPERTIES LINK_FLAGS "-Wl,-Bsymbolic")
+    if (APPLE)
+        set_target_properties(VkLayer_${target} PROPERTIES LINK_FLAGS "-Wl")
+    else ()
```
mikes-lunarg: Did a little more reading on -Wl and it is meaningless without the option that follows

> -Wl,option
Pass option as an option to the linker. If option contains commas, it is split into multiple options at the commas. You can use this syntax to pass an argument to the option. For example, -Wl,-Map,output.map passes -Map output.map to the linker. When using the GNU linker, you can also get the same effect with -Wl,-Map=output.map.

So I think this can just be
```
    if (NOT APPLE)
        set_target_properties(VkLayer_${target} PROPERTIES LINK_FLAGS "-Wl,-Bsymbolic")
    else ()

File: layersvt/CMakeLists.txt [hunk 2]
```diff
@@ -62,18 +77,31 @@ else()
...
+        ${CMAKE_BINARY_DIR}
+        ${Vulkan-ValidationLayers_INCLUDE_DIR}
+    )
+else ()
+    include_directories(
+        ${CMAKE_CURRENT_SOURCE_DIR}
+        ${CMAKE_CURRENT_BINARY_DIR}
+        ${CMAKE_BINARY_DIR}
+        ${Vulkan-ValidationLayers_INCLUDE_DIR}
+        ${JSONCPP_INCLUDE_DIR}
```
mikes-lunarg: I couldn't figure out what the JSONCPP issue was, but I think we can revisit when/if devsim gets ported

scripts/known_good.json
```diff
@@ -23,10 +23,6 @@
       ],
       "cmake_options" : [
         "-DBUILD_TESTS=OFF"
-      ],
-      "build_platforms" : [
```
mikes-lunarg: Good catch! I totally missed this because I share dependencies between different projects and they were already built

jeremyk-lunarg:
Thanks for the input!
Pushed an update with those changes.

layersvt/CMakeLists.txt
mikew-lunarg: Doesn't VIA use JSONCPP?  I know somebody else does, as that's why I used it for DevSim.

mikes-lunarg: It appears so, then fixing the JSONCPP include issue would be a prerequisite for porting either via or devsim to mac.

PR #785, opened 2019-04-03
vktrace: remove vkWaitForFences sync for Windows platform

mizhen:
The sync operation in vkWaitForFences handling is only needed by Linux
platform, Windows side has similiar process in pageguard exception
handler and doesn't need the process. The change make related code
Linux only to avoid performance drop when capturing some title.

VKTRACE-97

Change-Id: I7b6ca6215917686c790dc515f912ba70b821aa9e


mizhen:
No.
The pageguardEnter/pageguardExit lock only protect page guard capture related structures. After removing the sync operation, there's no other structure access in the function for page guard capture.

mizhen:
The pageguardEnter/pageguardExit lock is only removed on Windows platform, no other data structures need the lock to protect on Windows.  The only part in the function that need protection by the lock is Ln 4584-4663 which is Linux only with the change.

PR #794, opened 2019-04-17
vktrace: fix black screen and random crash issues on game title under test

xooiamd:
Adding trim event state recreation and update the vkGetSwapchainImagesKHR handling to fix the game random crash issues. Also, fix the black screen issues for the same game title during trace and playback by updating the descriptors update templates handling.

### DIFF
--- a/vktrace/vktrace_layer/vktrace_lib_trim.cpp
+++ b/vktrace/vktrace_layer/vktrace_lib_trim.cpp
@@ -4463,8 +4463,24 @@ void recreate_fences(StateTracker &stateTracker) {
 }
 
 void recreate_events(StateTracker &stateTracker) {
+    VkResult result = VK_EVENT_RESET;
+    vktrace_trace_packet_header *pheader = nullptr;
     for (auto obj = stateTracker.createdEvents.begin(); obj != stateTracker.createdEvents.end(); obj++) {
         vktrace_write_trace_packet(obj->second.ObjectInfo.Event.pCreatePacket, vktrace_trace_get_trace_file());
+        auto eventobject = obj->second.vkObject;
+        result = mdd(reinterpret_cast<void *>(obj->second.belongsToDevice))
+                     ->devTable.GetEventStatus(obj->second.belongsToDevice, FromHandleId<VkEvent>(eventobject));
+        switch (result) {
+            case VK_EVENT_SET:
+                pheader = generate::vkSetEvent(false, obj->second.belongsToDevice, FromHandleId<VkEvent>(eventobject));
+                break;
+            default:
+                break;
+        }
+        if (pheader != nullptr) {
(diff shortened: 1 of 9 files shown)


vktrace/vktrace_replay/vkreplay_vkreplay.cpp
```diff
@@ -4316,12 +4349,18 @@ void vkReplay::manually_replay_vkUpdateDescriptorSetWithTemplate(packet_vkUpdate
...
     if (pPacket->descriptorUpdateTemplate != VK_NULL_HANDLE && remappedDescriptorUpdateTemplate == VK_NULL_HANDLE) {
         vktrace_LogError(
-            "Error detected in UpdateDescriptorSetWithTemplate() due to invalid remapped VkDescriptorUpdateTemplateKHR.");
+            "Error detected in UpdateDescriptorSetWithTemplate() due to invalid remapped VkDescriptorUpdateTemplate.");
         return;
     }
 
     // Map handles inside of pData
-    remapHandlesInDescriptorSetWithTemplateData(remappedDescriptorUpdateTemplate, (char *)pPacket->pData);
+    if (VK_NULL_HANDLE != remappedDescriptorUpdateTemplate) {
```
davidlunarg: Minor point: Could the check for a null handle be moved inside of remapHandlersInDescriptorSetWithTemplateData? The code would be easier to read. Avoiding the function call when it is null is probably not important, since the more common case is the handle is not null.

vktrace/vktrace_layer/vktrace_lib_trim.cpp
```diff
@@ -4463,8 +4463,27 @@ void recreate_fences(StateTracker &stateTracker) {
...
+    VkResult result = VK_EVENT_RESET;
+    vktrace_trace_packet_header *pheader = nullptr;
     for (auto obj = stateTracker.createdEvents.begin(); obj != stateTracker.createdEvents.end(); obj++) {
         vktrace_write_trace_packet(obj->second.ObjectInfo.Event.pCreatePacket, vktrace_trace_get_trace_file());
+        result = mdd(reinterpret_cast<void *>(obj->second.belongsToDevice))
+                     ->devTable.GetEventStatus(obj->second.belongsToDevice, reinterpret_cast<VkEvent>(obj->second.vkObject));
+        switch (result) {
+            case VK_EVENT_RESET:
+                pheader =
+                    generate::vkResetEvent(false, obj->second.belongsToDevice, reinterpret_cast<VkEvent>(obj->second.vkObject));
```
dustin-lunarg: When the event is created, its initial state will be unsignaled, so it shouldn't be necessary to call vkResetEvent after vkCreateEvent.

davidlunarg:
Changes look good.

This pull request has lots of commits, some if which do not pass CI. Could you squash the commits into one commit before I merge this change?

PR #805, opened 2019-04-24
screenshot: Use device to access correct dispatch table

davidlunarg:
Fixes problem with gfxrecon replay of a Dawn Of War 3 trace.

The dispatch table being used was being keyed off of the queue
handle passed in to vkQueuePresentKHR. It would work because
the queue handle was the same as the device handle for a previously
destroyed device.

Changed code to figure the device associated with the passed in
queue handle and to then use that device's displatch table.

Change-Id: Id5be18f67c48a785623c266ed884af742c378aeb

### DIFF
--- a/layersvt/screenshot.cpp
+++ b/layersvt/screenshot.cpp
@@ -159,8 +159,7 @@ typedef struct {
 } ImageMapStruct;
 static unordered_map<VkImage, ImageMapStruct *> imageMap;
 
-// unordered map: associates a device with a queue, commandPool, and physical
-// device also contains per device info including dispatch table
+// unordered map: associates a device with information about the device
 typedef struct {
     VkLayerDispatchTable *device_dispatch_table;
     bool wsi_enabled;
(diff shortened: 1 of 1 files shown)


layersvt/screenshot.cpp
```diff
@@ -1188,10 +1188,23 @@ VKAPI_ATTR VkResult VKAPI_CALL GetSwapchainImagesKHR(VkDevice device, VkSwapchai
 
 VKAPI_ATTR VkResult VKAPI_CALL QueuePresentKHR(VkQueue queue, const VkPresentInfoKHR *pPresentInfo) {
     static int frameNumber = 0;
-    if (frameNumber == 10) {
-        fflush(stdout); /* *((int*)0)=0; */
+    VkDevice device = NULL;
+    DeviceMapStruct *devMap;
+
+    // Find device associated with this queue
+    for (auto deviceMapIter = deviceMap.begin(); deviceMapIter != deviceMap.end(); deviceMapIter++) {
```
Tony-LunarG: deviceMap.find() ?

PR #806, opened 2019-04-25
vktrace: Fix trace/replay issues for some unity-based games

jyangsh:
1. Sometimes the memory is flushed with several ranges in one call, but
the pageguard data only need be copied once time.
2. Only the coherent memory need not be flushed in the special flush
memory call.


davidlunarg:
LGTM. @mizhen, could you review this too?

mizhen:
Ok, David.

Looks good, just seems no multiple PhysicalDevice handling.

PR #807, opened 2019-04-26
screenshot: Delete reused queue and cmdbuf keys from unordered maps

davidlunarg:
If a queue or cmdbuf handle is reused, the attempt to add an
entry to an unordered map using the reused queue or cmdbuf
handle as a key will fail. Added coded to remove the map entry
that is using the reused queue/cmdbuf key before adding the new
entry to the unordered map.

The following sequence caused this error:
   vkCreateDevice() returns device D1
   vkGetDeviceQueue(D1) returns queue Q1
   vkDestroyDevice(D1)
   vkCreateDevice() returns device D2
   vkGetDeviceQueue(D2) returns queue Q1
   vkQueuePresent(Q1)

The second call to vkGetDeviceQueue would not add Q1 to an unordered map
(leaving the previous entry in place), resulting in the vkQueuePresent
attempting to use the dispatch table from destroyed device D1.

An alternative fix would have been to change the behavior of vkDestroyDevice
to delete queue/cmdbuf entries that share the dispatch table with the device
that is being destroyed. Such a list wasn't readily available, so used the
above scheme instead.

### DIFF
--- a/layersvt/screenshot.cpp
+++ b/layersvt/screenshot.cpp
@@ -307,7 +307,11 @@ static void populate_frame_list(const char *vk_screenshot_frames) {
     } else {
         int parsingStatus = initScreenShotFrameRange(vk_screenshot_frames, &screenShotFrameRange);
         if (parsingStatus != 0) {
+#ifdef ANDROID
+            __android_log_print(ANDROID_LOG_ERROR, "screenshot", "range error\n");
+#else
             fprintf(stderr, "Screenshot range error\n");
+#endif
         }
     }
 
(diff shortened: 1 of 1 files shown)


layersvt/screenshot.cpp
```diff
@@ -707,6 +711,10 @@ static void writePPM(const char *filename, VkImage image1) {
     if (VK_SUCCESS != err) return;
 
     VkDevice cmdBuf = static_cast<VkDevice>(static_cast<void *>(data.commandBuffer));
+    if (deviceMap.find(cmdBuf) != deviceMap.end()) {
+        // Remove element with key cmdBuf from deviceMap so we can replace it
+        deviceMap.erase(cmdBuf);
```
Tony-LunarG: I would think that just calling erase() would be sufficient.   No need to find().

PR #809, opened 2019-04-27
android: interactive pauses, save logs

lunarpapillo:
Android tests are showing unreliable behavior when replaying
many vktrace traces; it seems the device doesn't always recognize
when an action has started.

I've added a couple of pauses as an experiment to try to get
the test to behave more like it were being driven by a human.
This seems to make at least some of the traces more reliable.

In addition, Android log files from the execution are now
saved for later examination in case of failure (or to ensure
that the device behaved as expected).

### DIFF
--- a/build-android/vktracereplay.sh
+++ b/build-android/vktracereplay.sh
@@ -283,6 +283,9 @@ fi
 # We want to halt on errors here
 set -e
 
+# Start up for logging
+adb $serialFlag logcat -c
+
 # Wake up the device
 adb $serialFlag shell input keyevent "KEYCODE_MENU"
 adb $serialFlag shell input keyevent "KEYCODE_HOME"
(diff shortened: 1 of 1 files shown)


lunarpapillo:
> It seems this would be a common problem when testing Android. Might there be a better solution than using sleep? How do we know that 1 second of sleep is sufficient?

All good questions.  I don't know that 1 second of sleep is sufficient, only that it seems to be better than 0 seconds, and IMHO would roughly mark how long it would take a user to trigger actions.

The only other example we have of Android adb operation doesn't use sleep, but also has additional steps that take additional time in between the operations.

I have a goal to unify all these scripts; that's in the middle-term future.  For now, the important part of this change is that the Android logs are saved for perusal, so hopefully I can examine the occasional failure more closely.

Thanks for reviewing...

PR #810, opened 2019-04-29
build: Update cmake_minimum_required(VERSION 3.11.3)

mikew-lunarg:
Also, cmake_minimum_required() must occur before the project() command,
see https://cmake.org/cmake/help/latest/command/project.html

Change-Id: I39a22c0722cdf33244b12d079853a895d6433f1b


mikew-lunarg:
Withdrawing.  Will be using 3.12.4 instead.

PR #815, opened 2019-05-14
api_dump: Fixed core enums not dumping extension elements

charles-lunarg:
Core spec enums didn't print elements that were added by an extension.

Change-Id: I4b2cb4f7eb0cc3f3805386cd23de38898967deaf

### DIFF
--- a/scripts/api_dump_generator.py
+++ b/scripts/api_dump_generator.py
@@ -1703,8 +1703,25 @@ class VulkanEnum:
             childValue = child.get('value')
             childBitpos = child.get('bitpos')
             childComment = child.get('comment')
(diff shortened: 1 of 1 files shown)


scripts/api_dump_generator.py
```diff
@@ -1661,6 +1661,7 @@ def __init__(self, rootNode, extensions):
 
             self.options.append(VulkanEnum.Option(childName, childValue, childBitpos, childComment))
 
+
```
jeremyk-lunarg: Was this extra line intentional?

charles-lunarg: Likely a fat finger on the return key.

PR #819, opened 2019-05-23
api_dump: Removed unecessary memory info

charles-lunarg:
The memoryType and memoryHeap members of VkPhysicalDeviceMemoryProperties will now
only dump valid elements, up to memoryTypeCount and memoryHeapCount values respectively.
Similar change for VkPhysicalDeviceGroupProperties's physicalDevices & physicalDeviceCount.
Before it would dump 16/32 entries, of which often only a handful were valid and useful.

Opted to add special case instead of modifying code-gen due to limited instances of this issue.

### DIFF
--- a/scripts/api_dump_generator.py
+++ b/scripts/api_dump_generator.py
@@ -540,7 +540,7 @@ inline std::ostream& dump_text_{pfnName}({pfnName} object, const ApiDumpSettings
 
 //========================== Struct Implementations =========================//
 
-@foreach struct where('{sctName}' not in ['VkShaderModuleCreateInfo', 'VkSurfaceFullScreenExclusiveWin32InfoEXT'])
+@foreach struct where('{sctName}' not in ['VkShaderModuleCreateInfo', 'VkSurfaceFullScreenExclusiveWin32InfoEXT', 'VkPhysicalDeviceMemoryProperties','VkPhysicalDeviceGroupProperties'])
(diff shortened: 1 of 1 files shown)


davidlunarg:
json output has been added to api dump, and merged to master. Can you update this pull request and also make the same changes to the json functions?

cdgiessen:
Addressed and amended to include specialization for json output.

scripts/api_dump_generator.py
```diff
@@ -1320,6 +1379,50 @@
...
+    needMemberComma = true;
+    settings.stream() << "\\n" << settings.indentation(indents) << "]";
+    return settings.stream();
+}}
+
+std::ostream& dump_json_VkPhysicalDeviceGroupProperties(const VkPhysicalDeviceGroupProperties& object, const ApiDumpSettings& settings, int indents)
+{{
+    settings.stream() << settings.indentation(indents) << "[\\n";
+
+    bool needMemberComma = false;
```
davidlunarg: Generated code needs to use the needMemberComma variable to keep track of where commas are to be inserted between members of structs. (And not before the first member and not after the last member). Since this function is not really generated code and we know where commas are needed, I think that the needMemberComma variable can be eliminated here. Doing so would make this code much simpler, straightforward, and easier to read.

This should also be done for the previous function dump_json_VkPhysicalDeviceMemoryProperties

cdgiessen:
Agreed, that does help readability. Fixed now.

charles-lunarg:
Yes, that was unnecessary code. 
If there aren't any other additions, I'll rebase it in.

PR #822, opened 2019-06-04
vkreplay: fix pNext exceptions during playback

xooiamd:
Game title under test throws exception in processing
the pNext struct in both vkGetImageMemoryRequirements2 and
vkGetBufferMemoryRequirements2 calls during playback.
Split the calls with their KHR extension calls and remove
process pNext struct for these calls as there is no valid
pointers to be processed for these calls. This fix the
game playback exception issue.

VKTRACE-120

Change-Id: I5590abc4f019621ff79c41cee34e04ff1c972330

### DIFF
--- a/vktrace/vktrace_replay/vkreplay_vkreplay.cpp
+++ b/vktrace/vktrace_replay/vkreplay_vkreplay.cpp
@@ -3005,6 +3005,31 @@ void vkReplay::manually_replay_vkGetImageMemoryRequirements(packet_vkGetImageMem
     return;
 }
 
+void vkReplay::manually_replay_vkGetImageMemoryRequirements2(packet_vkGetImageMemoryRequirements2 *pPacket) {
+    VkDevice remappeddevice = m_objMapper.remap_devices(pPacket->device);
+    VkImage remappedimage = VK_NULL_HANDLE;
+
+    if ((pPacket->device != VK_NULL_HANDLE) && (remappeddevice == VK_NULL_HANDLE)) {
+        vktrace_LogError("Error detected in GetImageMemoryRequirements2KHR() due to invalid remapped VkDevice.");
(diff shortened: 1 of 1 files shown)


davidlunarg:
I don't understand why the pNext structures should not be processed. Is it because the pNext pointers or the structures they point to are invalid? If so, it seems that should be fixed rather than not processing them. In the future, an extension may be added that uses the pNext structures, and it would be good if the code to process them was already in place.

xooiamd:
hi @davidlunarg , from the doc, 
https://www.khronos.org/registry/vulkan/specs/1.1-extensions/man/html/vkGetBufferMemoryRequirements2.html
https://www.khronos.org/registry/vulkan/specs/1.1-extensions/man/html/VkMemoryRequirements2.html
https://www.khronos.org/registry/vulkan/specs/1.1-extensions/man/html/VkMemoryDedicatedRequirements.html

for vkGetBufferMemoryRequirements2 or vkGetImageMemoryRequirements2 , the pNext chain of the VkMemoryRequirements2 structure which is in the parameter of  pMemoryRequirements may have a  VkMemoryDedicatedRequirements structure but this structure contain no pointers to be interpreted/ processed during the playback.

davidlunarg:
I have looked at this some more, and still believe the calls to process the pNext structures are needed. Even though the pNext structures passed in by current apps do not have pointers that need to be interpreted, future apps may have them. The vkMemoryDedicatedRequirements structure may not have pointers to be interpreted, but other pNext structures in the future might.

xooiamd:
hi @davidlunarg , sorry I remembered why i removed these pnext, because they are already beiing handled by the interpret_body_as_ function in the vktrace_vk_vk_packets.h function due to recent fix to the generation script

PR #824, opened 2019-06-05
scripts: Add multiple-require support to api_dump

shannon-lunarg:
Previously, `api_dump_generator.py` did not support commands with
multiple require blocks.

This change is required to update the Validation Layers known good for the 108 headers.


shannon-lunarg:
@mikes-lunarg Good call!

All changes passed internal CI.

PR #825, opened 2019-06-05
apidump: Implement json output

davidlunarg:


dustin-lunarg:
I'm unable to parse the JSON output with Python, due to the following syntax:
```
[
"Frame 0" :
[
    "Thread 0" :
    {
        "vkCreateInstance(pCreateInfo, pAllocator, pInstance)" : 
        {
            "rval" : "VK_SUCCESS",
            "args" :
            {
```

The first two '[' characters are creating lists, but the elements in the lists use the dictionary key/value syntax.  I think it makes sense for this to be a dictionary keyed by frame ID, with a value that is a list of API calls performed by the frame, which would require the following changes:
1. Change the outermost '[' to '{'.
2. Move the 'Thread 0' from a dictionary key to an entry of the API call dictionary.  In its current form, this is defining a dictionary where all elements have the same 'Thread 0' key.

The output would look something like this:
```
{
"Frame 0" :
[
    "vkCreateInstance(pCreateInfo, pAllocator, pInstance)" : 
    {
        "thread" : "Thread 0" ,
        "rval" : "VK_SUCCESS",
        "args" :
        {
```

mikew-lunarg:
Is there any chance of retrofitting a jsonschema to this output format before (or soon after) it goes live?
That will ensure the format conforms to the defined format, and give consumers a specified target to receive the data.

dustin-lunarg:
This [gist](https://gist.github.com/dustin-lunarg/df4c899d3918357d042cd9a8616732f9) contains a proposed JSON output, along with a sample Python script to process it, which I think would work.  It can be used as a starting point for creating a JSON schema.

The syntax is roughly:
```
data = "[" frame_data { frame_data } "]" ;
frame_data = "[" call_data { call_data } "]" ;
call_data =
"{"
    "name" ":" string ","
    "thread" ":" string ","
    "returnType" ":" string ","
    [ "returnValue" ":" string "," ]    # Only present when return type is not 'void'
    "args" ":" "[" arg_data { arg_data } "]"
"}" ;
arg_data =
"{"
    "type" ":" string ","
    "name" ":" string ","
    [ "address" ":" string "," ]    # Only present for pointers
    ( "value" ":" string |    # For a value/pointer
      "elements" ":" "[" arg_data { arg_data } "]" |    # For an array (name for each element is the "[i]" pattern)
      "members" ":" "[" arg_data { arg_data } "]" )    # For a struct
"}" ;
```

charles-lunarg:
vkconfig/layer_info.json isn't updated to have Json as a output type.

```
"output_format": {
                "name": "Output Format",
                "description": "Specifies the format used for output; can be Text (default -- outputs plain text) or Html",
                "type": "enum",
                "options": [ "Text", "Html" ],
                "default": "Text"
},
```
to
```
"output_format": {
                "name": "Output Format",
                "description": "Specifies the format used for output; can be Text (default -- outputs plain text), Html, or Json",
                "type": "enum",
                "options": [ "Text", "Html", "Json"],
                "default": "Text"
},
```

mikew-lunarg:
FWIW certainly, I'm not up on the requirements for this json interface.

I'd recommend adding some form of format version declaration, so applications can easily detect how the format evolves over time.  (In json-schema, that versioning is accomplished using a "$schema" element)  I'd also recommend defining some sort of comment element, since json does not support commenting (It's really handy to make data inert without being forced to delete it from the document)

This is an all-values-are-strings design.  [json-schema](https://json-schema.org/) provides json-native types, from which producer/consumer code can be generated, and most importantly, can validate json documents against the schema (useful for detecting the error above) independently of any application code.

FYI the json-schema defined for the Device Simulation layer directly implements Vulkan datastructures (when possible; there are some C-isms that dont map directly to json types)  See https://schema.khronos.org/vulkan/devsim_1_0_0.json

davidlunarg:
The last commit (29c059fa31d3f0cf03db5cca2220c228596d962d) changes the way arrays and structures are output. I modified sample_json.py, available here (https://gist.github.com/davidlunarg/45e864658104c79a0cff8324e9886158). This version of the code produces output that can be correctly processed by the new version of the python script.

davidlunarg:
Merging this, but will create an issue that a json schema should be created.

PR #832, opened 2019-06-17
via: allow vkvia to find SDK in 1.1.108

bradgrantham-lunarg:
add loop over SDK explicit layer path dirs in PrintSystemSdkInfo

add 1.1.108 explicit layer path dir so vkvia succeeds on 1.1.108

### DIFF
--- a/via/via_system_linux.cpp
+++ b/via/via_system_linux.cpp
@@ -1139,8 +1139,6 @@ ViaSystem::ViaResults ViaSystemLinux::PrintSystemSdkInfo() {
         PrintTableElement("");
         PrintEndTableRow();
 
-	// Because of the ordering here, SDK >= 1.1.108 will be detected
-	// after SDK <= 1.1.106, thus a newer SDK will take precedence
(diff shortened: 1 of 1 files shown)


via/via_system_linux.cpp
```diff
@@ -1139,23 +1139,28 @@ ViaSystem::ViaResults ViaSystemLinux::PrintSystemSdkInfo() {
...
+		    if (NULL != strstr(cur_ent->d_name, vulkan_so_prefix) && strlen(cur_ent->d_name) == 14) {
+		    }
+		}
+		closedir(sdk_dir);
+
+		result = PrintExplicitLayersInFolder("", explicit_layer_path);
+
+		_found_sdk = true;
+		_sdk_path = sdk_path;
+		sdk_exists = true;
```
MarkY-LunarG: Should we break here?

bradgrantham-lunarg: We could.  Then it would preferentially find earlier SDKs.

bradgrantham-lunarg: Sure!  I guess we won't have both explicit layer paths in one VULKAN_SDK.

bradgrantham-lunarg:
Abandoned

PR #836, opened 2019-06-25
asst_layer: API Version check

camden-lunarg:
If the API version of the device is lower than the API version specified in CreatInstance, print out a warning.

Change-Id: I01c94a69ca777a8c7f08ab1495d9a95f73c9c186

### DIFF
--- a/layer_factory/assistant_layer/api_version_warning.h
+++ b/layer_factory/assistant_layer/api_version_warning.h
@@ -21,16 +21,31 @@
 
 #include <sstream>
 #include <iostream>
+#include <iomanip>
 
 class ApiVersionWarning : public layer_factory {
    public:
     // Constructor for interceptor
     ApiVersionWarning(){};
 
+    // get the API name is proper format
+    std::string GetAPIVersionName(uint32_t version) {
+        std::stringstream version_name;
+        uint32_t major = VK_VERSION_MAJOR(version);
+        uint32_t minor = VK_VERSION_MINOR(version);
+        uint32_t patch = VK_VERSION_PATCH(version);
+
+        version_name << major << "." << minor << "." << patch << " (0x" << std::setfill('0') << std::setw(8) << std::hex << version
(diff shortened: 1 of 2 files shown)


layer_factory/assistant_layer/api_version_warning.h
```diff
@@ -0,0 +1,60 @@
...
+        // get API version of physical device passed when creating device.
+        instance_layer_data* instance_data = GetLayerDataPtr(get_dispatch_key(physicalDevice), instance_layer_data_map);
+        VkPhysicalDeviceProperties physical_device_properties{};
+        instance_data->dispatch_table.GetPhysicalDeviceProperties(physicalDevice, &physical_device_properties);
+
+        // check api versions and warn if instance api Version is higher than version on device.
+        // NOTE: Should be able to use ">" operator as other places in validation layers does so
+        if (instance_api_version > physical_device_properties.apiVersion) {
+            std::stringstream message;
+            message << "vkCreateDevice: API Version of current instance is higher than API Version on device.";
```
bradgrantham-lunarg: Is it worth printing the device's API version here?

camden-lunarg: Yea I would say so. What form of the api_version would you suggest I print out? Should I do the major and minor conversion, or should I just print out the actual value of VK_API_VERSION_1_0 etc?

bradgrantham-lunarg:
If it doesn't suck too bad to implement, I think both, maybe in major,
minor, and patch form as "XX.YY.ZZ" and then in raw hex number form as
"0xABCD0123", maybe like "...version of current instance, XX.YY.ZZ
(0xABCD0123), is higher than API version on device, XX.YY.ZZ
(0xABCD0123)."  But maybe check with Mark Lobodzinski to make sure he
agrees; it's starting to look pretty verbose.

camden-lunarg:
Pushed to master

PR #839, opened 2019-07-01
docs: Improve instructions for using update_deps.

davidlunarg:
Same change as pull request
https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/955

Change-Id: Ia653f165ecce5e2c4cd1a1d145e63c65036a2377

### DIFF
--- a/update_external_sources.bat
+++ b/update_external_sources.bat
@@ -21,3 +21,4 @@ git submodule update --init --recursive
       echo JsonCPP Amalgamation failed to generate %JSONCPP_DIR%\dist\json\json.h
       set errorCode=1
    )
+   cd %BUILD_DIR%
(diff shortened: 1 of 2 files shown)


update_external_sources.bat
```diff
@@ -21,3 +21,4 @@ git submodule update --init --recursive
       echo JsonCPP Amalgamation failed to generate %JSONCPP_DIR%\dist\json\json.h
       set errorCode=1
    )
+   cd %BUILD_DIR%
```
lunarpapillo: Just curious: does this actually do something in a .bat file?  I.e. a .bat file runs in the same environment as its caller, and can change said environment?

davidlunarg: Yes, a .bat file runs in the same environment as the caller. Without this final cd command, the script will exit and the caller will have a new cwd. This would have necessitated adding a cd command back up a few levels in the set of instructions in the BUILD.md file.

PR #843, opened 2019-07-05
vktrace: Fix an issue of PMB host memory capture

mizhen:
Mapped shadow memory start from a system page starting address is a basic
assumption in PMB capture code, but for PMB host memory capture, it's not
true any more, it cause the changed page status doesn't match the changed
page index and cause wrong write/read in trace file packet and sometimes
crash in capture. The change fix the problem.

VKTRACE-135

Change-Id: I3353632160f064efd06c498f615562e3ef1b22a3


davidlunarg:
Changes look good. Please address the travis CI build errors.

PR #847, opened 2019-07-10
apidump: No address mode adds unique handles

charles-lunarg:
When no address mode is selected, apidump will now create
new api_dump specific unique values per handle. This
can help with reproducing the same dump file by removing
external factors.

files changed:

	scripts/api_dump_generator.py
        layersvt/api_dump.h

Change-Id: I7b8e82f498357ad4782227f6fca83d84eabbd370


charles-lunarg:
Current solution doesn't account for vulkan re-using handles. Will need to either require unique objects or recreate the same functionality.

PR #850, opened 2019-07-16
Memory Freed While Still Bound Warning

camden-lunarg:
Resolves issue tracked in KhronosGroup/Vulkan-ValidationLayers#24 which states "Warn if memory freed while still bound to other objects. See note in core_validation.cpp CoreChecks::PreCallRecordFreeMemory related to removing the INFORMATION_BIT message".

This specific issue is located at KhronosGroup/Vulkan-ValidationLayers#872

### DIFF
--- a/layer_factory/assistant_layer/mem_bound_warning.cpp
+++ b/layer_factory/assistant_layer/mem_bound_warning.cpp
@@ -1,17 +1,25 @@
+/*
+ * Copyright (c) 2015-2018 Valve Corporation
+ * Copyright (c) 2015-2018 LunarG, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Author: Camden Stocker <camden@lunarg.com>
+ */
+
 #include "mem_bound_warning.h"
 #include <iostream>
-
-VkResult MemBoundWarning::PostCallAllocateMemory(VkDevice device, const VkMemoryAllocateInfo *pAllocateInfo,
-                                                 const VkAllocationCallbacks *pAllocator, VkDeviceMemory *pMemory,
-                                                 VkResult result) {
-    if (result == VK_SUCCESS) {
-        // create new memory entry into the map but give it no values
-        std::unordered_set<uint64_t> set;
-        memory_handle_map.insert(std::make_pair(*pMemory, set));
-    }
-
-    return VK_SUCCESS;
-}
+#include <assert.h>
 
 // helper function that update all the dictionaries for both image and buffer handles
 template <typename T>
@@ -19,17 +27,10 @@ void MemBoundWarning::SetBindMemoryState(T handle, VkDeviceMemory memory) {
     uint64_t handle_64 = HandleToUint64(handle);
 
     if (memory != VK_NULL_HANDLE) {
-        // If the entry does not exist yet, then create it
-        if (handle_memory_map.count(handle_64) <= 0) {
-            std::unordered_set<VkDeviceMemory> set;
-            handle_memory_map.insert(std::make_pair(handle_64, set));
-        }
+        auto &map_pair = handle_memory_map[handle_64];
+        map_pair.insert(memory);
 
-        // insert the memory and handle into their respective maps
-        auto &memory_set = handle_memory_map.find(handle_64)->second;
-        memory_set.insert(memory);
-
-        auto &handle_set = memory_handle_map.find(memory)->second;
+        auto &handle_set = memory_handle_map[memory];
         handle_set.insert(handle_64);
     }
 }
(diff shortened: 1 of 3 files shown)


layer_factory/assistant_layer/mem_bound_warning.cpp
```diff
@@ -0,0 +1,210 @@
...
+
+// intercept for the sparse memory to track memory state
+VkResult MemBoundWarning::PostCallQueueBindSparse(VkQueue queue, uint32_t bindInfoCount, const VkBindSparseInfo *pBindInfo,
+                                                  VkFence fence, VkResult result) {
+    for (uint32_t bindIdx = 0; bindIdx < bindInfoCount; bindIdx++) {
+        const VkBindSparseInfo &bindInfo = pBindInfo[bindIdx];
+
+        for (uint32_t j = 0; j < bindInfo.bufferBindCount; j++) {
+            for (uint32_t k = 0; k < bindInfo.pBufferBinds[j].bindCount; k++) {
+                auto sparse_binding = bindInfo.pBufferBinds[j].pBinds[k];
```
dustin-lunarg: It looks like this is making a copy. Did you intend to use a reference instead? Same for the two cases immediately below.

layer_factory/assistant_layer/mem_bound_warning.cpp
```diff
@@ -0,0 +1,210 @@
...
+        }
+    }
+
+    return VK_SUCCESS;
+}
+
+VkResult MemBoundWarning::PostCallGetSwapchainImagesKHR(VkDevice device, VkSwapchainKHR swapchain, uint32_t *pSwapchainImageCount,
+                                                        VkImage *pSwapchainImages, VkResult result) {
+    assert(swapchain);
+    auto images = swapchain_image_map[swapchain];
```
dustin-lunarg: It looks like this is making a copy. Did you intend to use a reference instead?

layer_factory/assistant_layer/mem_bound_warning.cpp
```diff
@@ -0,0 +1,210 @@
...
+}
+
+void MemBoundWarning::PostCallDestroySwapchainKHR(VkDevice device, VkSwapchainKHR swapchain,
+                                                  const VkAllocationCallbacks *pAllocator) {
+    assert(swapchain);
+
+    auto map_pair = swapchain_image_map.find(swapchain);
+    if (map_pair == swapchain_image_map.end())
+        return;  // this will happend if handle was never Binded to memory...prevents segfault
+    auto images = map_pair->second;
```
dustin-lunarg: It looks like this is making a copy. Did you intend to use a reference instead?

layer_factory/assistant_layer/mem_bound_warning.cpp
```diff
@@ -0,0 +1,210 @@
...
+    for (const auto &handle : handle_set) {
+        char handle_string[64];
+        char memory_string[64];
+        uint64_t handle_64 = HandleToUint64(handle);
+        uint64_t memory_64 = HandleToUint64(memory);
+        sprintf(handle_string, "0x%" PRIxLEAST64, handle_64);
+        sprintf(memory_string, "0x%" PRIxLEAST64, memory_64);
+
+        std::stringstream message;
+        message << "Vk Object " << handle_string << " still has a reference to Memory Object " << memory_string;
```
dustin-lunarg: You should be able to avoid the sprintf and operate on the uint64_t values directly with something like:
`message << "Vk Object 0x" << std::hex << handle_64 << " still has a reference to Memory Object 0x" << memory_64;`

layer_factory/assistant_layer/mem_bound_warning.cpp
camden-lunarg: I agree. Fixed in upcoming commit

layer_factory/assistant_layer/mem_bound_warning.cpp
camden-lunarg: Fixed in upcoming commit

layer_factory/assistant_layer/mem_bound_warning.cpp
camden-lunarg: In this case, since the images are only read, and not written to. A reference may not be necessary. However, making it a reference improves memory efficiency right? I will also make it a constant since it is read only. 
Thoughts?

layer_factory/assistant_layer/mem_bound_warning.cpp
camden-lunarg: As per our discussion, we decided this should be left the way it is to stay consistent with validation_layers and for a few other consistency reasons.

layer_factory/assistant_layer/mem_bound_warning.cpp
dustin-lunarg: Yes, I would make this a const reference to avoid the cost of copying a std::vector.

layer_factory/assistant_layer/mem_bound_warning.cpp
```diff
@@ -0,0 +1,209 @@
...
+    return VK_SUCCESS;
+}
+
+// Get memory associated with the handle being destroyed then find handle set using this memory and erase it
+template <typename T>
+void MemBoundWarning::RemoveBindedMemory(T handle) {
+    assert(handle);
+    uint64_t handle_64 = HandleToUint64(handle);
+    auto map_pair = handle_memory_map.find(handle_64);
+    if (map_pair == handle_memory_map.end()) return;  // this will happend if handle was never Binded to memory...prevents segfault
```
dustin-lunarg: Spelling 'happend'

layer_factory/assistant_layer/mem_bound_warning.cpp
```diff
@@ -0,0 +1,209 @@
...
+    const auto &memory_set = map_pair->second;
+
+    // remove all memory in the memory set. If it is not sparse memory, there will only be one item in set
+    for (VkDeviceMemory memory : memory_set) {
+        auto map_pair = memory_handle_map.find(memory);
+        if (map_pair == memory_handle_map.end())
+            continue;  // this will happend if handle was never Binded to memory...prevents segfault
```
dustin-lunarg: Spelling 'happend'

layer_factory/assistant_layer/mem_bound_warning.cpp
```diff
@@ -0,0 +1,209 @@
...
+    }
+}
+
+void MemBoundWarning::PostCallDestroySwapchainKHR(VkDevice device, VkSwapchainKHR swapchain,
+                                                  const VkAllocationCallbacks *pAllocator) {
+    assert(swapchain);
+
+    auto map_pair = swapchain_image_map.find(swapchain);
+    if (map_pair == swapchain_image_map.end())
+        return;  // this will happend if handle was never Binded to memory...prevents segfault
```
dustin-lunarg: Spelling 'happend'

PR #851, opened 2019-07-17
via: Port VIA to MacOS

jeremyk-lunarg:


jeremyk-lunarg:
Ping.
I am specifically looking to see if this change plays well with the rest of the program's style.
Also if there are suggestions for any other information VIA on MacOS should display.

via/via_system_macos.cpp
```diff
@@ -0,0 +1,1435 @@
...
+        settings_files.push_back(full_file);
+        full_file = "/usr/share/vulkan/settings.d/";
+        full_file += settings_file_name;
+        settings_files.push_back(full_file);
+        full_file = "/usr/local/etc/vulkan/settings.d/";
+        full_file += settings_file_name;
+        settings_files.push_back(full_file);
+        full_file = "/usr/local/share/vulkan/settings.d/";
+        full_file += settings_file_name;
+        settings_files.push_back(full_file);
```
mikes-lunarg: same ~/.local comment as in `ViaSystemMacOS::PrintSystemImplicitLayerInfo`

via/via_system_macos.cpp
```diff
@@ -0,0 +1,1435 @@
...
+#include <sys/utsname.h>
+#include <dirent.h>
+#include <unistd.h>
+#include <dlfcn.h>
+
+#include "via_system_macos.hpp"
+
+ViaSystemMacOS::ViaSystemMacOS() : ViaSystem() {
+    char temp_c_string[1024];
+    ssize_t len = ::readlink("/proc/self/exe", temp_c_string, 1023);
```
bradgrantham-lunarg: I think you may want to use _NSGetExecutablePath() here to read the current process' path; I don't think /proc is available on MacOS.

via/via_system_macos.cpp
```diff
@@ -0,0 +1,1431 @@
...
+                if (FindMacOSSystemObject(this, driver_name, location, CheckDriver, true)) {
+                    found_lib = true;
+                    could_load = VerifyOpen(location, load_error);
+                }
+            }
+        }
+        if (!found_lib) {
+            uint32_t image_count = _dyld_image_count();
+            if (image_count == 0) {
+                    snprintf(generic_string, 1023,
```
bradgrantham-lunarg: I don't think you need this check - if you just have the for-loop on image_count and then the if(!found) afterwards, it will catch the case where image_count is 0 and then there's only one copy of this HTML output.  Do you have time to make this edit?

jeremyk-lunarg:
Updated

via/via_system_macos.cpp
```diff
@@ -0,0 +1,1410 @@
...
+                if (FindMacOSSystemObject(this, driver_name, location, CheckDriver, true)) {
+                    found_lib = true;
+                    could_load = VerifyOpen(location, load_error);
+                }
+            }
+        }
+        if (!found_lib) {
+            bool found = false;
+            uint32_t image_count = _dyld_image_count();
+            if (image_count > 0) {
```
bradgrantham-lunarg: I don't think you even need this check.  If image_count <= 0, the for loop will never be entered, and you'll just drop through to line 511 with found = false, which I think is just fine.  However, weighing the cost of one more CI submission, I'm okay with leaving this as it is.

via/via_system_macos.cpp
```diff
@@ -0,0 +1,1410 @@
...
+    std::string sdk_env_name;
+    const char vulkan_so_prefix[] = "libvulkan.dylib.";
+    DIR *sdk_dir;
+    dirent *cur_ent;
+    char *env_value;
+
+    PrintBeginTable("LunarG Vulkan SDKs", 4);
+
+    // First, try environmental variables
+    for (uint32_t dir = 0; dir < 2; dir++) {
```
mikes-lunarg: You don't need this loop anymore since there is only one var to check.  As currently written, there is a mismatch between possible values of `dir` (0 and 1) and case statements (0 only).  VIA terminates early when run without the `VULKAN_SDK` env var set because `dir` gets to `1` and ends up in the `default:` error case

PR #852, opened 2019-07-17
vkreplay: Use CMAKE_BUILD_RPATH instead of setting linker flags

johnzupin:
The RPATH will not be stripped if `make install` is ran when setting RPATH with linker flags. 

Using CMAKE_BUILD_RPATH to set RPATH to $ORIGIN will have RPATH set to $ORIGIN in the build directory, and the RPATH will be stripped if `make install` is ran.


johnzupin:
For context there was an old discussion [here](https://github.com/LunarG/Vulkan-SDK-Packaging /issues/37) about this.

PR #853, opened 2019-07-17
devsim: Fixed new warning from gcc-8

charles-lunarg:
Using `(sizeof(array) / sizeof(array[0])) to get the size of an array on a pointer
    creates a new warning with gcc-8. Warning is -Wsizeof-pointer-div.

    changes:
            layersvt/device_simulation.cpp

Change-Id: I29b668c3f3e2e60739710b0ba9180933763866ab

### DIFF
--- a/layersvt/device_simulation.cpp
+++ b/layersvt/device_simulation.cpp
@@ -83,11 +83,11 @@ const uint32_t kLayerPropertiesCount = (sizeof(kLayerProperties) / sizeof(kLayer
 const char *kOurLayerName = kLayerProperties[0].layerName;
 
 // Instance extensions that this layer provides:
-const std::array<VkExtensionProperties,16> kInstanceExtensionProperties = {};
+const std::array<VkExtensionProperties, 0> kInstanceExtensionProperties = {};
 const uint32_t kInstanceExtensionPropertiesCount = kInstanceExtensionProperties.size();
 
 // Device extensions that this layer provides:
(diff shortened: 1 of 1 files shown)


layersvt/device_simulation.cpp
```diff
@@ -82,12 +83,12 @@ const uint32_t kLayerPropertiesCount = (sizeof(kLayerProperties) / sizeof(kLayer
 const char *kOurLayerName = kLayerProperties[0].layerName;
 
 // Instance extensions that this layer provides:
-const VkExtensionProperties *kInstanceExtensionProperties = {};
-const uint32_t kInstanceExtensionPropertiesCount = (sizeof(kInstanceExtensionProperties) / sizeof(kInstanceExtensionProperties[0]));
+const std::array<VkExtensionProperties,16> kInstanceExtensionProperties = {};
```
mikew-lunarg: Concerned about magic numbers (16 here and 15 below) defeating the intent of compile-time auto-sizing.

charles-lunarg: std::vector is likely the best alternative even if it has a heap allocation.
std::array could have the size explicitly set to 0 until someone added an element to it, which would be annoying but then there is no wasted space.

charles-lunarg: An array size of 0 would be an appropriate value, considering it is currently empty.

Issue #800, opened 2018-12-06
Add the ability to print timestamps to api_dump layer


KarenGhavam-lunarG:
It would be useful to see where things slow down / hang

charles-lunarg:
Added this functionality in PR #854 but is off by default, to preserve the current behavior. The timestamp is the time in microseconds since apidump started. 

While it wouldn't be hard to dump the duration of a vulkan call with apidump, I don't think it is a very appropriate place to do so. A proper profiler should be used instead, as it could better diagnose time spent inside the driver versus validation layers versus writing data out, which is most of what apidumps time is spent doing.

PR #854, opened 2019-07-18
apidump: Timestamps in dump output

charles-lunarg:
apidump will now show the time a function was called since start
in microseconds. This is useful to determine which functions are
taking significant time to execute.

Added the VK_APIDUMP_TIMESTAMP environmental variable and the
lunarg_api_dump.show_timestamp setting to the api_dump settings file

changed files:
	layersvt/api_dump.h
	layersvt/api_dump_layer.md
	layersvt/vk_layer_settings.txt
	scripts/api_dump_generator.py
	vkconfig/layer_info.json

Change-Id: Ie223747d947c803617018034b7c5f66d3a243d9b

### DIFF
--- a/layersvt/api_dump_layer.md
+++ b/layersvt/api_dump_layer.md
@@ -106,7 +106,7 @@ No Addresses/Handles | `VK_APIDUMP_NO_ADDR` | `lunarg_api_dump.no_addr` | false 
 Flush After Every Command | `VK_APIDUMP_FLUSH` | `lunarg_api_dump.flush` | true | Flush after every API command's output
 Output format | `VK_APIDUMP_OUTPUT_FORMAT` | `lunarg_api_dump.output_format` | `text` | Output the API Dump information as a text file (`text`), an HTML-formated file (`html`), or a json file (`json`).
 Selective Output Range | `VK_APIDUMP_OUTPUT_RANGE` | `lunarg_api_dump.output_range` | `0-0` | Only output frames within the specified range. Given by a comma separated list of frames or a range with a start, count, and optional interval separated by dashes. A count of 0 will output every frame after the start of the range. Example: "5-8-2" will output frame 5, continue until frame 13, dumping every other frame. Example: "3,8-2" will output frames 3, 8, and 9.
- 
+Show Timestamps | `VK_APIDUMP_TIMESTAMP` | `lunarg_api_dump.show_timestamp` | true | Show the timestamp of function calls since start in microseconds
 
 ### Settings Priority
 
--- a/layersvt/api_dump_layer.md
+++ b/layersvt/api_dump_layer.md
@@ -106,7 +106,7 @@ No Addresses/Handles | `VK_APIDUMP_NO_ADDR` | `lunarg_api_dump.no_addr` | false 
 Flush After Every Command | `VK_APIDUMP_FLUSH` | `lunarg_api_dump.flush` | true | Flush after every API command's output
 Output format | `VK_APIDUMP_OUTPUT_FORMAT` | `lunarg_api_dump.output_format` | `text` | Output the API Dump information as a text file (`text`), an HTML-formated file (`html`), or a json file (`json`).
 Selective Output Range | `VK_APIDUMP_OUTPUT_RANGE` | `lunarg_api_dump.output_range` | `0-0` | Only output frames within the specified range. Given by a comma separated list of frames or a range with a start, count, and optional interval separated by dashes. A count of 0 will output every frame after the start of the range. Example: "5-8-2" will output frame 5, continue until frame 13, dumping every other frame. Example: "3,8-2" will output frames 3, 8, and 9.
-Show Timestamps | `VK_APIDUMP_TIMESTAMP` | `lunarg_api_dump.show_timestamp` | true | Show the timestamp of function calls since start in microseconds
+Show Timestamps | `VK_APIDUMP_TIMESTAMP` | `lunarg_api_dump.show_timestamp` | false | Show the timestamp of function calls since start in microseconds
 
 ### Settings Priority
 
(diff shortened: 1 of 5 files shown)


scripts/api_dump_generator.py
```diff
@@ -698,7 +698,11 @@
 std::ostream& dump_text_{funcName}(ApiDumpInstance& dump_inst, {funcReturn} result, {funcTypedParams})
 {{
     const ApiDumpSettings& settings(dump_inst.settings());
-    settings.stream() << "Thread " << dump_inst.threadID() << ", Frame " << dump_inst.frameCount() << ":\\n";
+    settings.stream() << "Thread " << dump_inst.threadID() << ", Frame " << dump_inst.frameCount();
+    if(settings.showTimestamp())
+        settings.stream() << ", Time " << dump_inst.current_time_since_start().count() << " μs\\n";
+    else 
+        settings.stream() << "\\n";
```
lunarpapillo: The formatting in the new code is subtly different than the original: the original `Thread` line had a trailing colon, while the new `Thread` line doesn't have a trailing colon in either case.

I'm not sure whether we promise the output format of this line anywhere.  But I'm concerned that changing it could break people relying on the original format.

scripts/api_dump_generator.py
```diff
@@ -725,7 +729,11 @@
 std::ostream& dump_text_{funcName}(ApiDumpInstance& dump_inst, {funcTypedParams})
 {{
     const ApiDumpSettings& settings(dump_inst.settings());
-    settings.stream() << "Thread " << dump_inst.threadID() << ", Frame " << dump_inst.frameCount() << ":\\n";
+    settings.stream() << "Thread " << dump_inst.threadID() << ", Frame " << dump_inst.frameCount();
```
lunarpapillo: Same issues with the "micro" letter and the trailing colon format.

vkconfig/layer_info.json
```diff
@@ -135,14 +141,19 @@
                 "type": "bool",
                 "default": true
             },
+            "show_timestamp": {
+                "name": "Show Timestamp",
+                "description": "Show the timestamp of function calls since start in microseconds",
+                "type": "bool",
+                "default": true
```
lunarpapillo: defaulting this to `true` will change existing behavior, right?  Is this intended?

charles-lunarg:
Feature request comes from #800 (https://github.com/LunarG/VulkanTools/issues/800)

vkconfig/layer_info.json
charles-lunarg: as discussed in #800, the ability to see the timing was thought to be useful enough to have on by default, with the option to turn it off when doing diffs. This could be added to the "hide address" option, which is the main reason to turn it off

scripts/api_dump_generator.py
charles-lunarg: Thats an oversight on my part.
Adding the colon back is simple enough, and if they are looking for a colon, then there is no good reason to not have it. Putting timestamps in is still a breaking change (if the feature is enabled) so parsers still need to update. Or disable timestamps, in which their code should still work fine.

scripts/api_dump_generator.py
```diff
@@ -698,7 +698,11 @@
 std::ostream& dump_text_{funcName}(ApiDumpInstance& dump_inst, {funcReturn} result, {funcTypedParams})
 {{
     const ApiDumpSettings& settings(dump_inst.settings());
-    settings.stream() << "Thread " << dump_inst.threadID() << ", Frame " << dump_inst.frameCount() << ":\\n";
+    settings.stream() << "Thread " << dump_inst.threadID() << ", Frame " << dump_inst.frameCount();
+    if(settings.showTimestamp())
+        settings.stream() << ", Time " << dump_inst.current_time_since_start().count() << " μs\\n";
```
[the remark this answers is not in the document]
charles-lunarg: @davidlunarg Is there any guarantee of the output format? μs happens to be a very succinct way to describe microseconds. Most functions take 5-10 μs to run, so the timer has enough precision. Milliseconds would be ASCII compatible but wouldn't have enough precision. 
While μs is the "official" symbol, us could be used as a replacement if we can't guarantee UTF-8.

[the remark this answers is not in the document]
davidlunarg: apidump has output ascii in the past, we should continue that. So I would opt for "us".

vkconfig/layer_info.json
charles-lunarg: The defaults have been changed to false, to preserve existing behavior.
While api_dump isn't meant as a profiler, it is at least nice to have the option to see timing.

davidlunarg: My opinion is that it should be off by default, so that scripts that process api dump output are not broken.

layersvt/api_dump_layer.md
```diff
@@ -105,6 +105,7 @@ Detailed Output | `VK_APIDUMP_DETAILED` | `lunarg_api_dump.detailed` | true | Ge
 No Addresses/Handles | `VK_APIDUMP_NO_ADDR` | `lunarg_api_dump.no_addr` | false | Generate output without addresses or handles (which can vary run to run. Instead use the placeholder value "address".
 Flush After Every Command | `VK_APIDUMP_FLUSH` | `lunarg_api_dump.flush` | true | Flush after every API command's output
 Output format | `VK_APIDUMP_OUTPUT_FORMAT` | `lunarg_api_dump.output_format` | `text` | Output the API Dump information as a text file (`text`), an HTML-formated file (`html`), or a json file (`json`).
+Show Timestamps | `VK_APIDUMP_TIMESTAMP` | `lunarg_api_dump.show_timestamp` | true | Show the timestamp of function calls since start in microseconds
```
davidlunarg: start of what? Might be better to describe it as "Show a timestamp for function calls. Timestamps are microseconds relative to the first Vulkan function call."

charles-lunarg: It starts the clock whenever API Dump is initialized. I'll amend it to reflect that.

scripts/api_dump_generator.py
charles-lunarg: Fixed and made timestamps disabled by default.

layersvt/api_dump_layer.md
charles-lunarg: Amended.

PR #857, opened 2019-07-23
vktrace: fix type mismatch on error message

lunarpapillo:
"%u" is a format reference to a 32-bit unsigned integer.
We're passing a uint64_t for formatting, which is unlikely
to work correctly.

"%ju" is a conversion for an intmax_t on both Windows and
Linux.  Casting the corresponding argument to this type
ensures that the format string and argument are in sync
regardless of how uint64_t is defined.

Change-Id: I92381b3b20d1c356fcf7b2df23834f9aaec03b8e


davidlunarg:
Looks good, just need to address the clang format error.

PR #867, opened 2019-08-05
apidump: Dump functions before callind ICD

charles-lunarg:
Functions without a result or a non-const pointer
parameter now will dump their values before calling
into the ICD. Calls that crash the driver can be
dumped before the crash happens, helping to diagnose
bad calls.

files modified:
	scripts/api_dump_generator.py

Change-Id: I4b8ecbcb43875055f7a5cc6253316be00ba963c0


scripts/api_dump_generator.py
```diff
@@ -434,6 +446,8 @@
 std::ostream& dump_text_{unName}(const {unName}& object, const ApiDumpSettings& settings, int indents);
 @end union
 
+//======================== pNext Chain Implementation =======================//
+
```
davidlunarg: Best to not include changes not related to what this pull request implements/fixes. Submit a separate pull request.

charles-lunarg:
What cases are there were parameters are modified? 
Currently any non const pointer parameters disable the early dumping. Are there places where parameters are modfied other than this?

charles-lunarg:
Fixed the vktracedump issue that the previous commit worked around. Compile time should be more reasonable now.

Issue #872, opened 2019-08-08
Add vk_layer_settings.txt support to screenshot layer


bradgrantham-lunarg:
environment variable |  proposed vk_layer_settings key  |  Description
--------------------|-----------|----------------------------------------
VK_SCREENSHOT_FRAMES     |  lunarg_screenshot.frames        |  Set to a comma-separated list of frame numbers.
VK_SCREENSHOT_DIR        |  lunarg_screenshot.dir           |  Set to specify the directory in which to create the screenshot files.
VK_SCREENSHOT_FORMAT     |  lunarg_screenshot.format        |  Set to desired pixel component format to read, one of UNORM, SNORM, USCALED, SSCALED, UINT, SINT, SRGB

Adding support for these keys to these layers and then adding these keys to vkconfig/layer_info.json will enable vkconfig to control the screenshot layer.

jeremyk-lunarg:
Fixed by PR #884 which has been merged.

PR #884, opened 2019-08-15
layers: Add settings file support to screenshot

jeremyk-lunarg:
Added support for vk_layer_settings.txt to
VK_LAYER_LUNARG_screenshot with the following options
equivalent to the following environment variable.

lunarg_screenshot.frames -> VK_SCREENSHOT_FRAMES
lunarg_screenshot.dir -> VK_SCREENSHOT_DIR
lunarg_screenshot.format -> VK_SCREENSHOT_FORMAT

Addresses issue #872

### DIFF
--- a/layersvt/screenshot.cpp
+++ b/layersvt/screenshot.cpp
@@ -50,16 +50,20 @@ using namespace std;
 #include <sys/system_properties.h>
 
 static char android_env[64] = {};
-const char *env_var = "debug.vulkan.screenshot";
-const char *env_var_old = env_var;
+const char *env_var_frames = "debug.vulkan.screenshot";
+const char *env_var_old = env_var_frames;
 const char *env_var_format = "debug.vulkan.screenshot.format";
 const char *env_var_dir = "debug.vulkan.screenshot.dir";
-#else  //Linux or Windows
+#else  // Linux or Windows
 const char *env_var_old = "_VK_SCREENSHOT";
-const char *env_var = "VK_SCREENSHOT_FRAMES";
+const char *env_var_frames = "VK_SCREENSHOT_FRAMES";
 const char *env_var_format = "VK_SCREENSHOT_FORMAT";
 const char *env_var_dir = "VK_SCREENSHOT_DIR";
 #endif
+
+const char *settings_option_frames = "lunarg_screenshot.frames";
+const char *settings_option_format = "lunarg_screenshot.format";
+const char *settings_option_dir = "lunarg_screenshot.dir";
 
 #ifdef ANDROID
 char *android_exec(const char *cmd) {
@@ -77,7 +81,6 @@ char *android_exec(const char *cmd) {
         android_env_str.erase(android_env_str.find_last_not_of(" \n\r\t") + 1);
         snprintf(android_env, sizeof(android_env), "%s", android_env_str.c_str());
         return android_env;
-
     }
 
     return nullptr;
@@ -126,7 +129,8 @@ namespace screenshot {
 static int globalLockInitialized = 0;
 static loader_platform_thread_mutex globalLock;
 
-const char *vk_screenshot_format = nullptr;
+const char *vk_screenshot_dir = nullptr;
+bool vk_screenshot_dir_used_env_var = false;
 
 bool printFormatWarning = true;
 
@@ -174,7 +178,9 @@ static unordered_map<VkDevice, DeviceMapStruct *> deviceMap;
 static unordered_map<VkQueue, uint32_t> queueIndexMap;
 
 // unordered map: associates a physical device with an instance
-typedef struct { VkInstance instance; } PhysDeviceMapStruct;
+typedef struct {
+    VkInstance instance;
+} PhysDeviceMapStruct;
 static unordered_map<VkPhysicalDevice, PhysDeviceMapStruct *> physDeviceMap;
 
 // set: list of frames to take screenshots without duplication.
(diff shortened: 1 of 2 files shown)


vkconfig/layer_info.json
```diff
@@ -366,6 +366,35 @@
                 "type": "save_file",
                 "default": "stdout"
             }
+        },
+        "VK_LAYER_LUNARG_screenshot": {
+            "frames": {
+                "name": "Frames",
+                "description": "Comma separated list of frames to output as screen shots or a range of frames with a start, count, and optional interval separated by a dash. Setting the variable to \"all\" will output every frame. Example: \"5-8-2\" will output frame 5, continue until frame 13, dumping every other frame. Example: \"3,8-2\" will output frames 3, 8, and 9.",
+                "type": "string",
+                "default": "0-0"
```
bradgrantham-lunarg: Does this mean it will take a screenshot of frame 0?  Could we make the default be the empty string?

vkconfig/layer_info.json
```diff
@@ -366,6 +366,35 @@
...
+                "options": {
+                    "UNORM": "UNORM",
+                    "SNORM": "SNORM",
+                    "USCALED": "USCALED",
+                    "SSCALED": "SSCALED",
+                    "UINT": "UINT",
+                    "SINT": "SINT",
+                    "SRGB": "SRGB"
+                },
+                "default": "SRGB"
```
bradgrantham-lunarg: In order to match the default (which appears to be to use the swapchain colorspace) where there's no env var set or the env var is the empty string, this may need to have an option "UNDEFINED" or "USE_SWAPCHAIN_COLORSPACE" or some such, which would be the default.  I am uncertain.  What do you think?

layersvt/screenshot.cpp
```diff
@@ -254,11 +267,32 @@ void readScreenShotFormatENV(void) {
...
-                            "Swapchain Colorspace will be used instead\n", vk_screenshot_format);
+            fprintf(stderr,
+                    "Selected format:%s\nIs NOT in the list:\nUNORM, SNORM, USCALED, SSCALED, UINT, SINT, SRGB\n"
+                    "Swapchain Colorspace will be used instead\n",
+                    vk_screenshot_format);
 #endif
         }
     }
+
+    if (env_var != NULL && strlen(env_var) > 0) {
```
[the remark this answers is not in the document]
jeremyk-lunarg: In the case of an empty string as an environment variable, wouldn't it be best to use the vk_layers_settings.txt file, just in case the user accidentally set the environment variable to an empty string instead of unsetting it?

layersvt/screenshot.cpp
```diff
@@ -254,11 +267,32 @@ void readScreenShotFormatENV(void) {
...
+        local_free_getenv(vk_screenshot_format);
```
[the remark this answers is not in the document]
jeremyk-lunarg: Yeah, I can do that.

vkconfig/layer_info.json
jeremyk-lunarg: It will take a screenshot of frame 0 which is just the first frame.
I can make the default and empty string.

vkconfig/layer_info.json
jeremyk-lunarg: I think the default should be "USE_SWAPCHAIN_COLORSPACE".
Also, I think there should be an added conditional to the function that gets the format that tells it not to print the error message if the user has the environment variable or settings set to "USE_SWAPCHAIN_COLORSPACE".

bradgrantham-lunarg: Sounds good.

jeremyk-lunarg:
Updated.

vkconfig/layer_info.json
```diff
@@ -366,6 +366,35 @@
...
+                "options": {
+                    "UNORM": "UNORM",
+                    "SNORM": "SNORM",
+                    "USCALED": "USCALED",
+                    "SSCALED": "SSCALED",
+                    "UINT": "UINT",
+                    "SINT": "SINT",
+                    "SRGB": "SRGB"
+                },
+                "default": "USE_SWAPCHAIN_COLORSPACE"
```
bradgrantham-lunarg: Remember to add USE_SWAPCHAIN_COLORSPACE to the options list.

bradgrantham-lunarg: I don't see it in the source yet, so I assume this change is in progress.

jeremyk-lunarg:
Updated.

layersvt/screenshot.cpp
```diff
@@ -247,18 +264,43 @@ void readScreenShotFormatENV(void) {
             userColorSpaceFormat = UINT;
         } else if (!strcmp(vk_screenshot_format, "SINT")) {
             userColorSpaceFormat = SINT;
-        } else {
+        } else if (strcmp(vk_screenshot_format, "USE_SWAPCHAIN_COLORSPACE")) {
 #ifdef ANDROID
```
bradgrantham-lunarg: Thanks for adding this!  I think my preference here is to have the "else" with USE_SWAPCHAIN_COLORSPACE not warn, and then have an "else" that warns like before.

layersvt/screenshot.cpp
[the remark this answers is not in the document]
bradgrantham-lunarg: Great!

layersvt/screenshot.cpp
jeremyk-lunarg: I'm a bit confused by this request.
strcmp(char*, char*) would return a zero if format is USE_SWAPCHAIN_COLORSPACE, thus preventing the warning from being printed.

bradgrantham-lunarg: Yeah, you're right.  Ugh, I can't read the use of strcmp if it's not "!= 0" or "== 0".  It's consistent with the previous commits, but I really dislike the use in previous commits as if strcmp returns a bool.  Only the use of strcmp at line 1045 is the way I like to see it.  As a bonus, if you wanted to turn the use of "!strcmp(...)" to "strcmp(...) == 0" and "strcmp(...)" to "strcmp(...) != 0", it would make me very happy.  But I'll resolve this particular request.  Thanks!

jeremyk-lunarg: Sure I can do that!
After your comment I was thinking they are easy to misread and I was wondering if I should change them to be more explicit.
Do you think I should also use strncmp to prevent potential buffer overflow or do you think that is unnecessary?

bradgrantham-lunarg: I think strcmp is probably sufficient.  I don't think I expect getenv nor the layer settings functions nor a compile-time string to lack a NUL under normal operation.

@mikes-lunarg, @davidlunarg  do you think we should use strncmp?  I'm not sure about the security mindset we should have for getenv, vk_layer_settings.txt in Vulkan layers.

jeremyk-lunarg:
Updated

Issue #873, opened 2019-08-08
Add vk_layer_settings.txt support to device simulation layer 


bradgrantham-lunarg:
Environment Variable     |  Suggested Key Name                     |  Description
-------------------------|-----------------------------------------|--------------
VK_DEVSIM_FILENAME       |  lunarg_device_simulation.filename      |  Name of one or more configuration file(s) to load.
VK_DEVSIM_DEBUG_ENABLE   |  lunarg_device_simulation.debug_enable  |  A non-zero integer enables debug message output.
VK_DEVSIM_EXIT_ON_ERROR  |  lunarg_device_simulation.exit_on_error |  A non-zero integer enables exit-on-error.

Adding support for these keys to these layers and then adding these keys to vkconfig/layer_info.json will enable vkconfig to control the device simulation layer.

Issue #900, opened 2019-08-20
DevSim Layer: integration with vkConfig


mikew-lunarg:
(Edited and moved from https://github.com/LunarG/Projects/issues/142 at @bradgrantham suggestion)

DevSim currently gets settings via environment variables.  It will be enhanced to get settings from vk_layer_settings.txt, see #873.  Need to document precedence between envars and settings, if both are active.

DevSim usemodel is distinct from both explicit and implicit layers (a separate category?)  It is not loaded by the app, so not explicit.  It is not always loaded before app startup, so not implicit.  It is only loaded before app startup when enabled by envars.  So maybe a "conditional implicit layer, disabled by default"?  vkConfig's interface could make this an easier concept.

DevSim is sensitive to it's placement in the layer stack.  Ideally it is "right next to the driver", so it can cleanly override queries of the driver.  There are other layers (e.g. Optimus) that also expect to be "right next to the driver".  The relative order of Optimus and DevSim depends on the usecase.

mark-lunarg:
Can this be closed?

mikew-lunarg:
Overdue for closing

PR #903, opened 2019-08-23
layers: Add settings file support to DevSim

jeremyk-lunarg:
Added support for vk_layer_settings.txt to
VK_LAYER_LUNARG_device_simulation with the following options
equivalent to the following environment variable.

lunarg_device_simulation.filename -> VK_DEVSIM_FILENAME
lunarg_device_simulation.debug_enable -> VK_DEVSIM_DEBUG_ENABLE
lunarg_device_simulation.exit_on_error -> VK_DEVSIM_EXIT_ON_ERROR

Fixes issue #873 
Helps progress issue #900

### DIFF
--- a/layersvt/device_simulation.md
+++ b/layersvt/device_simulation.md
@@ -152,6 +152,14 @@ If you wish DevSim to terminate on errors, set the `VK_DEVSIM_EXIT_ON_ERROR` env
 * `VK_DEVSIM_DEBUG_ENABLE` - A non-zero integer enables debug message output.
 * `VK_DEVSIM_EXIT_ON_ERROR` - A non-zero integer enables exit-on-error.
 
+#### vk_layer_settings.txt Options
+
+* `lunarg_device_simulation.filename` - Equivalent to the `VK_DEVSIM_FILENAME` environment variable.
+* `lunarg_device_simulation.debug_enable` - Equivalent to the `VK_DEVSIM_DEBUG_ENABLE` environment variable.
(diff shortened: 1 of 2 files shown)


jeremyk-lunarg:
If you'd like, I can do that.  Would this just be added to the README.md file in the `layersvt`?  I was thinking I should put a note about this under the "Using Layers" section since this same standard would apply to all utility layers.

Should I also add to each individual layer document the names of that layer's settings variables in vk_layer_settings.txt?

mikew-lunarg:
I think docs should be part of this PR, since this code implements a precedence, but it's not described.

I think it should be in device_simulation.md, as there's no hint of another setting mechanism there.  But where else?  Fair question.

jeremyk-lunarg:
I didn't add documentation for Screenshot Layer, but it implements a similar change.
Maybe an update to docs for both should be done in a separate PR all together?

bradgrantham-lunarg:
Yeah, not updating screenshot's docs is an oversight.  Sorry for not
catching that!

Yes, a separate PR for docs update for screenshot makes sense.  I would add
the DevSim docs update to the devsim issue, since it really needs to be
done with the code changes.

jeremyk-lunarg:
Sounds good! I'll update the docs here and push them up to this PR in a separate commit.

KarenGhavam-lunarG:
Would it make sense to add some typical entries to the vk_layers_settings.txt file found in this repository  for both the devsim and the screenshot layer? Today this file gives examples for apidump only. Then these examples would get shipped with the SDK as well...

jeremyk-lunarg:
It looks like they already do this with Vulkan-ValidationLayers.  I think that's a good idea.

PR #917, opened 2019-09-17
vkvia: update search for installed SDK packages

johnzupin:
The "lunarg-vulkan-sdk" package has been out of commission for some time now. I think this package is from the old .run days.

Here is link to the arch 1.0.3 package that is marked as out-of-date in arch: https://aur.archlinux.org/packages/lunarg-vulkan-sdk/. When a user runs vkvia an error is thrown to the console that the package was not found, this has led to some confusion for some users.

Anyway, this change only searches Debian systems (the only package LunarG provides for) and has been fixed to look for the correct "vulkan-sdk" package.

### DIFF
--- a/via/via_system_linux.cpp
+++ b/via/via_system_linux.cpp
@@ -1173,15 +1173,15 @@ ViaSystem::ViaResults ViaSystemLinux::PrintSystemSdkInfo() {
     // Next, try system install items
     std::string upper_os_name = _os_name;
     std::transform(upper_os_name.begin(), upper_os_name.end(), upper_os_name.begin(), ::toupper);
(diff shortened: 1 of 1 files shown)


KarenGhavam-lunarG:
@johnzupin Is this still a desired PR we should complete?

johnzupin:
Yeah, let me rebase and review

johnzupin:
@jeremy-lunarg and/or @davidlunarg could you guys give this a quick review?

johnzupin:
For reference here is the issue that sparked this PR: https://vulkan.lunarg.com/issue/view/5d789552b736866d88e7ea10

PR #919, opened 2019-09-24
vkvia: Add option to VIA to disable cube tests

jeremyk-lunarg:
Added command line option `--disable_cube_tests` to vkvia
to prevent cube from being run at the end of VIA tests.


mark-lunarg:
Be sure to squash the commits before pushing them.  LGTM.

PR #924, opened 2019-10-17
Up VkConfig minimum UI control sizes to correctly display text

mark-lunarg:


jeremyk-lunarg:
It works on my machine.
When I build it with Qt 5.13.0 I get warnings that qSort is deprecated.
Do we want to do what is suggested [here](https://doc.qt.io/qt-5/qtalgorithms-obsolete.html#qSort-2) to remove these warnings?

mark-lunarg:
Sure, could you do that in a separate PR?

jeremyk-lunarg:
Yep!

PR #926, opened 2019-10-29
tests: Restore original vulkaninfo behvavior

lunarpapillo:
A recent change to vulkaninfo made the extensive format
listing optional (available via the "--show-formats" switch).

The demo layer tests happen to exercise the vulkaninfo
executable, looking for and counting the number of calls
to vkGetPhysicalDeviceFormatProperties; these tests started
failing with the new vulkaninfo because the number of format
queries went from over 200 to 16.

These changes to the test scripts pass the "--show-formats"
switch to vulkaninfo, which restores the original behavior.

Change-Id: I6076c3b573751a4bfcacd552b8f6a021ec7743fd


mark-lunarg:
> Should the --show-formats option be in the vulkaninfo help text? If so, adding it would probably be outside the scope of this pull request.  

👍

PR #934, opened 2019-11-25
api_dump: keep lock held between printing head and body

charles-lunarg:
Previously, api_dump would release the output lock when finished printing
the head of a call and reaquiring it when printing the body. This enabled
multi threaded applications to interleave output resulting in garbled output.
This commit fixed that by keeping the output lock held for the duration of
the call, at the cost of serializing any calls that are made to vulkan.

This commit also switched api_dump to std::mutex & std::thread::id instead of using
vk_loader_platform.h for the threading primitives.

Changes:
	modified:   layersvt/api_dump.h
	modified:   scripts/api_dump_generator.py

Change-Id: Ia4b0cabc839c93798eeeb811d6cb322b5b197194

### DIFF
--- a/layersvt/api_dump.h
+++ b/layersvt/api_dump.h
@@ -25,7 +25,6 @@
 
 #define NOMINMAX
 
-#include "vk_loader_platform.h"
 #include "vulkan/vk_layer.h"
 #include "vk_layer_config.h"
 #include "vk_layer_table.h"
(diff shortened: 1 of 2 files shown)


charles-lunarg:
The change to std::recursive_mutex is necessary as a means to prevent deadlock, and the previous locks didn't have a way to create recursive mutexes.
Switching to std::thread::id would remove the need for "vk_loader_platform.h". It appears that the dependency was only there due to standard conformance being an issue, and with C++11 being a hard requirement, std::thread::id is a drop in replacement for the same functionality.

PR #950, opened 2020-01-22
VulkanTools: bump vk version in several files

davidlunarg:
Change-Id: I56863d3c74cda35253e8b86dbade05448b537e3a

PR #951, opened 2020-01-22
layersvt: Fix layer versions

mikew-lunarg:
Layers hardcoded to specVersion 1.0 should not use VK_HEADER_VERSION as
the patchlevel.  Hardcoding the patchlevel to 1.0.68.

Change-Id: I469eeb44d84facea8c0c4b28b97d9589fb364406


jeremyk-lunarg:
Why 1.0.68?

mikew-lunarg:
That was the last 1.0 version.  The next header was 1.1.70

lunarpapillo:
> I see devsim and screenshot, but not vktrace... is that handled in another PR?
It looks like David's pull request #950 covers these and more.

mikew-lunarg:
Merged into #950

PR #952, opened 2020-01-22
scripts: Fix api_dump layer version

mikew-lunarg:
Update codegen layer using VK_HEADER_VERSION to 1.2

Change-Id: I7c85b418439b2812acd0e64ddcb8ddb31a1e140a


mikew-lunarg:
Merged into #950

PR #958, opened 2020-02-11
via: Add output usable by VkConfig

jeremyk-lunarg:
Added argument `--vkconfig_output` to send VIA output to
a JSON file that can be parsed by VkConfig.

### DIFF
--- a/via/via_system.hpp
+++ b/via/via_system.hpp
@@ -72,7 +72,7 @@ class ViaSystem {
 
     enum ViaElementAlign { VIA_ALIGN_LEFT = 0, VIA_ALIGN_CENTER, VIA_ALIGN_RIGHT };
 
-    // HTML methods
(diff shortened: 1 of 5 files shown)


via/via_system.hpp
```diff
@@ -158,12 +183,15 @@ class ViaSystem {
...
     std::string _cur_path;
-    std::string _html_file;
-    std::ofstream _html_ofstream;
+    std::string _out_file;
+    std::ofstream _out_ofstream;
 
     // Command Line Argument items
     bool _run_cube_tests;
 
+    enum ViaFileFormat { VIA_HTML_FORMAT = 0, VIA_VKCONFIG_FORMAT };
```
charles-lunarg: Nit: I generally prefer to use `enum class` wherever possible.

Issue #965, opened 2020-03-05
Android screenshot layer settings don't work


lunarpapillo:
The code to extract screenshot layer settings in Android uses the `getprop` command to read back values for `debug.vulkan.screenshot` and `debug.vulkan.screenshot.dir`.  But the code itself returns values in a static character array:
```
        static char android_env[64] = {};

        char *android_exec(const char *cmd) {
            FILE *pipe = popen(cmd, "r");
            if (pipe != nullptr) {
                fgets(android_env, 64, pipe);
                pclose(pipe);
            }

            // Only if the value is set will we get a string back
            if (strlen(android_env) > 0) {
                __android_log_print(ANDROID_LOG_INFO, "screenshot", "Vulkan screenshot layer capturing: %s", android_env);
                // Do a right strip of " ", "\n", "\r", "\t" for the android_env string
                string android_env_str(android_env);
                android_env_str.erase(android_env_str.find_last_not_of(" \n\r\t") + 1);
                snprintf(android_env, sizeof(android_env), "%s", android_env_str.c_str());
                return android_env;
            }

            return nullptr;
        }

        char *android_getenv(const char *key) {
            std::string command("getprop ");
            command += key;
            return android_exec(command.c_str());
        }
```
This means that the same buffer is returned for each query, which effectively causes all settings to become equal to the last setting queried.

An excerpt from a `logcat` is illustrative.  After:
```
setprop debug.vulkan.screenshot 0,89,185
setprop debug.vulkan.screenshot.dir /mnt/sdcard
```
the `logcat` shows:
```
        03-05 01:01:32.640 24567 24675 I vulkan  : Loaded layer VK_LAYER_LUNARG_screenshot
        03-05 01:01:32.657 24567 24675 I screenshot: Vulkan screenshot layer capturing:
        03-05 01:01:32.670 24567 24675 I screenshot: Vulkan screenshot layer capturing: /mnt/sdcard
        03-05 01:01:32.683 24567 24675 I screenshot: Vulkan screenshot layer capturing: 0,89,185
        03-05 01:01:33.004 24567 24675 I screenshot: Screen capture file is: 0,89,185/0.ppm
        03-05 01:01:33.011 24567 24675 D screenshot: Failed to open output file: 0,89,185/0.ppm.  Be sure to grant read and write permissions.
        03-05 01:01:35.669 24567 24675 I screenshot: Screen capture file is: 0,89,185/89.ppm
[...]

PR #966, opened 2020-03-06
Android: screenshot layer #965

lunarpapillo:
https://github.com/LunarG/VulkanTools/issues/965

On Android, A single static buffer was being used to hold the
results of all of the screenshot layer settings
(debug.vulkan.screenshot, debug.vulkan.screenshot.dir,
debug.vulkan.screenshot.format).  All values would appear to
be the same as the last queried value.

This fix saves separate static buffers for each setting.
It also uses the Android PROP_VALUE_MAX to indicate the
size of each buffer (the old value of 64 could be too small
in some circumstances), and uses the Android
__system_property_get() function to query each property
(rather than using the "getprop" command and reading results back
from a pipe, which is unnecessarily complicated).

Finally, the screenshot documentation was updated;
the old documentation incorrectly referred to "debug.vulkan.dir"
as the location where screenshots would be saved.

### DIFF
--- a/layersvt/screenshot.cpp
+++ b/layersvt/screenshot.cpp
@@ -49,11 +49,14 @@ using namespace std;
 #include <android/log.h>
 #include <sys/system_properties.h>
 
-static char android_env[64] = {};
 const char *env_var_frames = "debug.vulkan.screenshot";
 const char *env_var_old = env_var_frames;
 const char *env_var_format = "debug.vulkan.screenshot.format";
 const char *env_var_dir = "debug.vulkan.screenshot.dir";
+static char property_frames[PROP_VALUE_MAX] = {};
+static char property_format[PROP_VALUE_MAX] = {};
+static char property_dir[PROP_VALUE_MAX] = {};
+static char property_unknown[PROP_VALUE_MAX] = {};
 #else  // Linux or Windows
 const char *env_var_old = "_VK_SCREENSHOT";
 const char *env_var_frames = "VK_SCREENSHOT_FRAMES";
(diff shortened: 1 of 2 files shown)


lunarpapillo:
FYI... the change passed current **VulkanTools** tests in CI on Windows, Linux, and Android (though I note that the current tests are insufficient to detect the problem).

`gfxreconstruct` tests run by hand verify that both `debug.vulkan.screenshot` and `debug.vulkan.screenshot.dir` can be used at the same time, and that the Android log shows the properties being queried:
```
$ adb -s 93TAX0AQL4 logcat -d | grep screenshot
03-06 00:36:35.939 31228 31283 D vulkan  : added global layer 'VK_LAYER_LUNARG_screenshot' from library '/data/app/com.lunarg.gfxreconstruct.replay-yZYIsMomnHw2xm1kj_CP5Q==/lib/arm64/libVkLayer_screenshot.so'
03-06 00:49:04.741 31228 31406 I vulkan  : Loaded layer VK_LAYER_LUNARG_screenshot
03-06 00:49:04.745 31228 31406 I screenshot: Vulkan screenshot layer property: debug.vulkan.screenshot.format=
03-06 00:49:04.745 31228 31406 I screenshot: Vulkan screenshot layer property: debug.vulkan.screenshot.dir=/mnt/sdcard
03-06 00:49:04.745 31228 31406 I screenshot: Vulkan screenshot layer property: debug.vulkan.screenshot=0,89,185
03-06 00:49:05.093 31228 31406 I screenshot: Screen capture file is: /mnt/sdcard/0.ppm
03-06 00:49:07.754 31228 31406 I screenshot: Screen capture file is: /mnt/sdcard/89.ppm
03-06 00:49:10.242 31228 31406 I screenshot: Screen capture file is: /mnt/sdcard/185.ppm
```

lunarpapillo:
Abandoned in favor of David's superior solution.

PR #974, opened 2020-03-27
cmake: Add per-layer BUILD switches and remove BUILD_LAYERSVT

mark-lunarg:
Apidump files were generated regardless of the build target if any layer was requested.

Wish I could run it through CI but it's still indisposed.


mark-lunarg:
This also passed CI.

PR #975, opened 2020-03-28
apidump: Fix json output in multithreaded apps

### LINKED ISSUE #962: api_dump: corrupted json dump in multithreaded application.
Hi,
After text dump was fixed #932, I think it is still not correct around frame start/end.
My dump using api_dump from SDK 1.2.131.2 produced corrupted json.

Validation output:
`jq-win64.exe . vk_apidump.json`

> parse error: Unmatched ']' at line 39955226, column 1

And this looks like this:

```
[...]
39955193                     "type" : "VkMemoryRequirements*",
39955194                     "name" : "pMemoryRequirements",
39955195                     "address" : "000000E3175B02D0",
39955196                     "members" :
39955197                     [
39955198                         {
39955199                             "type" : "VkDeviceSize",
39955204                             "type" : "VkDeviceSize",
39955205                             "name" : "alignment",
39955206                             "value" : "1024"
39955207                         },
39955208                         {
39955209                             "type" : "uint32_t",
39955210                             "name" : "memoryTypeBits",
39955211                             "value" : "258"
39955212                         }
39955213                     ]
39955214                 }
39955215             ]
39955216         },
39955217         {
39955218             "name" : "vkBindImageMemory",
39955219             "thread" : "Thread 0",
39955220             "time" : "47907344 us",
39955221             "returnType" : "VkResult",
39955222             "returnValue" : "VK_SUCCESS",
39955223             "args" :
39955224             [
39955225                 {
39955226 ]
39955227 }                    "type" : "VkDevice",
39955228                     "name" : "device",
39955229 ,
39955230 {
39955231     "frameNumber" : "1",
39955232                         "value" : "00000243F675BAA0"
39955233                 },
39955234                 {
39955235 "apiCalls" :
39955236     [
39955237                     "type" : "VkImage",
39955238                     "name" : "image",
39955239                     "value" : "00000244AA326DF0"
39955240                 },
39955241                 {
39955242                     "type" : "VkDeviceMemory",
39955243                     "name" : "memory",
39955244                     "value" : "00000244DDFB7180"
39955245                 },
39955246                 {
39955247                     "type" : "VkDeviceSize",
39955248                     "name" : "memoryOffset",
39955249                     "value" : "23658496"
39955250                 }
39955251             ]
39955252         }        {
39955253             "name" : "vkCreateImageView",
39955254             "thread" : "Thread 0",
[...]
```

This app is multithreaded, so my guess is that something is wrong with synchronizing frames and output.
Text dump does not need to close frame section, it is included in every chunk (empty setupInterFrameOutputFormatting())
Json is more structured and have to close frame section for all threads. Maybe starting new frame should wait for other calls to finish first.

charles-lunarg:
The call to next frame has to do setup, of which requires printing to
the output. The issue was this call happened outside of the lock,
causing corrupt output. This commit fixes it by keeping the lock while
the output occurs.

Fixes #962 

Change-Id: I6b05e53b9587f91bf7d85cee7399ca1b16d500fa

### DIFF
--- a/scripts/api_dump_generator.py
+++ b/scripts/api_dump_generator.py
@@ -382,6 +382,7 @@ VK_LAYER_EXPORT VKAPI_ATTR {funcReturn} VKAPI_CALL {funcName}({funcTypedParams})
 @foreach function where('{funcName}' == 'vkQueuePresentKHR')
 VK_LAYER_EXPORT VKAPI_ATTR {funcReturn} VKAPI_CALL {funcName}({funcTypedParams})
 {{
+    ApiDumpInstance::current().outputMutex()->lock();
     dump_head_{funcName}(ApiDumpInstance::current(), {funcNamedParams});
 
     {funcReturn} result = device_dispatch_table({funcDispatchParam})->{funcShortName}({funcNamedParams});
@@ -389,6 +390,7 @@ VK_LAYER_EXPORT VKAPI_ATTR {funcReturn} VKAPI_CALL {funcName}({funcTypedParams})
     dump_body_{funcName}(ApiDumpInstance::current(), result, {funcNamedParams});
 
     ApiDumpInstance::current().nextFrame();
+    ApiDumpInstance::current().outputMutex()->unlock();
     return result;
 }}
 @end function


charles-lunarg:
No because all the relevant function's bodies have `dump_head_funcName()` at the beginning and `dump_body_funcName()` at the end which do the locking. 
vkQueuePresentKHR is the special case as the nextFrame() has to occur after the function's body has been dumped, meaning it can't use the lock that occurs in dump_head (& unlock in dump_body)

PR #988, opened 2020-04-16
apidump: remove special case from autogen

charles-lunarg:
The VkSurfaceFullScreenExclusiveWin32InfoEXT was special cased in the
autogen, but shows no reason to be so. This commit reverts it back to
just anothe plain old struct to be processed and printed.

Change-Id: Ib53adcf10a37d9c4603da4bfee1a1efd761b3420

### DIFF
--- a/scripts/api_dump_generator.py
+++ b/scripts/api_dump_generator.py
@@ -527,13 +527,8 @@ TEXT_CODEGEN = """
 
 #include "api_dump.h"
 
-@foreach struct where('{sctName}' != 'VkSurfaceFullScreenExclusiveWin32InfoEXT')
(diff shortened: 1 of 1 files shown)


shannon-lunarg:
I remember CI was failing on 64-bit Windows builds without the ifdef guards around `dump_text_HMONITOR` and FullScreenExclusiveWin32-related code but I don't remember specifics past that. I believe implementing the guards at that time was a temporary solution and wasn't revisited as intended until now.

charles-lunarg:
Seeing as it has passed tests, are there any special tests we need to perform before merging? (after someone else can give it a proper code review sign off of course)

PR #999, opened 2020-05-21
build: Changes for 1.2.140 header

davidlunarg:
- Propagated 1.2.140 `genvk.py` changes

- Bumped the commit IDs in scripts/known_good.json to 1.2.140.

- Made several changes to code gen because of changes in entries
  in vk.xml.

Change-Id: I515f20442019ca848620825005b5de408219cfc6


scripts/api_dump_generator.py
```diff
@@ -618,12 +618,28 @@
 
 //========================= Basetype Implementations ========================//
 
-@foreach basetype
+@foreach basetype where(not '{baseName}' in ['ANativeWindow', 'AHardwareBuffer', 'CAMetalLayer'])
```
charles-lunarg, line 621: Should this be based instead of the POINTER_TYPES list instead of hardcoding it?

davidlunarg: POINTER_TYPES is not available in the middle of the *_CODEGEN strings.

PR #1003, opened 2020-06-05
Fix generator scripts

djdeath:
Using positional arguments instead of named ones leads to desynchro
between the Vulkan-Headers repo and this one. Since all arguments are
named, better use names and avoid breakage.


davidlunarg:
The change looks good. Please sign the CLA and then I can merge this pull request.

davidlunarg:
This PR will have to be updated because it involves changes to vktraces files, but vktrace has been removed from VulkanTools.

charles-lunarg:
I'm closing this PR because all of the changes to the formatting have been done for the python files in question. While the formatting may not be optimal, they are at least 'one arugment per line', which this PR mainly adds.

PR #1006, opened 2020-06-12
apidump: Append file extension is not present

### LINKED ISSUE #940: Vulkan Configurator: apidump layer outputfile should use an extension consistent with output format
If I specify an output format of html when enabling the api dump layer using vkconfig, the default log filename has an extension of .txt. Would be better if the default file name extension was changed to .html when the output format is changed.

charles-lunarg:
If the filename provided in the settings file does not end with the
corresponding extension for the output type, api dump will append it.

Resolves #940 

The default for vkconfig needs to be extension-less so that if the user does write to a file, the api dump layer uses the correct file extension.

Change-Id: Ib99f7a6ad3c686d8730cf58ed7b5223eebba4617


davidlunarg:
I'm still conflicted on this change. It seems that if I request the output file be "filename", the output should be to "filename". If vkconfig or the apidump layer has to generate a default output file name, it should include an extension appropriate for the type of output requested, i.e. "filename.txt", "filename.html", or "filename.json".

christophe-lunarg:
@charles-lunarg is that change still a think?

charles-lunarg:
Which change are you referring to? Truncating the wrong file type or just adding the correct one on the end?

PR #1011, opened 2020-06-19
Modifications to use new Logo

KarenGhavam-lunarG:
LunarG has a new logo. Modify documentation to leverage the new logo


KarenGhavam-lunarG:
Tony and David. I still haven't modified the VIA logo. That will require some additional artwork changes. These changes are to incorporate new LunarG logos we have received.

PR #1021, opened 2020-07-14
Richard vkconfig layer discovery rebase

richard-lunarg:
This addresses the remaining issues discussed here:
https://github.com/LunarG/Projects/issues/378

Layer discovery from vkconfig is as follows:
VK_LAYER_PATH (if it is present)
User defined layer search paths
Standard Layer Paths for the given OS
VULKAN_SDK path if present

No two layers with the same name may be used with vkconfig. So if there are duplicates, the above order is used as precedence for which one's are used. The specific path for each layer is displayed in layer editor dialog (select which layers to override).
Also, a visual indicator on the main screen is shown if the VK_LAYER_PATH environment variable is detected.

### DIFF
--- a/vkconfig/configurator.h
+++ b/vkconfig/configurator.h
@@ -120,6 +120,7 @@ class Configurator {
     enum { PathCount = LastPath - FirstPath + 1 };
 
     static Configurator& Get();
+    bool InitializeConfigurator(void);
 
     // Need this to check vulkan loader version
     uint32_t vulkan_instance_version;
@@ -174,6 +175,11 @@ class Configurator {
     int GetCustomLayersPathSize() const;
     const QString& GetCustomLayersPath(int path_index) const;
 
+    QStringList VK_LAYER_PATH;  // If this environment variable is set, this contains
+                                // a list of paths that should be searched first for
+                                // Vulkan layers. (Named as environment variable for
+                                // clarity as to where this comes from).
+
    private:
     QStringList custom_layers_paths_;
 
@@ -203,7 +209,7 @@ class Configurator {
     QVector<LayerFile*> available_Layers;  // All the found layers, lumped together
     void LoadAllInstalledLayers();
     const LayerFile* FindLayerNamed(QString layer_name, const char* location = nullptr);
-    void LoadLayersFromPath(const QString& path, QVector<LayerFile*>& layer_list, LayerType type);
+    void LoadLayersFromPath(const QString& path, QVector<LayerFile*>& layer_list);
 
     QVector<Configuration*> available_configurations;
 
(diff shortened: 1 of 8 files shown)


vkconfig/mainwindow.cpp
```diff
@@ -78,6 +78,9 @@ MainWindow::MainWindow(QWidget *parent) : QMainWindow(parent), ui_(new Ui::MainW
     ///////////////////////////////////////////////
     Configurator &configurator = Configurator::Get();
 
+    if (!configurator.VK_LAYER_PATH.isEmpty())
+        ui_->groupBoxProfiles->setTitle("Vulkan Layers Configurations With VK_LAYER_PATH Precedence");
```
christophe-lunarg, line 82: Is this the right place for this? I would expect that layers will come from various place, so it should be a per layer information. Probably for the layer window and the console log at start up

vkconfig/configurator.h
```diff
@@ -174,6 +175,11 @@ class Configurator {
     int GetCustomLayersPathSize() const;
     const QString& GetCustomLayersPath(int path_index) const;
 
+    QStringList VK_LAYER_PATH;  // If this environment variable is set, this contains
```
christophe-lunarg, line 178: vk_layer_path in theory and private ideally so vk_layer_path_

vkconfig/mainwindow.cpp
richard-lunarg: Karl's request was that it be somewhere really up front and obvious. It's not likely to be in effect much on Windows, and almost never on macOS. Just on Linux when someone is "tinkering", so I think it's okay.

vkconfig/configurator.h
richard-lunarg: I think mirroring the environment variable name is more clear. Both in purpose and intent. We are allowed to depart from the typical naming convention if we have a good reason. I think this is a good reason.

richard-lunarg:
Still may be some WIP elements to this update, but sufficient for the July SDK release.

christophe-lunarg:
Was it ok the merge the PR?

richard-lunarg:
I believe so...

Richard S. Wright Jr.
Sr. Graphics Software Engineer
LunarG, Inc. - 3D Graphics Software Solutions
richard@lunarg.com
321-262-9049

christophe-lunarg:
I would expect that at least the reviewer should green / +1 the PR but I am not sure what's LunarG approach

richard-lunarg:
I thought there was... I may have jumped the gun.

Richard S. Wright Jr.
Sr. Graphics Software Engineer
LunarG, Inc. - 3D Graphics Software Solutions
richard@lunarg.com
321-262-9049

PR #1024, opened 2020-07-15
vkconfig: Fix vkconfig initialization when the version run for the first time

christophe-lunarg:
When a specific new version of vkconfig is running, we reinitialize some states to ensure it's running correctly.

- Fix VUID filtering not cropping on rescale
- Fix missing UNASSIGNED VUIDs
- Improve 1440p screen scaling up to 150%
- Default main window height is 640px so that vkconfig fit on
1440p screen with 150% scaling.
- Reinitialize some states when opening a new version

Change-Id: I0f388cce474c1d0c41a7f4ee304a9899323057f8


christophe-lunarg:
Replaced https://github.com/LunarG/VulkanTools/pull/1023 but with the same result...

vkconfig/configurator.cpp
```diff
@@ -1013,7 +1059,7 @@ void Configurator::LoadAllConfigurations() {
         dir.setNameFilters(QStringList() << "*.json");
         QFileInfoList configuration_files = dir.entryInfoList();
 
-        // Loop through all the profiles found and remove them
+        // Loop through all the configuration found and remove them
```
mark-lunarg, line 1062: configurations

vkconfig/dlgprofileeditor.cpp
```diff
@@ -154,7 +154,7 @@ dlgProfileEditor::dlgProfileEditor(QWidget *parent, Configuration *configuration
     }
 
     ui_->lineEditName->setText(configuration_->name);
-    ui_->lineEditDesc->setText(configuration_->description);
+    //ui_->lineEditDesc->setText(configuration_->description);
```
mark-lunarg, line 157: Let's not check in commented-out source if at all possible.

vkconfig/mainwindow.cpp
```diff
@@ -701,8 +712,8 @@ void MainWindow::EditCustomPathsClicked(ContigurationListItem *item) {
     LoadConfigurationList();  // Force a reload
 }
 
-void MainWindow::toolsSetCustomPaths(bool bChecked) {
-    (void)bChecked;
+void MainWindow::toolsSetCustomPaths(bool checked) {
```
mark-lunarg, line 715: Thank you!

richard-lunarg: Why was the second line removed? Is the variable used elsewhere? Otherwise you get a compiler warning about an unused variable.

christophe-lunarg: (void)checked is still below, no idea why the change is not visible here...

vkconfig/dlgprofileeditor.cpp
christophe-lunarg: Fixed

vkconfig/configurator.cpp
christophe-lunarg: Fixed

vkconfig/dlgprofileeditor.cpp
```diff
@@ -391,7 +390,7 @@ void dlgProfileEditor::currentLayerChanged(QTreeWidgetItem *pCurrent, QTreeWidge
     //    // Is this layer Force on?
     //    settingsEditor.SetEnabled(pLayerItem->pLayer->bActive);
 
-    ui_->lineEditDesc->setText(pLayerItem->layer_file->description);
+    //ui_->lineEditDesc->setText(pLayerItem->layer_file->description);
```
mark-lunarg, line 393: Why are these lines commented out?

christophe-lunarg: The description UI field doesn't exist anymore as it was not really implemented...

This was used on multiple instances, remove both of them...

christophe-lunarg:
Replaced with https://github.com/LunarG/VulkanTools/pull/1027..

PR #1027, opened 2020-07-15
vkconfig: Fix vkconfig initialization when the version run for the first time

christophe-lunarg:
- Fix VUID filtering not cropping on rescale

- Fix missing UNASSIGNED VUIDs

- Improve 1440p screen scaling up to 150%

Default main window height is 640px so that vkconfig fit on
1440p screen with 150% scaling.

- Reinitialize some states when opening a new version

Change-Id: I6f7c9326e70db870bf36473165f70658154d785c


mark-lunarg:
@christophe-lunarg, please squash these two commits and force push to this branch.  The clang-format changes should go in the same commits with the formatting issues.

vkconfig/configurator.cpp
```diff
@@ -1198,13 +1238,30 @@ Configuration *Configurator::LoadConfiguration(const QString &path_to_configurat
         QJsonValue layer_value = layer_objects.value(layer_list[layer_index]);
         QJsonObject layer_object = layer_value.toObject();
 
-        // To match the layer we just need the name, paths are not
```
richard-lunarg, line 1201: This was a change so that VK_LAYER_PATH and user defined paths can change where the layers are pulled from. The layer paths are no longer encoded in the json files.

PR #1029, opened 2020-07-17
vkconfig: Catching up issues for last beta

richard-lunarg:
Linux versions now use a lunarg-vkconfig folder for app specific files
Can now delete VUID item when only one remains #418
Improvements to VUID filter control #400
User is notified on exit of override status #418
Added link to Vulkand spec and layers# 367
Cleanup app selection/launcher dialog #381

Change-Id: I1dbbfeec26bc5f1cf5b041990061d07c84e2a730

PR #1030, opened 2020-07-17
[WIP] vkconfig: Make version 1.2.145 a vkconfig compatibility break

christophe-lunarg:
Manage compatibility between version. Version 1.2.145 is a compatibility break which force a re-initialization. Future version won't break compatibility until we decide it break it.

Change-Id: I8eb20b5aefd14677c1e732c5f91c8c8064c8f69c


christophe-lunarg:
Part of #1029

PR #1032, opened 2020-07-17
vkconfig: Allow Linux build w/Qt5.5.1

richard-lunarg:
Change-Id: Iec43489e3bba5f8ed513bffb61ac438c60bf5c97
Should be ready to go for Qt 5.5.1 on Linux.


johnzupin:
Lower the required qt version in vkconfig/CMakeList.txt?

richard-lunarg:
CMakeLists.txt updated as well. Forgot about that!

johnzupin:
Was able to build on 16.04 and run vkconfig on 20.04!

PR #1033, opened 2020-07-17
vkconfig: Fixed linux directory regression

richard-lunarg:
Change-Id: I6851221ec11d693f126ee8cbaaa69af2aea912db
Fixes bad regression with Linux paths for storage of configuration files.

### DIFF
--- a/vkconfig/configurator.cpp
+++ b/vkconfig/configurator.cpp
@@ -240,8 +240,8 @@ Configurator::Configurator()
     }
 
     home = QDir::home();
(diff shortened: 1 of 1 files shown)


richard-lunarg:
The only thing that goes in lunarg-vkconfig is the configuration files. Basically, vkConfig's internal working files. None of the layers or anything else is stored there, and the override files go in the same place. Karl raised the point that we were cluttering up the parent folder, and we should tuck them away a little better. We already do on Windows, so this just cleans up Linux/Mac a little better.

PR #1034, opened 2020-07-20
vkconfig: Fix cmake regression for macOS builds

richard-lunarg:
Change-Id: I4f5f42f9d5d79b0075516533bf15282ccff54020
Fixed regression for cmake on macOS.


richard-lunarg:
Travis builds have succeeded. Somehow the status isn't being mirrored here. Change is to macOS cmake file anyway, so no real code changes.

PR #1035, opened 2020-07-20
Richard vkconfig single instance

richard-lunarg:
Added new utility class so that only one instance of vkconfig can be run at a time.

### DIFF
--- a/vkconfig/appsingleton.h
+++ b/vkconfig/appsingleton.h
@@ -26,8 +26,7 @@
  * - Richard S. Wright Jr. <richard@lunarg.com>
  */
 
-#ifndef APPSINGLETON_H
-#define APPSINGLETON_H
+#pragma once
 
 #include <QtNetwork/QLocalServer>
 #include <QtNetwork/QLocalSocket>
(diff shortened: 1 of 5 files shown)


vkconfig/appsingleton.h
```diff
@@ -0,0 +1,46 @@
...
+ * is not freed if the app crashes. The trade off here is a slight
+ * startup delay (500ms by default) to check if a local server is already
+ * running. This is 100% robust from an application crash on all
+ * platforms.
+ *
+ * Authors:
+ * - Richard S. Wright Jr. <richard@lunarg.com>
+ */
+
+#ifndef APPSINGLETON_H
```
christophe-lunarg, line 29: lunarg code uses #pragram once

richard-lunarg: The Google code guidelines specifically say not to use #pragma once. However, inspecting some other modules, I see that this is indeed the convention. I went ahead and changed this before the merge.

richard-lunarg:
Cosmetic change only. Not expected to build on Android either.

PR #1036, opened 2020-07-20
vkconfig: Fix for issue 421

richard-lunarg:
Change-Id: I01168d451d60245f5753470c5dbcfe579096ad48
If a configuration uses a layer that is missing, it is disabled. The same goes if a configuration has no layers at all. However, a valid use case is to have no layers listed, but to have some layers blacklisted (in other words, a configuration that explicitly turns off a validation layer).
As of this PR, this is now allowed/accounted for.


christophe-lunarg:
"vkconfig: Fix for issue 421" not a good title to browser the closed PR in the future and try to find anything

richard-lunarg:
>"vkconfig: Fix for issue 421" not a good title to browser the closed PR in the future and try to find anything
Agreed.

PR #1037, opened 2020-07-20
vkconfig: Fixes for issue 413

richard-lunarg:
Change-Id: I8d232375806d13739184e2f144afa2fd73d34cb5
This build fixes a situation where changing the log file may not be taken into account.
You can press the enter key in the VUID editor to add a selected message, and the edit field is cleared.
WIP: Removing messages that have been added and re-adding when removed.


christophe-lunarg:
"vkconfig: Fixes for issue 413" not a good title to browser the closed PR in the future and try to find anything

PR #1043, opened 2020-07-22
vkconfig: Beta 4 updates

richard-lunarg:
Change-Id: I023ae6f9654de74b69b66b3b0fb2ac1581d7e067
Cleanup for beta 4
Class member variables are prefixed with _ rather than suffix (discussed with larger LunarG team)
Removed unused test tab from vkvia output dialog.
Settings tree properly saved/restored after configuration edit operations and on program shutdown/restart.
VUID fixes after regression. Still a WIP, but better and works correctly for beta 4.

### DIFF
--- a/vkconfig/configuration.h
+++ b/vkconfig/configuration.h
@@ -48,17 +48,17 @@ class Configuration {
     Configuration();
     ~Configuration();
 
-    QString name;                   // User readable display of the profile name (may contain spaces)
-                                    // This is the same as the filename, but with the .json stripped off.
-    QString file;                   // Root file name without path (by convention, no spaces and .profile suffix)
-    QString description;            // A friendly description of what this profile does
-    QByteArray setting_tree_state;  // Recall editor tree state
-    ValidationPreset preset;        // Khronos layer presets. 0 = none or user defined
+    QString _name;                   // User readable display of the profile name (may contain spaces)
+                                     // This is the same as the filename, but with the .json stripped off.
+    QString _file;                   // Root file name without path (by convention, no spaces and .profile suffix)
+    QString _description;            // A friendly description of what this profile does
+    QByteArray _setting_tree_state;  // Recall editor tree state
+    ValidationPreset _preset;        // Khronos layer presets. 0 = none or user defined
 
     // A configuration is nothing but a list of layers and their settings in truth
-    QVector<LayerFile *> layers;
+    QVector<LayerFile *> _layers;
 
-    QStringList excluded_layers;  // Just the names of blacklisted layers
+    QStringList _excluded_layers;  // Just the names of blacklisted layers
 
     LayerFile *FindLayer(const QString &layer_name, const QString &full_path) const;  // Find the layer if it exists
     LayerFile *FindLayerNamed(const QString &layer_name) const;  // Find the layer if it exists, only care about the name
@@ -69,7 +69,7 @@ class Configuration {
 
     LayerFile *GetKhronosLayer();  // Retrieve the Khronos validation layer if it is included
 
-    bool IsValid() { return all_layers_available; }
+    bool IsValid() { return _all_layers_available; }
 
-    bool all_layers_available;
+    bool _all_layers_available;
 };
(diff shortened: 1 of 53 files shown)


richard-lunarg:
Also fixed when closing, don't clear the settings tree if the user decides not to close.
Removed some unused signal/slot connections to eliminate runtime warnings.
These changes were made after beta4 was posted.

vkconfig/mainwindow.cpp
```diff
@@ -759,18 +800,22 @@ void MainWindow::DuplicateClicked(ConfigurationListItem *item) {
...
     (void)item;  // We don't need this
-
+    PushCurrentItem();
     Configurator &configurator = Configurator::Get();
 
     QString full_suggested_path = configurator.GetPath(Configurator::LastImportPath);
     QString full_import_path =
         QFileDialog::getOpenFileName(this, "Import Layers Configuration File", full_suggested_path, "*.json");
-    if (full_import_path.isEmpty()) return;
+    {
```
lunarpapillo, line 809: Why is this couple of lines in its own block?

vkconfig/vuidsearchwidget.cpp
```diff
@@ -19,72 +19,80 @@
...
+    _user_box->installEventFilter(this);
 
-    search_vuid_ = nullptr;  // Safe to delete a pointer
+    _search_vuid = nullptr;  // Safe to delete a pointer
     ResetCompleter();
 
-    connect(add_button_, SIGNAL(pressed()), this, SLOT(addButtonPressed()));
-    connect(user_box_, SIGNAL(returnPressed()), this, SLOT(addButtonPressed()));
+    connect(_add_button, SIGNAL(pressed()), this, SLOT(addButtonPressed()));
+    // connect(_user_box, SIGNAL(returnPressed()), this, SLOT(addButtonPressed()));
```
lunarpapillo, line 48: Is this an intentional removal?

If so, do we need to keep the commented-out code?

vkconfig/mainwindow.cpp
```diff
@@ -632,55 +636,81 @@ void MainWindow::on_pushButtonAppList_clicked() {
...
+}
+
+///////////////////////////////////////////////////////////////
+// When changes are made to the layer list, it forces a reload
+// of the configuration list. This wipes everything out, so we
+// need a way to restore the currently selected item whenever
+// certain kinds of edits occur. These push/pop functions
+// accomplish that. If nothing can be found it should simply
+// leave nothing selected.
+ConfigurationListItem *MainWindow::PushCurrentItem(void) {
```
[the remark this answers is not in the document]
richard-lunarg: Originally, it was a stack... I renamed to SaveLastItem, and RestoreLastItem, with an override. They are always used in pairs, and sometimes across slots (a Qt message system), basically across function calls, so the saved value needs to be somewhere that will still be in scope, thus the class membere. This was a good critique.

vkconfig/vuidsearchwidget.cpp
richard-lunarg: Removed for testing. made permanent... forgot to delete. Another good catch, and further justification for not cramming too much into one PR.

vkconfig/mainwindow.cpp
richard-lunarg: Um... too much caffeine? Looks like refactoring gone bad here. Fixed for final PR.

vkconfig/configuration.h
```diff
@@ -48,17 +48,17 @@ class Configuration {
     Configuration();
     ~Configuration();
 
-    QString name;                   // User readable display of the profile name (may contain spaces)
-                                    // This is the same as the filename, but with the .json stripped off.
-    QString file;                   // Root file name without path (by convention, no spaces and .profile suffix)
-    QString description;            // A friendly description of what this profile does
-    QByteArray setting_tree_state;  // Recall editor tree state
-    ValidationPreset preset;        // Khronos layer presets. 0 = none or user defined
+    QString _name;                   // User readable display of the profile name (may contain spaces)
```
christophe-lunarg, line 51: I think, it's only private members that uses "_"

PR #1051, opened 2020-07-24
vkconfig: Final touches for SDK release

christophe-lunarg:
- Fix link to online documentation
- Layer property text is no longer cropped
- Delete debug code left over
- Fix remove custom path assert
- Clean up context menu
- Add VK_LAYER_PATH to the Vulkan status on start
- More consistent custom path behavior: no message box to remove
- More consistent naming accross the tool

Change-Id: I81fba4a45663345538eeb2d898039f9fb709a332

### DIFF
--- a/vkconfig/configurator.cpp
+++ b/vkconfig/configurator.cpp
@@ -31,7 +31,6 @@
 #include <QLibrary>
 #include <QMessageBox>
 #include <QCheckBox>
-#include <QDebug>
 
 #ifdef _WIN32
 #include <shlobj.h>
@@ -386,6 +385,9 @@ QString Configurator::CheckVulkanSetup() const {
         return log;
     }
 
+    // if (!VK_LAYER_PATH.isEmpty())
+    log += "- Using Layers from VK_LAYER_PATH\n";
+
     // Check layer paths
     if (_custom_layers_paths.count() > 0) {
         log += "- Custom Layers Paths:\n";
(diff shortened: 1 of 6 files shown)


vkconfig/mainwindow.cpp
```diff
@@ -78,9 +78,6 @@ MainWindow::MainWindow(QWidget *parent) : QMainWindow(parent), ui(new Ui::MainWi
     ///////////////////////////////////////////////
     Configurator &configurator = Configurator::Get();
 
-    if (!configurator.VK_LAYER_PATH.isEmpty())
-        ui->groupBoxProfiles->setTitle("Vulkan Layers Configurations With VK_LAYER_PATH Precedence");
```
richard-lunarg, line 82: In a previous review, Karl had said he thought this needed to be much more "in your face" when the VK_LAYER_PATH is in effect. I see you added it to the log, and that's good, I think it should go there too. Removing it from here might get some blow back. I don't strictly object to it's removal, but we did have some feedback that it needed to be more obvious to the user when in use, and this is one way to do that.

vkconfig/configurator.cpp
```diff
@@ -386,6 +385,9 @@ QString Configurator::CheckVulkanSetup() const {
         return log;
     }
 
+    // if (!VK_LAYER_PATH.isEmpty())
+    log += "- Using Layers from VK_LAYER_PATH\n";
```
richard-lunarg, line 389: Good addition. I should have done this too.

PR #1059, opened 2020-07-28
Vkconfig adhock fixes

richard-lunarg:
Fixes for Linux issues that arose during SDK testing.

### DIFF
--- a/vkconfig/configurator.cpp
+++ b/vkconfig/configurator.cpp
@@ -841,7 +841,7 @@ void Configurator::FindVkCube() {
     // put the log in the home folder.
 #ifdef __linux__
     QDir dir = QDir::homePath();
-    application->log_file = QDir::homePath() + "/vkcube";
+    application->log_file = QDir::homePath() + "/vkcube.txt";
 #endif
 
     _overridden_application_list.push_back(application);
--- a/vkconfig/configurator.cpp
+++ b/vkconfig/configurator.cpp
@@ -953,11 +953,11 @@ void Configurator::LoadAllInstalledLayers() {
     if (lp != 0)
         for (int i = 0; i < lp; i++) LoadLayersFromPath(VK_LAYER_PATH[i], _available_Layers);
 
-    // SECOND: Standard layer paths, in standard locations
-    for (std::size_t i = 0, n = vku::countof(szSearchPaths); i < n; i++) LoadLayersFromPath(szSearchPaths[i], _available_Layers);
-
-    // THIRD: Any custom paths? Search for those too
+    // SECOND: Any custom paths? Search for those too
     for (int i = 0; i < _custom_layers_paths.size(); i++) LoadLayersFromPath(_custom_layers_paths[i], _available_Layers);
+
+    // THIRD: Standard layer paths, in standard locations. The above has always taken precidence.
+    for (std::size_t i = 0, n = vku::countof(szSearchPaths); i < n; i++) LoadLayersFromPath(szSearchPaths[i], _available_Layers);
 
     // FOURTH: Finally, see if thee is anyting in the VULKAN_SDK path that wasn't already found elsewhere
     QString vulkanSDK = qgetenv("VULKAN_SDK");
--- a/vkconfig/configurator.cpp
+++ b/vkconfig/configurator.cpp
@@ -836,6 +836,14 @@ void Configurator::FindVkCube() {
     }
 
     Application *application = new Application(local.absoluteFilePath(), "--suppress_popups");
+
+    // On Linux, vkcube may be in /usr/bin, which is not a directory we can write into. In that case,
+    // put the log in the home folder.
+#ifdef __linux__
+    QDir dir = QDir::homePath();
+    application->log_file = QDir::homePath() + "/vkcube";
+#endif
+
     _overridden_application_list.push_back(application);
 }
 
--- a/vkconfig/configurator.cpp
+++ b/vkconfig/configurator.cpp
@@ -964,7 +964,7 @@ void Configurator::LoadAllInstalledLayers() {
     // SECOND: Any custom paths? Search for those too
     for (int i = 0; i < _custom_layers_paths.size(); i++) LoadLayersFromPath(_custom_layers_paths[i], _available_Layers);
 
-    // THIRD: Standard layer paths, in standard locations. The above has always taken precidence.
+    // THIRD: Standard layer paths, in standard locations. The above has always taken precedence.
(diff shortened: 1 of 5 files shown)


vkconfig/configurator.cpp
```diff
@@ -953,12 +961,12 @@ void Configurator::LoadAllInstalledLayers() {
     if (lp != 0)
         for (int i = 0; i < lp; i++) LoadLayersFromPath(VK_LAYER_PATH[i], _available_Layers);
 
-    // SECOND: Standard layer paths, in standard locations
-    for (std::size_t i = 0, n = vku::countof(szSearchPaths); i < n; i++) LoadLayersFromPath(szSearchPaths[i], _available_Layers);
-
-    // THIRD: Any custom paths? Search for those too
+    // SECOND: Any custom paths? Search for those too
     for (int i = 0; i < _custom_layers_paths.size(); i++) LoadLayersFromPath(_custom_layers_paths[i], _available_Layers);
```
charles-lunarg, line 965: nit: could use a range-for loop here
```cpp
for (int i = 0; i < _custom_layers_paths.size(); i++) LoadLayersFromPath(_custom_layers_paths[i], _available_Layers);
```
to
```cpp
for (auto& path : custom_layers_paths) LoadLayersFromPath(path, _available_Layers);
```

vkconfig/mainwindow.cpp
```diff
@@ -1127,20 +1130,19 @@ bool MainWindow::eventFilter(QObject *target, QEvent *event) {
         }
     }
 
-    // If not configuration active, disable the setting area
+    // If no configuration active, disable the setting area
     // ui->groupBoxEditor->setEnabled(SaveLastItem() != nullptr);
```
charles-lunarg, line 1134: Dead code?

richard-lunarg: Dang it. Yes, dead code. Copy and paste error from Christophe I think. Removed and recommited.

vkconfig/configurator.cpp
```diff
@@ -953,12 +961,12 @@ void Configurator::LoadAllInstalledLayers() {
...
 
+    // THIRD: Standard layer paths, in standard locations. The above has always taken precidence.
```
lunarpapillo, line 967: precidence => precedence

vkconfig/dlgcreateassociation.cpp
```diff
@@ -37,7 +37,12 @@ dlgCreateAssociation::dlgCreateAssociation(QWidget *parent)
     Configurator &configurator = Configurator::Get();
     configurator._override_application_list_updated = false;
 
-    if (!configurator.HasActiveOverrideOnApplicationListOnly()) setWindowTitle("Applications Launcher Shortcuts");
+    if (!configurator.HasActiveOverrideOnApplicationListOnly())
```
lunarpapillo, line 40: The two branches of the `if` aren't parallel at all.  I'd think that if `ui->treeWidget->setHeaderHidden(false)` had to be called in one branch, that `setHeaderHidden(true)` would have to be called in the other, etc.

Please double-check that this does what you think it does.  If you're convinced it does, consider a comment explaining why the two branches are not parallel.

vkconfig/dlgcreateassociation.cpp
```diff
@@ -144,8 +149,7 @@ QTreeWidgetItem *dlgCreateAssociation::CreateApplicationItem(const Application &
     ui->treeWidget->addTopLevelItem(item);
 
     if (configurator.HasActiveOverrideOnApplicationListOnly()) {
-        item->setText(0, tr("    ") + application.executable_path);
-        QCheckBox *check_box = new QCheckBox("");
+        QCheckBox *check_box = new QCheckBox(application.executable_path);
```
lunarpapillo, line 152: This looks to be related to the above... 

I don't know Qt well enough to understand the use of this structure, so I'll just ask: what happens if we repeatedly select and unselect whatever provides `HasActiveOverrideOnApplicationListOnly`?  Will we keep allocating new checkboxes each time?  Will the old checkboxes be cleaned up properly?

I'd be more comfortable with a static UI configuration, with elements that are made visible and invisible as needed, than I am with dynamically reconfiguring the UI repeatedly.

vkconfig/mainwindow.cpp
```diff
@@ -731,12 +731,15 @@ void MainWindow::NewClicked() {
 void MainWindow::addCustomPaths() {
     SaveLastItem();
 
-    dlgCustomPaths dlg(this);
-    dlg.exec();
-
     // Get the tree state and clear it.
+    // This looks better athetically after the dialog
```
lunarpapillo, line 735: athetically => aesthetically

(I think - maybe you mean "athletically" ;-)

vkconfig/mainwindow.ui
```diff
@@ -100,119 +100,127 @@
...
-             <font>
-              <family>Arial</family>
-              <pointsize>10</pointsize>
-             </font>
-            </property>
-            <property name="text">
-             <string>Edit...</string>
-            </property>
-           </widget>
+           <layout class="QGridLayout" name="gridLayout_7">
```
lunarpapillo, line 103: all of these changes are due to the addition of a wrapping `QGridLayout`, yes?  I didn't scrutinize the details since they seem to be changes in whitespace only.  (If there was another substantive change, point me at it and I'll look.)

vkconfig/configurator.cpp
richard-lunarg: Details always matter... ;)

vkconfig/dlgcreateassociation.cpp
richard-lunarg: The default state is hidden, as set by the .ui file. I did look twice though, so it warrants a comment.

vkconfig/dlgcreateassociation.cpp
richard-lunarg: These are only created when items are added to the list. We don't delete/re-add them when they are clicked/unclicked. Once added to the tree, they are owned and destroyed by the tree as well when it goes away. If the user deletes a row, that also destroys the tree. This is standard Qt design, all Widgets cleanup widgets that belong to them. Even if you replace the widget, the original widget get's destroyed by the set method for you so you don't have to keep a pointer around and do it yourself.

vkconfig/mainwindow.cpp
richard-lunarg: aesthetically
You are an eagle eyed man.

vkconfig/mainwindow.ui
richard-lunarg: Yes, a new layout had to be added for spacing reasons. Layouts are wonderful, it's taken me a while to master their use though.

PR #1060, opened 2020-07-29
vkconfig: Warnings, misspellings, singleton fix

richard-lunarg:
Change-Id: I960be1345af8562be34a881238dba897bef17737
Spelling errors, etc. Biggest change is to the app singleton feature.
Also, checked for file not existing, as failure was not as silent as thought; prevents warning output to console.
Less "scary" text when $VULKAN_SDK is not set.


richard-lunarg:
Cherry picked to SDK branch.

PR #1064, opened 2020-08-03
vkconfig: API Dump is now the default configuration

richard-lunarg:
Change-Id: I0ecda5c374a7a21b41a1ef936dd2bfdfb55442ef
API Dump is the default layer configuration
Log file browser no longer blanks edit control on cancel.
Changed description of the AP dump configuration.


christophe-lunarg:
What's the reason for this?

richard-lunarg:
API dump's description was deemed too "colloquial".
API dump was made the default to ensure layer output could be verified on first use of vkconfig.

PR #1070, opened 2020-08-06
vkconfig post sdk fixes

richard-lunarg:
Corrected spacing issue described here:
https://github.com/LunarG/Projects/issues/448
This required changing the two column approach in the main config tree to a one column approach. Better spacing behavior on all platforms.

Debug-printf configuration was actually identical to GPU-Assisted. Changed default to use debug-printf instead for this config.

Fixed issue where a MessageBox was stalling code flow when a log file could not be opened. Meanwhile, log messages came through in the background and tried writing to the unopened log file.

This PR additionally removes the editor state field from the .json's which should not have been in the template in the first place.
A message box that warns the user is displayed if this is the first post 1.2.148 install that the default configurations must be removed and user configs recreated. This is not displayed if it is a true first time run of Vulkan Configurator.

### DIFF
--- a/vkconfig/main.cpp
+++ b/vkconfig/main.cpp
@@ -70,7 +70,7 @@ int main(int argc, char* argv[]) {
                 return -1;
             }
 
-            // SDK release 1.2.148.5 and earlier had some bad tokens in the default configurations and this was the only way to
+            // SDK release 1.2.148 and earlier had some bad tokens in the default configurations and this was the only way to
             // "repair" them. In the future, we will be more surgical about this when it occurs, or when upates
             // are made to support additional validation layer settings. The initial first run key was "Initialized", which
             // set to false also was non-intuitive (it was backwards), so good to change that as well. The logic below will
--- a/vkconfig/main.cpp
+++ b/vkconfig/main.cpp
@@ -74,7 +74,7 @@ int main(int argc, char* argv[]) {
             // "repair" them. In the future, we will be more surgical about this when it occurs, or when upates
             // are made to support additional validation layer settings. The initial first run key was "Initialized", which
             // set to false also was non-intuitive (it was backwards), so good to change that as well. The logic below will
-            // not show this dialog to everyone, only those who installed the SDK on the first week of release.
+            // not show this dialog to everyone, only those who installed the SDK that was posted the first week of release.
             QSettings settings;
             if (!settings.value("Initialized", true).toBool() && settings.value(VKCONFIG_KEY_INITIALIZE_FILES, true).toBool()) {
                 QMessageBox alert;
--- a/vkconfig/main.cpp
+++ b/vkconfig/main.cpp
@@ -70,6 +70,24 @@ int main(int argc, char* argv[]) {
                 return -1;
             }
 
+            // SDK release 1.2.148.5 and earlier had some bad tokens in the default configurations and this was the only way to
+            // "repair" them. In the future, we will be more surgical about this when it occurs, or when upates
+            // are made to support additional validation layer settings. The initial first run key was "Initialized", which
+            // set to false also was non-intuitive (it was backwards), so good to change that as well. The logic below will
+            // not show this dialog to everyone, only those who installed the SDK on the first week of release.
+            QSettings settings;
+            if (!settings.value("Initialized", true).toBool() && settings.value(VKCONFIG_KEY_INITIALIZE_FILES, true).toBool()) {
+                QMessageBox alert;
+                alert.setText(
+                    "Vulkan Configurator needs to be reset. "
+                    "All standard configurations are being reset to their default state, and user "
+                    "configurations need to be recreated. Future releases will allow for repair of faulty configuration files.");
+                alert.setWindowTitle("Vulkan Configurator");
+                alert.setIcon(QMessageBox::Warning);
+                alert.exec();
+                // DO NOT set VKCONFIG_KEY_INITIALIZED_FILES to false here
+            }
+
             // We simply cannot run without any layers
             if (Configurator::Get().InitializeConfigurator() == false) return -1;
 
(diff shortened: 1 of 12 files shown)


vkconfig/resourcefiles/Validation - Standard.json
```diff
@@ -7,15 +7,15 @@
         "layer_options": {
             "VK_LAYER_KHRONOS_validation": {
                 "debug_action": {
-                    "default": "VK_DBG_LAYER_ACTION_LOG_MSG",
+                    "default": "VK_DBG_LAYER_ACTION_DEBUG_OUTPUT",
```
charles-lunarg, line 10: This shouldn't be changed at all, the Log Message should be the default.

vkconfig/resourcefiles/Validation - Shader Printf.json
```diff
@@ -46,11 +46,10 @@
...
                 "enables": {
-                    "default": [
-                        "VK_VALIDATION_FEATURE_ENABLE_GPU_ASSISTED_EXT",
-                        "VK_VALIDATION_FEATURE_ENABLE_GPU_ASSISTED_RESERVE_BINDING_SLOT_EXT"
-                    ],
-                    "description": "Setting an option here will enable specialized areas of validation",
+                "default": [
+                     "VK_VALIDATION_FEATURE_ENABLE_DEBUG_PRINTF_EXT"
+                     ],
+                "description": "Setting an option here will enable specialized areas of validation",
```
charles-lunarg, line 52: Is this description correct? Seems it shouldn't refer to validation at all, but just the debug-printf option.

vkconfig/mainwindow.cpp
```diff
@@ -378,7 +377,7 @@ void MainWindow::toolsResetToDefault(bool checked) {
     _settings_tree_manager.CleanupGUI();
     configurator.LoadAllConfigurations();
 
-    // Find the API dump and make it current if we are active
+    // Find the API dump  and make it current if we are active
```
charles-lunarg, line 380: Extra space.

vkconfig/resourcefiles/Validation - Standard.json
richard-lunarg: Yeah... how'd that happen... Second set of eyes is always good.

vkconfig/resourcefiles/Validation - Shader Printf.json
richard-lunarg: That description field is not for the current setting (debug printf), but the settings as a whole, of which there are 4 to choose from:
                     "VALIDATION_CHECK_ENABLE_VENDOR_SPECIFIC_ARM": "ARM specific validation",
                        "VK_VALIDATION_FEATURE_ENABLE_BEST_PRACTICES_EXT": "Best Practices warning checks",
                        "VK_VALIDATION_FEATURE_ENABLE_DEBUG_PRINTF_EXT": "Debug printf",
                        "VK_VALIDATION_FEATURE_ENABLE_GPU_ASSISTED_EXT": "GPU-Assisted Validation",
                        "VK_VALIDATION_FEATURE_ENABLE_GPU_ASSISTED_RESERVE_BINDING_SLOT_EXT": "Reserve a descriptorSet binding slot for internal use"

charles-lunarg: Ah I was mistaken then.

vkconfig/main.cpp
```diff
@@ -70,6 +70,24 @@ int main(int argc, char* argv[]) {
...
+            if (!settings.value("Initialized", true).toBool() && settings.value(VKCONFIG_KEY_INITIALIZE_FILES, true).toBool()) {
+                QMessageBox alert;
+                alert.setText(
+                    "Vulkan Configurator needs to be reset. "
+                    "All standard configurations are being reset to their default state, and user "
+                    "configurations need to be recreated. Future releases will allow for repair of faulty configuration files.");
+                alert.setWindowTitle("Vulkan Configurator");
+                alert.setIcon(QMessageBox::Warning);
+                alert.exec();
+                // DO NOT set VKCONFIG_KEY_INITIALIZED_FILES to false here
```
charles-lunarg, line 88: We need to be sure that vkconfig will not bring up this dialog box more than once. I am not a Qt expert, so I cannot simply look at the change and know, I will need to do a bit of manual building & testing to confirm it.

vkconfig/main.cpp
```diff
@@ -70,6 +70,24 @@ int main(int argc, char* argv[]) {
                 return -1;
             }
 
+            // SDK release 1.2.148.5 and earlier had some bad tokens in the default configurations and this was the only way to
```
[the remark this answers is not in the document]
richard-lunarg: On the web site, it shows 1.2.148.5... I just realized, that's probably the private SDK build as I'm logged in. So no... I'll change before I merge to master, but for the binary build it's okay.

vkconfig/main.cpp
richard-lunarg: It should only appear once. But yes, it needs to be tested thoroughly.

vkconfig/main.cpp
```diff
@@ -70,6 +70,24 @@ int main(int argc, char* argv[]) {
...
+            // "repair" them. In the future, we will be more surgical about this when it occurs, or when upates
+            // are made to support additional validation layer settings. The initial first run key was "Initialized", which
+            // set to false also was non-intuitive (it was backwards), so good to change that as well. The logic below will
+            // not show this dialog to everyone, only those who installed the SDK on the first week of release.
```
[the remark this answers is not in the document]
richard-lunarg: I tweaked this information.

christophe-lunarg:
"This PR additionally removes the editor state field from the .json's which should not have been in the template in the first place."

Why? Actually it was set on purpose to hide settings that the new user doesn't need to be expose to see use the tool. This is to avoid overflow of information on first use.

PR #1076, opened 2020-08-11
Vulkan Configurator: Default settings

christophe-lunarg:
- Bring back "Validation - Standard" as the default configuration
=> This is done by enabling "Info" messages to systematically getting something out of the validation layers

- Bring back editor settings states presets
=> This is to make sure that the users are exposed first to the most important settings: Presets, message filtering.

### DIFF
--- a/vkconfig/vkconfig.md
+++ b/vkconfig/vkconfig.md
@@ -149,7 +149,7 @@ We are working on defining layers development conventions to resolve this issue 
 --------------
 ## Known Issues
 
-- The UI still feels a little clunky... Need more polish.
+- UI still needs some refinement in some areas...
 - Message filtering using VUID name and index is not yet fully implemented.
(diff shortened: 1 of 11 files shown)


vkconfig/vkconfig.md
```diff
@@ -149,6 +149,8 @@ We are working on defining layers development conventions to resolve this issue
 --------------
 ## Known Issues
 
+- The UI still feels a little clunky... Need more polish.
```
richard-lunarg, line 152: This phrase in really bad I think in English. Clunky is a very derogatory and negative term, and you poison peoples expectations by using it. Don't tell them right away the user interface is horrible... tell them there are some issues we still want to clean up, and let them tell us what/if they find anything objectionable.

richard-lunarg: "UI still needs some refinement in some areas"
A much better spin...

PR #1077, opened 2020-08-11
apidump: Make options case-insensitive

### LINKED ISSUE #1075: case-sensitivity for boolean settings
Hello,

I was trying to get the `lunarg_api_dump.file` setting to work and was surprised to find that the setting is case-sensitive. From [the documentation](https://vulkan.lunarg.com/doc/sdk/1.2.148.0/windows/api_dump_layer.html), I would have assumed that I could use `true` instead of `TRUE`. From looking at the code, it appears that all the boolean settings are case-sensitive.

Would it make sense to update the docs to only use `TRUE` and `FALSE` or, alternatively, update the code to parse `true` and `false` options in a case-insensitive manner?

Thanks!

charles-lunarg:
Options previously would only work if they were capitalized,
causing confusion when it wasn't obvious this was required. This
commit relaxes that constraint to allow mixed case and lower case
text options, notably bool and file output type.

Fixes #1075 

Change-Id: I984408469e84f5773266cabc8bd08d4fb915a72b

### DIFF
--- a/layersvt/api_dump.h
+++ b/layersvt/api_dump.h
@@ -230,12 +230,15 @@ class ApiDumpSettings {
         // If the layer settings file has a flag indicating to output to a file,
         // do so, to the appropriate filename.
         const char *file_option = getLayerOption("lunarg_api_dump.file");
-        if (file_option != NULL && strcmp(file_option, "TRUE") == 0) {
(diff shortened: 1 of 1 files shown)


christophe-lunarg:
Perfect, I encounted this issue myself!

PR #1078, opened 2020-08-12
[WIP] Vulkan Configurator: Show version in title bar

christophe-lunarg:
Change-Id: I7b9432ad8b256329dc3735069e91dad8d848473d


richard-lunarg:
I think it might be better if vkconfig had it's own version scheme rather than being dependent on the Vulkan SDK version it was built against. It could get very confusing to have a new set of source code built against an older Vulkan header. The app identifies as an old version, but has features of the new version. Or, old source code built against a newer Vulkan Header would similarly be misleading. I have no objection to putting this in the Window caption instead of the about box, just where it's coming from. Every time we intend to do a merge to master, we'd then bump the build number or version number (Version 1.0.001 - or something like that). Also, we could update the version number only on SDK releases, but then have a build number separate for intermediate steps. The reason this came up, was the alpha testers could not tell if they had a new version for not. All were built agains the same SDK, so this change would not have helped in that situation either. Identifying which version of the SDK it was built agains would also be a useful piece of information. Probably too crowded for the caption, so... maybe back to the about box again.

christophe-lunarg:
Vulkan Configurator is very tied to a specific version of the SDK (the configurations have settings for that version) and Vulkan Configurator is the front facing tool of the SDK. Hence, using a separated versioning scheming is plain dangerous and subject to trouble such as user not knowing what version of Vulkan Configurator to enable new features.

This said, we can add an implementation number like is done for the layers, particularly taken advantage by gfxRecon.

richard-lunarg:
Okay, very true. I think then just adding a build number to the about box would likely meet both needs then. The housekeeping detail for that then is how often do we update the build number, and is there a nice automated way to do so without having to edit a header file every time we do a merge.

PR #1080, opened 2020-08-13
DevSim-layer-test-Android

johnFJ:
-test


Tony-LunarG:
Why do you want to change the test's exit code?

PR #1100, opened 2020-08-19
vkconfig: Initial unit test using Google Test

christophe-lunarg:
- Integrate Google Test to cmake build system
- Initial unit test for version, command line and util
- Move code to vkconfig_core to make a static library
necessary for Google Test to link to each test program.
Down the line this will help clearly separating core and GUI code
- Add aapveyor test: https://ci.appveyor.com/project/karl-lunarg/vulkantools/builds/34742562/job/m00xlbua2btfvdfo
- Usage:
cmake --build . -j 16 --config Debug
cmake --build . -j 16 --config Release
ctest -j 16 --output-on-failure -C Debug
ctest -j 16 --output-on-failure -C Release

Change-Id: I89a92ddfeca5034cab216ca798557a663df96543


vkconfig/mainwindow.cpp
```diff
@@ -186,7 +186,9 @@ void MainWindow::LoadConfigurationList() {
         item->configuration = configurator._available_configurations[i];
         ui->profileTree->addTopLevelItem(item);
         item->radio_button = new QRadioButton();
+#ifndef _WIN32 // This is a hack to workaround a Linux layout issue
         item->radio_button->setText("-");
```
richard-lunarg, line 190: Removing the - is not as simple as this. We still need some spaces, either in the radio button text, or prepended to the description in the next column. Roll this back, and I'll make the changes and test it on all platforms under a separate commit.

vkconfig/configurator.cpp
```diff
@@ -1560,21 +1561,21 @@ void Configurator::SetActiveConfiguration(Configuration *configuration) {
     // Loop through all the layers
     for (int layer_index = 0; layer_index < configuration->_layers.size(); layer_index++) {
         LayerFile *layer_file = configuration->_layers[layer_index];
-        stream << endl;
-        stream << "# " << layer_file->_name << endl;
```
richard-lunarg, line 1564: Curious why you replaced the endl with "\n"?

christophe-lunarg: On my machin using QtCreator, I have an error on this.

vkconfig/mainwindow.cpp
christophe-lunarg: I am not seeing any layout issue on Windows with my change but feel free to improve my change with a future PR. 

image

richard-lunarg: Yeah, looks fine on Windows... Okay, go ahead and do this and I'll fix it after the fact for the other platforms.

vkconfig/configurator.cpp
richard-lunarg: Hmmm. Odd, but plenty good enough reason.

Issue #1098, opened 2020-08-18
vkconfig: Show json files in folder with folder select dialog


richard-lunarg:
From @charles-lunarg 
Would be nice to have the file select dialog show .jsons as well as folders, making it easier to know if you are in the right folder or not.

Counter argument would be that showing the .json files makes it seem like the user should select one of those layers and only one of the layers would be added, even if it its mostly explicit that its looking for paths.

I appreciate the list of available layers being shown after you've confirmed the folder, but best to have the information available as soon as possible..

richard-lunarg:
This is non-standard, but may be possible.

richard-lunarg:
Addressed in PR #1101

PR #1101, opened 2020-08-19
vkconfig gui fixes and updates

richard-lunarg:
Fixed an unreported regression that made it impossible to remove a custom layer path.
Added the ability to see files in a folder when browsing for custom layer paths. This uses a non-OS supplied file browser dialog. (See issue #1098 )
WIP, possibly done: The appearance of the radio buttons for the configuration listing (radio button selects the configuration as the active one). It turns out the visual anomalies originally worked around originated on the Mac. We've gone back and forth a few times, but it seems the Mac fix was the actual problem, and then we fixed the fix... so to speak. No additional spacing seems to be required on Window, Ubuntu OR Fedora... only on macOS. The original macOS work around seems to really wreak havok on Fedora, but not having the work around actually looks better everywhere EXCEPT macOS. So I think the final solution is to put some #defines around it, and account for the issue only on macOS.

### DIFF
--- a/vkconfig/configurator.cpp
+++ b/vkconfig/configurator.cpp
@@ -1049,9 +1049,9 @@ void Configurator::LoadLayersFromPath(const QString &path, QVector<LayerFile *> 
 //////////////////////////////////////////////////////////////////////////
 // Populate a tree widget with the custom layer paths and the layers that
 // are being used in them.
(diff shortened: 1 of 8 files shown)


christophe-lunarg (approved):
Looks good to me. I pushed a trivial change to relaunch Github C.I. which I expect failed for no good reason / platfom instability

Issue #1091, opened 2020-08-18
vkconfig: Need a way to edit and display configuration descriptions


richard-lunarg:
The description field of a configuration can be seen in the GUI as a "tool tip" by hovering over the configuration; there is no way in the GUI to change it (either for a default configuration or a user-specified configuration).  There should be.  (In the worst case, when an existing configuration is used as the basis of a new user configuration, the description field is inherited, which would leave the user configuration with an incorrect tool tip/description.  (A workaround might be to export a configuration as a JSON, change the description field in the JSON, and then import.)

PR #1105, opened 2020-08-24
vkconfig: .pro file fix and description editiong

richard-lunarg:
Change-Id: I4a32b7fd8998a69225324057442ce8e30f089097
This addresses the issue #1091 and fixes a regression in the .pro file (CI builds with cmake, and the .pro file was broken for QtCreator builds.
Basic description editing was added to the layer editor dialog, as this was where it was originally. This can be iterated and/moved if needed, but seemed a good place to put this to me.


christophe-lunarg:
The description field looks great this way!

PR #1112, opened 2020-08-28
vkconfig: Improve default applications handling 

christophe-lunarg:
- If the executable doesn't exist, try to replace it.
- Handle the case where the user could include multiple versions of vkcube.
- First use the relative path to vkconfig to use the one from the SDK.
- Add vkcubepp.
- When the list is empty, add default applications.

### DIFF
--- a/vkconfig/configurator.cpp
+++ b/vkconfig/configurator.cpp
@@ -757,57 +757,99 @@ const QString &Configurator::GetCustomLayersPath(int path_index) const {
 void Configurator::SelectLaunchApplication(int application_index) {
     if (application_index < 0) return;
 
-    _active_launch_executable_path = _overridden_application_list[application_index]->executable_path;
+    _active_launch_executable_path = _overridden_applications[application_index]->executable_path;
 }
 
 int Configurator::GetLaunchApplicationIndex() const {
-    for (int i = 0; i < _overridden_application_list.size(); i++) {
-        if (_overridden_application_list[i]->executable_path == _active_launch_executable_path) return i;
+    for (int i = 0; i < _overridden_applications.size(); i++) {
+        if (_overridden_applications[i]->executable_path == _active_launch_executable_path) return i;
     }
 
     return 0;  // Not found, but the list is present, so return the first item.
 }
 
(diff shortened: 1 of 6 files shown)


vkconfig/configurator.cpp
```diff
@@ -757,57 +757,99 @@ const QString &Configurator::GetCustomLayersPath(int path_index) const {
...
-        if (_overridden_application_list[i]->executable_path == _active_launch_executable_path) return i;
+    for (int i = 0; i < _overridden_applications.size(); i++) {
+        if (_overridden_applications[i]->executable_path == _active_launch_executable_path) return i;
     }
 
     return 0;  // Not found, but the list is present, so return the first item.
 }
 
+static QString GetDefaultExecutablePath(const QString &executable_name) {
+#if defined(__APPLE__)
```
richard-lunarg, line 772: You may run into merge problems here as the last PR replaced these #def's.

vkconfig/configurator.cpp
```diff
@@ -757,57 +757,99 @@ const QString &Configurator::GetCustomLayersPath(int path_index) const {
...
+    static const char *DEFAULT_PATH = "/../..";
+#elif defined(__linux__) || defined(_WIN32)
+    static const char *DEFAULT_PATH = "";
+#else
+#error "Unknown platform"
```
richard-lunarg, line 777: You don't need this in code body. The header will catch this and halt compilation.

vkconfig/configurator.cpp
christophe-lunarg: I expect I will yes

vkconfig/configurator.cpp
christophe-lunarg: The idea is not to stop the compilation in the header. It's to catch all the places where platform specific code is required. 

If we don't, we likely need a lot more time to debug the code to figure out these places.

richard-lunarg: Okay. All my development environments break on #error. Why not just use #warning then since that's the explicit behavior.

Issue #1086, opened 2020-08-18
vkconfig: Don't show Windows only debug actions


richard-lunarg:
Reported by @lunarpapillo 

Minor, not new, not a showstopper.

If I understand correctly... Some or all of the validation Debug Action settings only apply to Windows, and do nothing on Linux (e.g. Debug Output, Break, Callback, maybe others).

These should be made obvious in vkconfig, either by removal, listing the settings in the GUI as Windows-only, or by presenting a dialog box if these are selected warning the user that these are no-ops on Linux (and Mac).

christophe-lunarg:
I have changes around to resolve this..

christophe-lunarg:
This issue is fixed with https://github.com/LunarG/VulkanTools/pull/1113

PR #1113, opened 2020-08-28
vkconfig: Filter out VK_DBG_LAYER_ACTION_DEBUG_OUTPUT on Linux and macOS #1086

christophe-lunarg:
Also add vkconfig dedicated PLATFORM_ define to allow using them within C++ expression to catch more build errors on all platforms.

Change-Id: I56bbaf013e69675f5284d59aed1e0c4645539e1e

### DIFF
--- a/vkconfig/configurator.cpp
+++ b/vkconfig/configurator.cpp
@@ -25,6 +25,7 @@
 
 #include "../vkconfig_core/version.h"
 #include "../vkconfig_core/util.h"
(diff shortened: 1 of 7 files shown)


christophe-lunarg:
Ohhhh very valid point! I'll use the Vulkan defines instead.

Oh damn it. There are using #ifdef isntead of #if. So I'll need to find something.

PR #1114, opened 2020-08-31
apidump: Make text pNext chain printing more legible

charles-lunarg:
Previous pNext chain printing put everything in a tree structure that is hard
to read. This makes the text output linear after the struct, making it easy
to understand.

Change-Id: I7d442d13fa9677bea42463f2bf6049948a2b7375


Tony-LunarG (approved):
Very nice!

PR #1128, opened 2020-09-08
vkconfig: Fix asserts on environment that ran old dev versions

christophe-lunarg:
Change-Id: I7c142d357282043018fbe932f1963c62f312abeb


richard-lunarg (approved):
I applaud your liberal use of asserts, but occasionally, we just need to handle the corner cases ;-)

christophe-lunarg:
I am not getting your comment. I think the corner case was detected by the assert so the assert did the job and this PR actually resolve it properly.

Issue #1130, opened 2020-09-08
Remove layer configs that don't work on macOS


richard-lunarg:
Both "Validation - GPU-Assisted", an the "Validation - Shader Printf" crash on macOS. MoltenVK does not support the required features of Vulkan for these types of validation to work. Even if the layer could be made to behave more gracefully, these two presets should be removed, but only for macOS builds.

christophe-lunarg:
This issue is fixed with #1133 PR

PR #1133, opened 2020-09-09
vkconfig: Mode to discard shader based configurations and presets #1130

christophe-lunarg:
Change-Id: I1438c51e7606013ce2ef9afb27f0e1d566893d11


richard-lunarg (approved):
I tested this branch on macOS and it seems to work fine with all cases accounted for.

PR #1145, opened 2020-09-14
Fix for recent macOS regressions

richard-lunarg:
Change-Id: I513681dbd3752af3b066dba90c66a744e896950c
Mostly pertaining to restoring the correct handling of the unique nature of .app program bundles.

### DIFF
--- a/vkconfig/dlgcreateassociation.cpp
+++ b/vkconfig/dlgcreateassociation.cpp
@@ -308,6 +308,9 @@ void dlgCreateAssociation::GetExecutableFromAppBundle(QString &app_path) {
             // Complete the partial path
             path += QString("MacOS/");
             path += QString(cExeName);
+
+            // Return original if not found, but root if found
+            app_path = path;
             delete[] cExeName;
             break;
         }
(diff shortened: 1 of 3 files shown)


vkconfig_core/path_manager.cpp
```diff
@@ -306,7 +306,13 @@ QString PathManager::SelectPathImpl(QWidget* parent, Path path, const QString& s
                 return "";
 
             SetPath(path, QFileInfo(selected_path).absolutePath());
+#if PLATFORM_WINDOWS
             return GetFullPath(path, QFileInfo(selected_path).baseName());
+#else
+            // On Linux, we have no extension to trim off. On macOS, we may or may not have a
+            // .app extension. If the later, it must be preserved at this point.
+            return GetFullPath(path, QFileInfo(selected_path).absoluteFilePath());
```
christophe-lunarg, line 314: This is a little weird as GetFullPath expect the filename not a path. So in theory, it should be QFileInfo(selected_path).filename() or something like this instead and actually that should work on Windows too.

vkconfig/dlgcreateassociation.cpp
```diff
@@ -308,6 +308,9 @@ void dlgCreateAssociation::GetExecutableFromAppBundle(QString &app_path) {
             // Complete the partial path
             path += QString("MacOS/");
             path += QString(cExeName);
+
+            // Return original if not found, but root if found
+            app_path = path;
```
christophe-lunarg, line 313: Oups indeed...

christophe-lunarg:
I am not 100% sure about one of the change but read my comment and I'll let you judge of it.

christophe-lunarg (approved):

Issue #1150, opened 2020-09-15
vkconfig: Changing a layer to exclude and then override cause the layer to have no settings


christophe-lunarg:
![no_settings](https://user-images.githubusercontent.com/62888873/93221899-92d33580-f76e-11ea-8002-3c2b108f8d3e.gif)

christophe-lunarg:
This issue is fixed with PR #1150

PR #1154, opened 2020-09-15
vkconfig: Fix no setting when switch between exclude and override #1150

christophe-lunarg:
- The settings are stored when switching to excluding the layer
- The settings are restored to the user selection when switching from exclude
to override
- When creating a new configuration with the validation layer the preset shows
'Standard' instead of 'User Defined'
- Fix a memory leak of default settings

Now fixed:
no_settings_fixed

Previously broken (and they were no way to restore the settings hence the P0):
no_settings

Change-Id: I429ba4f4b47917b5284ba69998425cd63a18a93f


richard-lunarg (approved):
Good to go. Nice touch.

Issue #1160, opened 2020-09-15
vkconfig: Ensure new devsim layer is supported with adequate layer ordering


richard-lunarg:
The Select Layers dialog has lost the ability to order layer precedence. While the override layers do not have "complete" control over the layers an application loads, it does control those that are provided by vkconfig. We have to have the ability to configure the override layers order. For example, the DevSim layer must be "closest to the driver". There are other dependencies as well according to @mark-lunarg .

Although not perfect in the first release, this feature is required by some use cases and has to be included until a new approach can be implemented.

christophe-lunarg:
I think we can deal with these specifics cases with built-in solution by default instead of confusing the users with something impossible to figure out.

For devsim, my plan is to add two configurations:
"Device simulation with validation"
"Device simulation" that exclude validation to prevent the application to validate against the drivers as application controlled layers are executed after overridden layers which are component layers of a implicit layer.

mark-lunarg:
I am confused, does this mean that `within `the set of layers 'managed' by vkconfig, that order cannot be specified by the user?

richard-lunarg:
@mark-lunarg That is correct. This ability was removed. The layers were ordered as the user specified in the layer override, and according to a past conversation with @charles-lunarg, the loader would respect that order. @christophe-lunarg is correct though that if DevSim has to be "next to driver" we should not leave this to the user to do, know about, or mess up. But I think there is some misunderstanding about the utility of the layer order within the vkconfig controlled list. Do you have some example use cases that might help?

richard-lunarg:
Also, @christophe-lunarg may have a very good point about the ordering. Can the layers really be reordered to different effect, or is it just that are there some specific requirements that some layers can't be behind or in front of others. If this is really the case, it should not be left to the user to determine.

mark-lunarg:
The layers can be ordered to different effect.  The tool can offer suggested orderings and/or warn for non-optimal ones, but should absolutely allow arbitrary changes to the layer order.

mark-lunarg:
As a single example, there are common cases where a utility layer might be placed before or after the Khronos layer, according to the desired result.

christophe-lunarg:
With #1165 Vulkan Configurator sorts the layers with what I tried to be the least invasive compared to the Vulkan loader bu design order making adding layers relevant for 99% or more of the users.

[...]

PR #1165, opened 2020-09-16
vkconfig: Add support for VK_KHR_portability_subset to devsim

christophe-lunarg:
- Add a setting to be able to enable VK_KHR_portability_subset emulation with devsim
- Sort the layers in a matter that reflect the layer properties. devsim layer will always be ordered last, close to the device.
- This PR resolved at least partially #1160

TODO:
- Check the version of devsim layer before displaying VK_KHR_portability_subset emulation setting.

### DIFF
--- a/vkconfig/configurator.cpp
+++ b/vkconfig/configurator.cpp
@@ -794,7 +794,14 @@ void Configurator::LoadDefaultLayerSettings() {
         const QJsonValue &layer_value = layers_options_object.value(layers_with_settings[i]);
         const QJsonObject &layer_object = layer_value.toObject();
 
-        ::LoadSettings(layer_object, settings_defaults.settings);
+        Parameter parameter;
+        parameter.name = settings_defaults.layer_name;
+        parameter.state = LAYER_STATE_APPLICATION_CONTROLLED;
+        parameter.settings = settings_defaults.settings;
+
+        ::LoadSettings(layer_object, parameter);
+
+        settings_defaults.settings = parameter.settings;
 
         // Add to my list of layer settings
         _default_layers_settings.push_back(settings_defaults);
--- a/vkconfig/configurator.cpp
+++ b/vkconfig/configurator.cpp
@@ -192,8 +192,8 @@ Configurator::Configurator()
 bool Configurator::Init() {
     // Load simple app settings, the additional search paths, and the
     // override app list.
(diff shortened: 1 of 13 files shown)


richard-lunarg (changes requested):
Almost there. None of the settings for the device simulation layer are saved. Even just switching from the active configuration and back clears them all to blank again. I stepped into the settings tree, and the editProfile slot is being called, but then it goes into unfamiliar code (I think you are caching settings somewhere else now). I figured you probably could fix it faster than me figuring it out.

christophe-lunarg:
Ok, this is a little weird, I'll investigate !

christophe-lunarg:
@richard-lunarg I fixed this issue with a dedicated PR: #1170 It's actually the display of the setting that was broken. One line fix!

richard-lunarg:
The json file was also not being saved. I'll double check, but that might still be broken.

richard-lunarg:
Okay, got to that quicker than I thought. Seems to be fixed.

richard-lunarg (approved):

jeremyk-lunarg (commented):
On my system VkConfig seems to always turn Portability emulation on.

jeremyk-lunarg (commented):
I'm still getting weird behavior when I test this change with vulkaninfo.

When I set up DevSim in VkConfig and run vulkaninfo, it acts as though portability emulation is always on and adds extra entries to the list of device extensions.

When I set up DevSim manually, everything works as expected.

christophe-lunarg:
@jeremyk-lunarg the problem was that I didn't rebase master that had this https://github.com/LunarG/VulkanTools/pull/1170 PR which fix the display of "bool_numeric" settings which this setting is.

johnzupin:
> None of the settings for the device simulation layer are saved. Even just switching from the active configuration and back clears them all to blank again.

I'm seeing this on Linux as well. The .json file is being saved between sessions.

johnzupin:
I'm not seeing the portability extension always being on when using vkconfig

johnzupin:
> >    None of the settings for the device simulation layer are saved. Even just switching from the active configuration and back 
> > clears them all to blank again.

> I'm seeing this on Linux as well. The .json file is being saved between sessions.

Please ignore this comment. Pulled in master and this looks good! Setting are being saved.

PR #1181, opened 2020-09-21
vkconfig: manual ordering of overridden layers

christophe-lunarg:
- Add manual ordering of overridden layers only
- Remove custom paths area
- Add custom paths button
- Add devsim override request all explicit tp be overridden
- Save layers window states
- Fix Configurator::HasMissingLayers
- Notify the user on start if layers are missing


johnzupin (approved):

johnzupin:
This looks good. I've tested Linux and Windows without any problems.

PR #1182, opened 2020-09-21
devsim: Add emulation for portability extension

jeremyk-lunarg:
Added emulation for VK_KHR_portability_subset to the Device Simulation
layer.

Also added the ability to override values in the
VK_KHR_portability_subset data structures with JSON config files.

This change requires a header update before it can be merged.

### DIFF
--- a/layersvt/device_simulation.cpp
+++ b/layersvt/device_simulation.cpp
@@ -54,6 +54,7 @@
 #include <json/json.h>  // https://github.com/open-source-parsers/jsoncpp
 
 #include "vulkan/vk_layer.h"
+#include "vulkan/vulkan_beta.h"
 #include "vk_layer_config.h"
 #include "vk_layer_table.h"
 
@@ -67,7 +68,7 @@ namespace {
 // layersvt/{linux,windows}/VkLayer_device_simulation*.json
 
 const uint32_t kVersionDevsimMajor = 1;
-const uint32_t kVersionDevsimMinor = 3;
+const uint32_t kVersionDevsimMinor = 4;
 const uint32_t kVersionDevsimPatch = 0;
 const uint32_t kVersionDevsimImplementation = VK_MAKE_VERSION(kVersionDevsimMajor, kVersionDevsimMinor, kVersionDevsimPatch);
 
@@ -88,7 +89,9 @@ const std::array<VkExtensionProperties, 0> kInstanceExtensionProperties = {};
 const uint32_t kInstanceExtensionPropertiesCount = kInstanceExtensionProperties.size();
 
 // Device extensions that this layer provides:
-const std::array<VkExtensionProperties, 0> kDeviceExtensionProperties = {};
+const std::array<VkExtensionProperties, 2> kDeviceExtensionProperties = {
+    {{VK_EXT_TOOLING_INFO_EXTENSION_NAME, VK_EXT_TOOLING_INFO_SPEC_VERSION},
+     {VK_KHR_PORTABILITY_SUBSET_EXTENSION_NAME, VK_KHR_PORTABILITY_SUBSET_SPEC_VERSION}}};
 const uint32_t kDeviceExtensionPropertiesCount = kDeviceExtensionProperties.size();
 
 // The "standard" core VkFormat enum values:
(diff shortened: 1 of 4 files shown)


Groovounet:
My perception is that this PR should come with more than a portability test json file but also with a default json file that the Vulkan developers would use.

Groovounet:
Aren't they full set of limits that comes with the Vulkan Portability? 
"VkPhysicalDevicePortabilitySubsetFeaturesKHR": {
        "imageViewFormatReinterpretation": 0,
        "pointPolygons": 0,
        "samplerMipLodBias": 0
    },

layersvt/device_simulation.cpp
```diff
@@ -1278,15 +1390,73 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateInstance(const VkInstanceCreateInfo *pCreat
...
+
         dt->GetPhysicalDeviceProperties(physical_device, &pdd.physical_device_properties_);
+
+        // Initialize PDD members to the actual Vulkan implementation's defaults.
+        if (get_physical_device_properties2_active || pdd.physical_device_properties_.apiVersion > VK_VERSION_1_0) {
+            VkPhysicalDeviceProperties2KHR property_chain = {VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_PROPERTIES_2_KHR};
+            VkPhysicalDeviceFeatures2KHR feature_chain = {VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_FEATURES_2_KHR};
+            VkPhysicalDeviceMemoryProperties2KHR memory_chain = {VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_MEMORY_PROPERTIES_2_KHR};
+
+            if(PhysicalDeviceData::HasExtension(physical_device, VK_KHR_PORTABILITY_SUBSET_EXTENSION_NAME)) {
```
mark-lunarg, line 1417: Looks like a glitch in clang-format?

layersvt/device_simulation.cpp
```diff
@@ -1331,6 +1507,30 @@ VKAPI_ATTR void VKAPI_CALL GetPhysicalDeviceProperties(VkPhysicalDevice physical
     }
 }
 
+// Utility function for iterating through the pNext chain of certain Vulkan structs.
+void fillPNextChain(PhysicalDeviceData *physicalDeviceData, void *place) {
```
mark-lunarg, line 1511: Function Name.

mark-lunarg (approved):
With an extremely limited exposure to portability which disqualifies me from commenting on completeness, I don't see anything glaringly wrong here.  Some nits which might lead to a green board after rebasing.

layersvt/device_simulation.md
```diff
@@ -135,19 +135,28 @@ If you wish DevSim to terminate on errors, set the `VK_DEVSIM_EXIT_ON_ERROR` env
...
 ```
 
+### `VK_KHR_portability_subset` Emulation
+
+The DevSim layer provides the ability to emulate the `VK_KHR_portability_subset` extension on devices that do not implement this extension.
+This feature allows users to test their application with limitations found on non-conformant Vulkan implementations.
+To turn on this feature, set the `VK_DEVSIM_EMULATE_PORTABILITY_SUBSET_EXTENSION` environment variable or `lunarg_device_simulation.emulate_portability`, the corresponding vk_layer_settings.txt option to a positive integer.
+DevSim config files that use this feature should validate to the following schema (link).
+
 ### Environment variables used by DevSim layer.
```
KarenGhavam-lunarG, line 145: To emphasize the availability and ease of use of the Vulkan Configurator, I believe a section needs to be added to this document referencing use of the Vulkan Configurator to enable the devsim layer and configure it.

KarenGhavam-lunarG: I am wondering if it is a better organization to talk about the environment variables and vk_layer_settings.txt options in the same section in the document, vs. pulling this into a separate section for the portability extension.

Also, shouldn't you update the DevSim Layer operation and configuration section with information on the new schema for the portability extension (indicate the schema presence and location)?

layersvt/device_simulation.cpp
```diff
@@ -1319,6 +1483,12 @@ VKAPI_ATTR void VKAPI_CALL DestroyInstance(VkInstance instance, const VkAllocati
     }
 }
 
+// Utility struct to check sType of pNext chain structures.
```
lenny-lunarg, line 1486: Look up `VkBaseInStructure` and `VkBaseOutStructure`. I'm pretty sure it's better to use those instead of creating your own type.

layersvt/device_simulation.cpp
```diff
@@ -1331,6 +1501,35 @@ VKAPI_ATTR void VKAPI_CALL GetPhysicalDeviceProperties(VkPhysicalDevice physical
     }
 }
 
+// Utility function for iterating through the pNext chain of certain Vulkan structs.
+void FillPNextChain(PhysicalDeviceData *physicalDeviceData, void *place) {
```
lenny-lunarg, line 1505: The only thing that's a little odd is that this will fill in a portability struct passed to the wrong API call. I couldn't find any spec language explicitly stating that it's illegal to do so, so I think this implementation is technically valid. The reason it still concerns me a little is because I wouldn't want applications to start depending on what is technically undefined behavior.

lenny-lunarg (commented):
As far as interfacing with the loader goes, it all looks right. Initialization is all done after calling down to the next layer (as it should be done). The json file looks right and since you're not adding any new API functions there's no need to add a `vkGetPhysicalDeviceProcAddr` function. I did note two minor things that I pointed to in the review. I don't think either is incorrect &mdash; I just think there's somewhat nicer ways of doing it.

layersvt/device_simulation.cpp
jeremyk-lunarg: Will make this change.  It's easy and cleaner.

layersvt/device_simulation.cpp
jeremyk-lunarg: This change might be a bit more difficult, so I'll put it off for now, but will fix it in the future.

layersvt/device_simulation.md
```diff
@@ -92,9 +92,16 @@ How the JSON configuration values are applied depends on whether the top-level s
...
-The JSON fileformat consumed by the DevSim layer is specified by a JSON schema, the canonical URI of which is "https://schema.khronos.org/vulkan/devsim_1_0_0.json#"
+JSON fileformats consumed by the DevSim layer are specified by one of the JSON schemas in the table below.
 
-The top-level sections of a configuration file are specified by the DevSim JSON schema, and are processed as follows:
+| Schema Use | Canonical URI |
+|:----------:|:-------------:|
+| Vulkan v1.0 | https://schema.khronos.org/vulkan/devsim_1_0_0.json# |
+| VK_KHR_portability_subset | https://schema.khronos.org/vulkan/devsim_VK_KHR_portability_subset-provisional-1.json# |
+
+Usually you will be using configuration files validated with the Vulkan v1.0 schema.
```
KarenGhavam-lunarG, line 102: Why do you say this?

KarenGhavam-lunarG (approved):
I think the documentation looks better now integrating the portability stuff throughout the document.

PR #1193, opened 2020-09-24
vkconfig: Hide VulkanInfo #1179

christophe-lunarg:
If we can't make VulkanInfo reliable, hide it until we have a fix...

Change-Id: I4c721a0431b773970e4a0686dacdc317962989cc


christophe-lunarg:
The issue was reproduced: 
NVIDIA 1.2.142 drivers + AMD 1.1.108 drivers => crash
NVIDIA 1.2.142 drivers + AMD 1.2.135 drivers => ok

Issue #1178, opened 2020-09-18
Devsim/Screenshot:  Visual Studio 2019 build warnings to fix


mark-lunarg:
Noticed these while building today...

```
VulkanTools\layersvt\device_simulation.cpp(88,87): warning C4267: 'initializing': conversion from 'size_t' to 'const uint32_t', possible loss of data
VulkanTools\layersvt\device_simulation.cpp(92,83): warning C4267: 'initializing': conversion from 'size_t' to 'const uint32_t', possible loss of data
VulkanTools\layersvt\screenshot.cpp(460,34): warning C4834: discarding return value of function with 'nodiscard' attribute
```

christophe-lunarg:
This issue is fixed

PR #1196, opened 2020-09-28
devsim: Fix conversion warnings

jeremyk-lunarg:
Fixed warnings in Device Simulation Layer about
conversion between 'size_t' and 'const uint32_t'.

Partially fixes issue #1178


jeremyk-lunarg:
@mark-lunarg, does this fix the DevSim warnings on your machine as shown in issue #1178?

PR #1197, opened 2020-09-29
macOS fixes for cmake builds

richard-lunarg:
Change-Id: I720dd07676ee57dbcc0be122e3ae10d7cbdc3d3f
This commit will be needed for macOS SDK builds.
Recent refactoring has broken macOS builds when using cmake.
In addition, the version of Qt used on macOS for SDK builds is less tolerant of forward slash/backslash issues, so it is corrected here as well.


christophe-lunarg (approved):
Looks good if tests are green!

PR #1206, opened 2020-10-01
vkconfig: Fixed macOS regression for vkcube search

richard-lunarg:
Change-Id: I3d5e1944600bfcc0cc09aa7b0e7a248d4a1afdf9
Fixed regressions for not being able to find vkcube on initial startup, and proper paths for the launcher feature.

### DIFF
--- a/vkconfig_core/environment.cpp
+++ b/vkconfig_core/environment.cpp
@@ -319,6 +319,25 @@ static QString GetDefaultExecutablePath(const QString& executable_name) {
 #error "Unknown platform"
 #endif
 
+#if PLATFORM_MACOS
+    // Using the standard install loation on macOS
+    {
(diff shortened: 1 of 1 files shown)


vkconfig_core/environment.cpp
```diff
@@ -319,6 +319,25 @@ static QString GetDefaultExecutablePath(const QString& executable_name) {
 #error "Unknown platform"
 #endif
 
+#if PLATFORM_MACOS
```
christophe-lunarg, line 322: Please use if (PLATFORM_MACOS) so that it always get compile on all platforms

christophe-lunarg (approved):

vkconfig_core/environment.cpp
richard-lunarg: Okay, done here.

PR #1207, opened 2020-10-02
devsim: Update API version check

dustin-lunarg:
Change the API version check used by the dev-sim layer to determine if vkGetPhysicalDeviceProperties2 is available from a comparison between the API version retrieved from vkGetPhysicalDeviceProperties and VK_VERSION_1_0, which is now defined as 1, to a comparison between the minor API version that the application specified through VkApplicationInfo and 0.


jeremyk-lunarg (approved):

jeremyk-lunarg:
@dustin-lunarg make sure to cherry-pick this over to the SDK branch after merging.

PR #1213, opened 2020-10-09
vkconfig: Add layers override using from command line arguments

christophe-lunarg:
Overriding layers using a command line:
image

Surrender layers using a command line:
image

Command line usage:
image

List the found layers on the system:
image

List the found layers with information regarding versions and locations
image


richard-lunarg (approved):
I had to tweak the .pro file to get it to build with QtCreator, but other wise looks good. Works on macOS too.

PR #1229, opened 2020-11-05
cmake: Update Vulkan-ValidationLayers_INCLUDE_DIR path

orbea:
This updates the `VULKAN_VALIDATIONLAYERS_INSTALL_DIR` path since it was changed in `Vulkan-ValidationLayers`. Please see PR https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/2307.


johnzupin:
I believe this PR needs to be applied before the next time known-good.json is updated.

Tony-LunarG:
Yes, I think this will need to update known-good.json to keep from breaking builds.  I'm running it through CI

Tony-LunarG (approved):
Ok, with update to headers version 1.2.159

orbea:
For future reference this is merged in https://github.com/LunarG/VulkanTools/commit/8a420c41285e3cb6d11e0e62a3fc0d39e667c04c.

Thanks!

PR #1237, opened 2020-11-10
via: remove newline from dpkg-query for system packages

johnzupin:
This newline was creating non-valid json files.


richard-lunarg (approved):
Good to go!

PR #1239, opened 2020-11-10
apidump: Use VkInstance instead of NULL in GIPA

### LINKED ISSUE #1233: API dump layer gets vkCreateDevice pointer illegally
Inside `vkCreateDevice` the api dump layer gets a pointer to the next layer like:

```c
PFN_vkCreateDevice fpCreateDevice = (PFN_vkCreateDevice)fpGetInstanceProcAddr(NULL, "vkCreateDevice");
```

The Vulkan spec [here](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#vkGetInstanceProcAddr) states that calling `vkGetInstanceProcAddr` with a NULL instance must return NULL for all but a few functions listed out in the spec. As such, a conforming implementation is required to return null here (though no implementation actually does).

The fix is to look up the correct `VkInstance` handle associate with the `VkPhysicalDevice` handle that is passed by this function and to pass that instance handle to `vkGetInstanceProcAddr`.

charles-lunarg:
The call to vkGetInstanceProcAddr to get vkCreateDevice used NULL for the
VkInstance handle. This is forbidden by the spec but was allowed on
various implementations. This commmit fixes it by keeping track of the
physical device and instance handles.

Fixes #1233 

Change-Id: Ib247c71913f1512f919349523a2058907616ee23

### DIFF
--- a/layersvt/api_dump.h
+++ b/layersvt/api_dump.h
@@ -808,6 +808,12 @@ class ApiDumpInstance {
 
     std::unordered_map<uint64_t, std::string> object_name_map;
 
+    void set_vk_instance(VkPhysicalDevice phys_dev, VkInstance instance) { vk_instance_map.insert({phys_dev, instance}); }
+    VkInstance get_vk_instance(VkPhysicalDevice phys_dev) const {
+        if (vk_instance_map.count(phys_dev) == 0) return VK_NULL_HANDLE;
+        return vk_instance_map.at(phys_dev);
+    }
+
    private:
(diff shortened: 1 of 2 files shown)


charles-lunarg:
As far as sync goes, `dump_head` locks the main mutex while `dump_body` unlocks it, so as long as all state mutation happens between those two calls, there shouldn't be any issue.

lenny-lunarg (approved):

PR #1249, opened 2020-11-18
vkconfig: Fix initialization of added applications on macOS

richard-lunarg:
Fixed a couple of regressions, one not Mac related at all. Code to extract an executable from app bundles was not being called for the default applications vkcube and vkcubepp on setup for macOS.
Paths to these applications in the override json file had an extra forward slash after Applications in the path (as in Applications//vkcube.app)
On all platforms, the list of applications was always added to the override json file, even when "Apply only to the selected list of Vulkan applications" was unchecked.

### DIFF
--- a/vkconfig/CHANGELOG.md
+++ b/vkconfig/CHANGELOG.md
@@ -26,6 +26,7 @@
 - Fix layers override update when changing layers management options #1225
 - Fix reset button in layers window #1227
 - Fix warning about old loader when trying to use application list #1230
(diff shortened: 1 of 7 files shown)


christophe-lunarg:
I also fixed the case where vkcube is not found and the override file created with relative path for the applications added with relative path.

christophe-lunarg (approved):
I made a little more changes as application added with a relative path would not work.

I tested my change on macOS and Windows but couldn't initialize gerrit as I broke my configuration...

PR #1258, opened 2020-12-01
Changes for 1.2.162 headers

Tony-LunarG:


charles-lunarg (commented):
Changes to Api Dump look good.

PR #1260, opened 2020-12-03
apidump: Add 2 dimentional arrays to autogen

charles-lunarg:
VkTransformMatrixKHR's printer did not print the float[3][4]
matrix subobject, leaving a whole in its output. This commit
fixes that whole by hardcoding a 2 dimensional array printer.
If more dimensions are needed in the future, those will have
to be hardcoded. Because that is highly unlikely, creating a
fully generic array printer that works for N dimensions was
not necessary.

Change-Id: I2c437314bd843adbc7cbe3a0dd4010d297d8989f

### DIFF
--- a/layersvt/api_dump.h
+++ b/layersvt/api_dump.h
@@ -958,6 +958,24 @@ inline void dump_text_array_hex(const uint32_t *array, size_t len, const ApiDump
     }
 }
 
+template <typename T, size_t len1, size_t len2, typename... Args>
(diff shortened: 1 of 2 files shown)


charles-lunarg:
As it turns out, the VkTransformMatrixKHR struct is not a structure that API dump will ever print.
VkTransformMatrixKHR is apart of VkAccelerationStructureInstanceKHR, but that struct is only to be written into a VkBuffer, so it isn't accessible by anything which passes through api dump.
As such, this PR is goign to be mothballed as theres no sense in adding code that no one uses, and is unlikely to be used in the near future (multi-dimentional arrays are not common)

PR #1272, opened 2020-12-10
vkconfig: fix clang build

christophe-lunarg:
It looks like I messed up!

### DIFF
--- a/vkconfig_core/util.h
+++ b/vkconfig_core/util.h
@@ -29,8 +29,8 @@
 #include <cstdio>
 #include <cassert>
 #include <string>
-#include <array>
 #include <vector>
+#include <array>
 
 // Based on https://www.g-truc.net/post-0708.html#menu
 template <typename T, std::size_t N>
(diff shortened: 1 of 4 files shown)


vkconfig_core/util.h
```diff
@@ -55,7 +55,7 @@ bool IsNumber(const std::string& s);
 
 template <typename T>
 typename std::vector<T>::iterator FindItByKey(std::vector<T>& container, const char* key) {
-    assert(key != nullptr && key != "");
+    assert(key != nullptr);
```
lunarpapillo, line 58: Is the change in behavior intentional?  The original assertion could be rewritten as something like:
```
assert(key != nullptr && key[0] != '\0');
```
or:
```
assert(key != nullptr && std::strcmp(key, "") != 0);
```
right?

lunarpapillo (commented):
The switch of `strcmp` to `std::strcmp` looks good.  The other changes alter behavior; if that's intentional, it's fine with me.

vkconfig_core/util.h
christophe-lunarg: Yes expect, key != "" cause a warning as this is undefined

lunarpapillo: I do see why `key != ""` would not work.  But removing the check entirely changes behavior, which I don't expect from a change intended only to fix compiler errors, as there were other ways to change the code that would have preserved the original behavior.

If the behavior change is intentional, I'm okay with it.

lunarpapillo: Why was the check for an empty string removed entirely, instead of being fixed to use `std::strcmp`?

christophe-lunarg: Ah good point, my bad. I've fixed it with upcoming PR.

PR #1284, opened 2020-12-14
apidump: update doc to add note about enabling layer with vkconfig

johnzupin:
update doc to add note about enabling layer with vkconfig

### DIFF
--- a/layersvt/api_dump_layer.md
+++ b/layersvt/api_dump_layer.md
@@ -20,6 +20,10 @@ or by using the `vk_layer_settings.txt` file.
 
 ## Enabling the Layer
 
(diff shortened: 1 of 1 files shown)


KarenGhavam-lunarG (commented):
Should the reference to the vkconfig document be the latest version in the SDK instead?
This document will get pulled into the SDK and it would be better to have it link to another SDK document.

Check the documentation for synchronization in the validation layer repository. I believe this is what is done there.

johnzupin:
Updated the link to point to the vulkan.lunarg.com documentation

KarenGhavam-lunarG (approved):

PR #1321, opened 2021-01-08
devsim: Only make properties2 when enabled

ncesario-lunarg:
Make sure VK_KHR_get_physical_device_properties2 function calls are only
made on physical devices that support the extension.

FYI @richard-lunarg @mikes-lunarg @jeremyk-lunarg. This at least gets vkcube working on my AMD machine.

### DIFF
--- a/layersvt/device_simulation.cpp
+++ b/layersvt/device_simulation.cpp
@@ -1482,7 +1482,7 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateInstance(const VkInstanceCreateInfo *pCreat
         dt->GetPhysicalDeviceProperties(physical_device, &pdd.physical_device_properties_);
 
         // Initialize PDD members to the actual Vulkan implementation's defaults.
-        if (get_physical_device_properties2_active || VK_VERSION_MINOR(requested_version) > 0) {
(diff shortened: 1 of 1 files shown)


layersvt/device_simulation.cpp
```diff
@@ -1482,7 +1482,7 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateInstance(const VkInstanceCreateInfo *pCreat
         dt->GetPhysicalDeviceProperties(physical_device, &pdd.physical_device_properties_);
 
         // Initialize PDD members to the actual Vulkan implementation's defaults.
-        if (get_physical_device_properties2_active || VK_VERSION_MINOR(requested_version) > 0) {
+        if (PhysicalDeviceData::HasExtension(physical_device, VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2_EXTENSION_NAME) && (get_physical_device_properties2_active || VK_VERSION_MINOR(requested_version) > 0)) {
```
lunarpapillo, line 1485: I think this would better be placed as a guard around lines 1466-1471, where `get_physical_device_properties2_active` is set.

lunarpapillo: Though it's strange to me that the extension could be enabled while it is not present... ?

ncesario-lunarg:
While this gets vkcube working, this is not correct. From JeremyK: "`PhysicalDeviceData::HasExtension` only checks for device extensions," and `VK_KHR_get_physical_device_properties2` is an instance extension. So this is just avoiding the crash, not fixing it.

lunarpapillo (changes requested):
Minor changes only.  You can ignore if your prefer your current arrangement.

PR #1324, opened 2021-01-08
devsim: Add build for macOS

richard-lunarg:
Change-Id: Ifb86d613e548bdce080124ab6886b03c3751c4ac

Build device simulation layer on macOS.

### DIFF
--- a/layersvt/CMakeLists.txt
+++ b/layersvt/CMakeLists.txt
@@ -61,6 +61,7 @@ if (NOT APPLE)
 else()
     set(TARGET_NAMES
         VkLayer_api_dump
+	VkLayer_device_simulation
         )
 endif ()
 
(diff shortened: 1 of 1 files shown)


layersvt/CMakeLists.txt
```diff
@@ -61,6 +61,7 @@ if (NOT APPLE)
 else()
     set(TARGET_NAMES
         VkLayer_api_dump
+	VkLayer_device_simulation
```
jeremyk-lunarg, line 64: Might want to fix this indent for neatness.

jeremyk-lunarg (approved):
LGTM other than one nit.

layersvt/CMakeLists.txt
richard-lunarg: Too quick on the trigger finger. Will fix right on master.

PR #1361, opened 2021-02-16
build: update Vulkan-Headers known_good to v1.2.170

jeremyg-lunarg:
This requires fixing the api_dump layer's code generation to
handle the new VkFlags64 type.


charles-lunarg (approved):
API Dump changes look good.

Issue #1320, opened 2021-01-08
DevSim: Disable simulation of memory type information


karl-lunarg:
Vulkan Memory Types really can't be simulated cleanly by DevSim.  Unlike other quantitative properties such as the number of descriptor slots, DevSim can't simulate memory types as easily.

Issue #1311 is an example of the basic problem, and it has occurred in other circumstances. 

To increase the usability of DevSim, I'd suggest that DevSim no longer replaces memory type information by default.  This will allow users to focus more easily on the other simulations provided by DevSim without memory type problems getting in the way. 
 For those cases where a user really wants to exercise memory type simulation, add an option, controlled by the usual layer settings mechanisms, that controls the memory type replacement as follows:

* 0 (default) - Don't replace memory type information at all.  All results of memory type queries come from the actual device.
* 1 - Replace the memory type information only if it can be fully simulated by a simple mapping.
* 2 - Replace the memory type information completely, as it does today.

Notes on Option 1:
This simulation option could fail and if it does, DevSim should emit a message and continue with the default option 0.

One way to implement this would be for DevSim to examine the memory type arrays from the physical device memory properties for both the simulated device and the real device.  If the arrays are equivalent, but just in a different order, DevSim could return the list from the real device.  Another approach would be to return the simulated list but perform an array index mapping whenever the memory type index is supplied.  This option is harder to implement but is more "true" in the sense that the simulated memory types are returned.

If the simulated array is a subset of the real array, then the subset can be reported by DevSim with potentially the need to map indices.  This would be very useful for the case where a user would want to test a case where a certain memory type was not available and the application uses a fallback path.

For example, if a real device has:
```
memoryTypes: count = 5
        memoryTypes[0]:
                heapIndex     = 1
                propertyFlags = 0x0000: count = 0
                        None
        memoryTypes[1]:
                heapIndex     = 0
                propertyFlags = 0x0001: count = 1
                        MEMORY_PROPERTY_DEVICE_LOCAL_BIT
        memoryTypes[2]:
                heapIndex     = 1
                propertyFlags = 0x0006: count = 2
                        MEMORY_PROPERTY_HOST_VISIBLE_BIT
                        MEMORY_PROPERTY_HOST_COHERENT_BIT
        memoryTypes[3]:
                heapIndex     = 1
                propertyFlags = 0x000e: count = 3
                        MEMORY_PROPERTY_HOST_VISIBLE_BIT
[...]

PR #1362, opened 2021-02-16
devsim: Add option to enable/disable memory sim

jeremyk-lunarg:
Added a new option `VK_DEVSIM_MODIFY_MEMORY_FLAGS` to allow users to
enable/disable devsim's ability to modify memory heap flags and memory
type flags.  Disabled by default.

Devsim's ability to modify memory heap size and count are not affected
by this change.

Fixes issue #1320


christophe-lunarg (approved):
Looks good to me!
I'll expose this new setting in vkconfig

PR #1365, opened 2021-02-22
build: Update known-good for VVL and Vulkan-Tools

jeremyg-lunarg:


Tony-LunarG (approved):
LGTM once CI passes

richard-lunarg (commented):
Both of these branches (1.2.170) exist or will exist. Are we sure this should not be branch instead of commit? If so, good to go. Double checking my own ignorance.

Tony-LunarG:
There will be an SDK branch at or after this commit.  But there should have been a tag added at the 1.2.170 update

jeremyg-lunarg:
> There will be an SDK branch at or after this commit. But there should have been a tag added at the 1.2.170 update

Oops. I went back and added the tag:

```
commit ddfdcd6c99a2c7150c277b59ab5ea407ec48bb1e (tag: v1.2.170)
Author: Jeremy Gebben <jeremyg@lunarg.com>
Date:   Tue Dec 29 12:25:17 2020 -0700

    build: update Vulkan-Headers known_good to v1.2.170
    
    This requires fixing the api_dump layer's code generation to
    handle the new VkFlags64 type.

```

PR #1385, opened 2021-03-04
devsim: Update to Vulkan 1.1

jeremyk-lunarg:
Updated Devsim to modify structs provided by extensions made core in Vulkan 1.1.
The following extensions are included:
- `VK_KHR_16bit_storage`
- `VK_KHR_get_physical_device_properties2`
- `VK_KHR_maintenance2`
- `VK_KHR_maintenance3`
- `VK_KHR_multiview`
- `VK_KHR_sampler_ycbcr_conversion`
- `VK_KHR_variable_pointers`

The following structs are now modified by the Devsim layer:
- `VkPhysicalDeviceProperties2`
- `VkPhysicalDeviceMaintenance3Properties`
- `VkPhysicalDeviceMultiviewProperties`
- `VkPhysicalDevicePointClippingProperties`
- `VkPhysicalDeviceFeatures2`
- `VkPhysicalDevice16BitStorageFeatures`
- `VkPhysicalDeviceMultiviewFeatures`
- `VkPhysicalDeviceSamplerYcbcrConversionFeatures`
- `VkPhysicalDeviceVariablePointersFeatures`
- `VkPhysicalDeviceMemoryProperties2`
- `VkQueueFamilyProperties2`

These updates require new JSON schemas to define format for new Devsim config files.
These JSON schema files and example test config files can be provided upon request.

### DIFF
--- a/layersvt/device_simulation.cpp
+++ b/layersvt/device_simulation.cpp
@@ -16,6 +16,7 @@
  *
  * Author: Mike Weiblen <mikew@lunarg.com>
  * Author: Arda Coskunses <arda@lunarg.com>
+ * Author: Jeremy Kniager <jeremyk@lunarg.com>
  */
 
 /*
@@ -64,12 +65,12 @@ namespace {
 
 // For new features/functionality, increment the minor level and reset patch level to zero.
 // For any changes, at least increment the patch level.  See https://semver.org/
-// When updating the version, be sure to make corresponding changes to the layer manifest files at
-// layersvt/{linux,windows}/VkLayer_device_simulation*.json
+// When updating the version, be sure to make corresponding changes to the layer manifest file at
+// layersvt/VkLayer_device_simulation.json.in
 
 const uint32_t kVersionDevsimMajor = 1;
-const uint32_t kVersionDevsimMinor = 5;
-const uint32_t kVersionDevsimPatch = 1;
+const uint32_t kVersionDevsimMinor = 6;
+const uint32_t kVersionDevsimPatch = 0;
 const uint32_t kVersionDevsimImplementation = VK_MAKE_VERSION(kVersionDevsimMajor, kVersionDevsimMinor, kVersionDevsimPatch);
 
 // Properties of this layer:
--- a/layersvt/device_simulation.cpp
+++ b/layersvt/device_simulation.cpp
@@ -578,6 +578,10 @@ class PhysicalDeviceData {
     ArrayOfVkLayerProperties arrayof_layer_properties_;
     ArrayOfVkExtensionProperties arrayof_extension_properties_;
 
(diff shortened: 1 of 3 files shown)


christophe-lunarg:
Out of curiosity, what's your method to ensure the modifications are exhaustive?

jeremyk-lunarg:
> Out of curiosity, what's your method to ensure the modifications are exhaustive?

What do you mean by "exhaustive"?

christophe-lunarg:
That nothing is missing. It doesn't seem trivial to me with Vulkan \o/

jeremyk-lunarg:
> That nothing is missing. It doesn't seem trivial to me with Vulkan \o/

I did the following process to get the final list of things to implement:
1. Created a list of extensions present in Vulkan 1.0 and a list of extensions deprecated in Vulkan 1.1.
2. Used a Python script to find which extensions were in both lists.
3. Removed extensions that had not been made core in Vulkan 1.1 from the deprecation list.
4. Since Devsim modifies queries which are provided by `VK_KHR_get_physical_device_properties2`, I removed all extensions that did not require `VK_KHR_get_physical_device_properties2`.
5. I further removed extensions that did not provided queries that Devsim would be interested in modifying.
6. This gave me the final list of extensions to implement.  From here I checked each extension for queries it provided that Devsim would be interested in modifying, and made a list of those.
7. Finally, I had @karl-lunarg and @mark-lunarg take a look at the list and give me the "okay".

layersvt/device_simulation.cpp
```diff
@@ -1019,6 +1179,35 @@ void JsonLoader::GetValue(const Json::Value &parent, const char *name, VkPhysica
     GET_VALUE(sparseProperties);
 }
 
+void JsonLoader::GetValue(const Json::Value &parent, const char *name, VkPhysicalDeviceMaintenance3PropertiesKHR *dest) {
```
ncesario-lunarg, line 1182: I'm guessing there's a good reason not to, but would it be possible to generate these from the XML? You might get a bunch of getters you don't care about, but then this would be one less thing to worry about updating.

ncesario-lunarg (approved):
From my limited knowledge...LGTM! My only concern would be the same as Christophe's. Would be awesome if there is a way to do a lot of this automatically from the XML, but--even if there is a way--it's probably not worth the effort given how frequently a new Vulkan version comes out.

layersvt/device_simulation.cpp
jeremyk-lunarg: This is one of the points of pain when adding new extensions and query overrides to Devsim.

I plan on refactoring the code to make it more easily extendable after it's been updated to simulate Vulkan 1.2 devices, and I'll keep this suggestion in mind while doing that.

christophe-lunarg (approved):

PR #1400, opened 2021-03-31
Include Guards Were Missing

brightprogrammer:
[ingene_build.log](https://github.com/LunarG/VulkanTools/files/6237937/ingene_build.log)
Multiple definition error was being raised by the compiler on Linux (g++/gcc) when included `vk_enum_string_helper.h` due to absence of include guards.

### DIFF
--- a/scripts/tool_helper_file_generator.py
+++ b/scripts/tool_helper_file_generator.py
@@ -157,17 +157,26 @@ class ToolHelperFileOutputGenerator(OutputGenerator):
         copyright += ' *\n'
         copyright += ' ****************************************************************************/\n'
         write(copyright, file=self.outFile)
+        
+        # begin include guards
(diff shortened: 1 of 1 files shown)


Tony-LunarG:
@brightprogrammer vk_enum_string_helper.h is generated from building the Vulkan-ValidationLayers repository:

external\Vulkan-ValidationLayers\build\install\include\vulkan\vk_enum_string_helper.h
external\Vulkan-ValidationLayers\layers\generated\vk_enum_string_helper.h

so if this PR is intended to add include guards to that file, this is not the repo to do it in.  Or is there a header file generated by this repo that needs guards, and if so, which one(s)?

brightprogrammer:
@Tony-LunarG thanks for replying, I thought that since it is a tool, it must belong to this repo but now it looks like I was mistaken. I will close the PR

thanks 🙂

PR #1423, opened 2021-04-09
devsim: allow multiple enumerations of phys devices

### LINKED ISSUE #1422: devsim: Crash when replaying D3D11 RenderDoc traces via Wine + DXVK
Using devsim layer when replaying any D3D11 RenderDoc trace with Wine + DXVK causes crash:

```
renderdoccmd.exe: ../layersvt/device_simulation.cpp:526: static void {anonymous}::PhysicalDeviceData::Destroy(VkPhysicalDevice): Assertion `Find(pd)' failed.
```

Which happens because there are several calls to `vkEnumeratePhysicalDevices` with different `VkInstance`, but layer populates phys device map only once.

werman:
Application may create several Vulkan instances and call `vkEnumeratePhysicalDevices` for each of them.
Thus, instead of populating physical device map only once, we should check whether we already seen the device very time.

Such case could be observed while replaying D3D11 RenderDoc trace via Wine + DXVK.

Fixes #1422

### DIFF
--- a/layersvt/device_simulation.cpp
+++ b/layersvt/device_simulation.cpp
@@ -2105,16 +2105,13 @@ VKAPI_ATTR VkResult VKAPI_CALL EnumeratePhysicalDevices(VkInstance instance, uin
                                                         VkPhysicalDevice *pPhysicalDevices) {
     // Our layer-specific initialization...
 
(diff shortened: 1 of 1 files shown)


jeremyk-lunarg (approved):
Passing internal and cloud CI.
LGTM

PR #1432, opened 2021-04-13
devsim: add settings to layer manifest

christophe-lunarg:
Surrender devsim layer settings to devsim developers.

Vulkan Configurator will use these settings when loading devsim layer.


jeremyk-lunarg (approved):
LGTM

PR #1455, opened 2021-04-20
Update to header 1.2.176

Tony-LunarG:


charles-lunarg (approved):
The cmake file could have some formatting if you feel up to it, but the rest of the content looks good.

PR #1490, opened 2021-05-06
license: add valijon license

christophe-lunarg:
The root license file says:
```
The majority of files in this project use the Apache 2.0 License.
There are a few exceptions and their license can be found in the source.
Any license deviations from Apache 2.0 are "more permissive" licenses.
```

Just adding this file seems to resolve the issue. 

compat/optional.hpp has the license embedded within the source so I guess it's ok.


christophe-lunarg:
Ok changes done. to the root license file.

KarenGhavam-lunarG:
Looks good.

PR #1507, opened 2021-05-21
api dump: add url and introduction to the layer manifest

christophe-lunarg:
Add an introduction and a URL in the layer manifest.
These are used by Vulkan Configurator for the Vulkan Developers to find the layer documentation.


charles-lunarg (approved):
Looks good.
Link works as expected.

PR #1513, opened 2021-05-22
screenshot: adjust synchronization

dustin-lunarg:
The screenshot layer was calling both vkQueueWaitIdle and vkDeviceWaitIdle after performing the swapchain image copy, without performing any synchronization before the swapchain image copy.  This change moves the vkDeviceWaitIdle call before the copy, to wait for any outstanding work that may modify the swapchain image to complete.

Because the copy is performed in the layer's implementation of vkQueuePresentKHR, prior to making the actual vkQueuePresentKHR call, it may be possible to further improve synchronization by stealing the wait semaphores intended for vkQueuePresentKHR to use with the vkQueueSubmit that performs the copy so that the vkDeviceWaitIdle call can be removed.

### DIFF
--- a/layersvt/screenshot.cpp
+++ b/layersvt/screenshot.cpp
@@ -994,13 +994,14 @@ static void writePPM(const char *filename, VkImage image1) {
     submitInfo.signalSemaphoreCount = 0;
     submitInfo.pSignalSemaphores = NULL;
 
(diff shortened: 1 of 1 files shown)


Tony-LunarG (approved):
LGTM

PR #1535, opened 2021-06-08
devsim: Add combination modes for format props

jeremyk-lunarg:
Implemented format list and property modification to be separately
controlled by array combination mode enums.


christophe-lunarg (approved):
Expected for the failed C.I. this is great.

PR #1541, opened 2021-06-10
devsim: first step to be able to read GPUinfo database

christophe-lunarg:
- Preventing to read a devsim file with more data than expected is more restrictive than useful.
- Read every core feature available in the devsim file indendently from the schema

I didn't add all the extensions because I wasn't 100% sure it would affect things but I actually expect it would just work. More refactor would be required so that the "missing extension support" warnings are emitted only when the structures are not found.


jeremyk-lunarg (approved):
LGTM

PR #1547, opened 2021-06-15
[FOR REVIEW] devsim: Add GPUinfo file format as valid config

christophe-lunarg:
Added the ability for Devsim to read in values from the GPUinfo file
format.

Change-Id: Ie2e031f723dc56286f5fee763ab2324f634f965c

### DIFF
--- a/layersvt/device_simulation.cpp
+++ b/layersvt/device_simulation.cpp
@@ -1457,6 +1457,8 @@ class PhysicalDeviceData {
 
     VkPhysicalDeviceShaderDrawParametersFeatures physical_device_shader_draw_parameters_features_;
 
+    VkPhysicalDeviceSubgroupProperties physical_device_subgroup_properties_;
+
     // VK_KHR_8bit_storage structs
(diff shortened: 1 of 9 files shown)


layersvt/device_simulation.cpp
```diff
@@ -2555,6 +2471,12 @@ void JsonLoader::GetValue(const Json::Value &parent, const char *name, VkPhysica
         return;
     }
     DebugPrintf("\t\tJsonLoader::GetValue(VkPhysicalDeviceSamplerFilterMinmaxPropertiesEXT)\n");
+    if (!PhysicalDeviceData::HasExtension(&pdd_, VK_EXT_SAMPLER_FILTER_MINMAX_EXTENSION_NAME)) {
```
christophe-lunarg, line 2474: Could you explain this change, this message was moved which is something I thought should be done so great.

Should the value be loaded with GET_VALUE_WARN when not supported?

jeremyk-lunarg: It probably shouldn't.  I can have it return if the extension isn't supported.

PR #1558, opened 2021-06-18
devsim: Fix layer extension list replacement

jeremyk-lunarg:
Devsim was replacing not just the device extension list, but also each
individual layer's list of provided extensions.

This PR fixes this issue.


christophe-lunarg (approved):
This is beyond my understanding but looks good

PR #1559, opened 2021-06-22
devsim: Fix AndroidGetEnv use after free error

### LINKED ISSUE #1525: devsim: use after free in AndroidGetEnv (layersvt/device_simulation.cpp)
I tried to compile the VulkanTools for Android and got the following error:

```
jni/../../layersvt/device_simulation.cpp:386:16: error: address of stack memory associated with local variable 'android_env' returned [-Werror,-Wreturn-stack-address]
        return android_env.c_str();
               ^~~~~~~~~~~1 error generated.
```

Looking at the [relevant code](https://github.com/LunarG/VulkanTools/blob/f22adf1de2989be7f720bb455e6f87b7fcf8ed14/layersvt/device_simulation.cpp#L386) this looks like a genuine error: `android_env` is a local variable, and a pointer to its allocation is returned. Any read access after the function returns is a use-after-free from what I can tell.

jeremyk-lunarg:
AndroidGetEnv was returning the `c_str` of a staticly allocated
`std::string` as a `const char*`.

This `const char*` gets converted into an `std::string` anyway, so this
commit modifies the AndroidGetEnv function to return a `std::string`.

Fixes #1525

### DIFF
--- a/layersvt/device_simulation.cpp
+++ b/layersvt/device_simulation.cpp
@@ -404,7 +404,7 @@ struct ArrayCombinationModeSetting modifyPresentModes;
 const uint32_t MAX_BUFFER_SIZE = 255;
 typedef enum { VK_LOG_NONE = 0, VK_LOG_ERROR, VK_LOG_WARNING, VK_LOG_VERBOSE, VK_LOG_DEBUG } VkLogLevel;
 
-const char *AndroidGetEnv(const char *key) {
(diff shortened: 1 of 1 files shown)


jeremyk-lunarg:
@matled Does this fix the issue on your end?

matled:
@jeremy-lunarg Didn't get back to this earlier but just verified that I can build the project for Android now. Thanks!

PR #1560, opened 2021-06-22
Update to 1.2.182 headers

Tony-LunarG:


charles-lunarg (commented):
I see that these changes both update the header version and prevent the HUAWEI extension from being generated, but I don't see any code gen for the other extensions. Shouldn't this commit also have updated autogen?

These were the other extensions that were added in the 1.2.182 header update.
* VK_EXT_acquire_drm_display
* VK_EXT_multi_draw
* VK_EXT_physical_device_drm
* VK_NV_ray_tracing_motion_blur

Tony-LunarG:
I see 
inline void dump_head_vkCmdDrawMultiEXT(ApiDumpInstance& dump_inst, VkCommandBuffer commandBuffer, uint32_t drawCount, const VkMultiDrawInfoEXT* pVertexInfo, uint32_t instanceCount, uint32_t firstInstance, uint32_t stride) 
as the first new extension I looked for.  Is there something in particular you believe is missing?

charles-lunarg (approved):
D'Oh, I forgot that this repo doesn't check in the generated source code!

PR #1561, opened 2021-06-23
build: Fix missing override warnings

ncesario-lunarg:
I ran into some `-Winconsistent-missing-override` warnings when trying to build this on clang 12.0. I think we did something similar for the validation layers a while back.


Tony-LunarG (approved):
LGTM

PR #1563, opened 2021-06-28
devsim: Fix Surface Capabilities

jeremyk-lunarg:
Fixes issue where `VkSurfaceCapabilitiesKHR` struct was not filled out
properly.

### DIFF
--- a/layersvt/device_simulation.cpp
+++ b/layersvt/device_simulation.cpp
@@ -4008,6 +4008,19 @@ VKAPI_ATTR VkResult VKAPI_CALL GetPhysicalDeviceSurfaceCapabilitiesKHR(VkPhysica
         VkSurfaceCapabilitiesKHR surf_caps = pdd->surface_capabilities_;
         surf_caps.currentExtent = pSurfaceCapabilities->currentExtent;
         surf_caps.currentTransform = pSurfaceCapabilities->currentTransform;
+
+        if (surf_caps.minImageCount == 0) surf_caps.minImageCount = pSurfaceCapabilities->minImageCount;
+        if (surf_caps.maxImageCount == 0) surf_caps.maxImageCount = pSurfaceCapabilities->maxImageCount;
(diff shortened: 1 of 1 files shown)


layersvt/device_simulation.cpp
```diff
@@ -4008,6 +4008,19 @@ VKAPI_ATTR VkResult VKAPI_CALL GetPhysicalDeviceSurfaceCapabilitiesKHR(VkPhysica
         VkSurfaceCapabilitiesKHR surf_caps = pdd->surface_capabilities_;
         surf_caps.currentExtent = pSurfaceCapabilities->currentExtent;
         surf_caps.currentTransform = pSurfaceCapabilities->currentTransform;
+
+        if (surf_caps.minImageCount == 0) surf_caps.minImageCount = pSurfaceCapabilities->minImageCount;
+        if (surf_caps.maxImageCount == 0) surf_caps.minImageCount = pSurfaceCapabilities->maxImageCount;
```
lunarpapillo, line 4013: `surf_caps.minImageCount` => `surf_caps.maxImageCount`

lunarpapillo (changes requested):
Mostly looks good - there's one copy/paste miss.

This fixes the issues I was seeing when I was upgrading the `devsim` tests, yes?

layersvt/device_simulation.cpp
jeremyk-lunarg: Codegen could, but that is a later improvement.
This fixes the issue on my windows system.

lunarpapillo: I'm okay with deferring the issue of whether to codegen.  I'll create an issue.

lunarpapillo: https://github.com/LunarG/VulkanTools/issues/1564

I don't really know what I'm talking about, so if y'all could look at the issue and expand on it and correct any terminology I might have misused, I'd be appreciative. :-)

lunarpapillo (approved):
The fix looks fine to me.

christophe-lunarg (approved):

PR #1570, opened 2021-06-30
via: Fix single backslash json failure in settings file

MarkY-LunarG:
Summary:
Fix single backslash in settings file setting that causes JSON error
when outputting.  Replace all settings single backslashes with double.

### DIFF
--- a/via/via_system.cpp
+++ b/via/via_system.cpp
@@ -1555,6 +1555,23 @@ std::string ViaSystem::ConvertPathFormat(std::string str) {
     return str;
 }
 
+static void ReplaceAllBackslashes(std::string& str) {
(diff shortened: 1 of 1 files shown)


MarkY-LunarG:
@richard-lunarg, can you build and test this?  I only have a Linux box.

richard-lunarg:
Is ReplaceAllBackslashes only called on Windows? On all other platforms it would need to be a forward slash '/'. Windows will recognize the forward slashes in a path and most other paths in the json look more like this:
"D:/Windows/System32/vulkan-1.dll"
etc.

MarkY-LunarG:
It's called on all platforms for VK Layer settings since any '\' could cause issues for the generated JSON.

PR #1580, opened 2021-08-03
Update to 1.2.186 headers

Tony-LunarG:


charles-lunarg (commented):
One thing that looks odd is the lack of extensions in ApiDump. These are the extensions I found added from 183 to 186, and I know that present_wait adds a new function which should have dumping enabled.

  * VK_EXT_shader_atomic_float2
  * VK_HUAWEI_persistent_thread
  * VK_KHR_present_id
  * VK_KHR_present_wait
  * VK_NV_extermal_memory_rdma

Tony-LunarG:
Yes, but remember that the generated files in this repo are not checked in to git.  I do see dump_head_vkWaitForPresentKHR in api_dump.cpp

charles-lunarg (approved):
D'oh!
Forgot that this is the one repo that doesn't check in auto-gen.

PR #1586, opened 2021-08-11
[vkconfig] Add new loader debug options

MarkY-LunarG:
Add the new loader debug options to the VkConfig.  This eventually
should not be a combobox, but for now just add the options.


christophe-lunarg (approved):
Possible improvements:
- Change the combobox indeed to something else but indeed, this is not simple work..
- Handle the loader version to expose these new messages types only when the loader is newer than 1.2.182

PR #1588, opened 2021-08-11
[WIP] vkconfig: Only validate what's necessary for devsim

christophe-lunarg:
The goal of this PR is to make more efficient the use case of testing Vulkan application portability with Vulkan Configurator.

So replacing the the use of the "Standard Preset" by the following setting set in the Vulkan Configurator "Portability" built-in configuration.
image

This need review of the validation layer code to ensure no check is missing...


christophe-lunarg:
Replaced by https://github.com/LunarG/VulkanTools/pull/1667

PR #1589, opened 2021-08-11
devsim: Integrate vku library

jeremyk-lunarg:
Integrated new `vku` library to read layer settings.


build-android/jni/Android.mk
```diff
@@ -18,6 +18,7 @@ LOCAL_PATH := $(call my-dir)
 SRC_DIR := ../..
 LAYER_DIR := ../generated
 THIRD_PARTY := ../third_party
+VKU_DIR := ../third_party
```
lunarpapillo, line 21: this looks like it is not used?

Perhaps it should be `$(SRC_DIR)/vku` and used in both `LOCAL_SRC_FILES` and `LOCAL_C_INCLUDES` below.

lunarpapillo (changes requested):
One minor change, then consider it approved by me.

christophe-lunarg (approved):

christophe-lunarg:
This PR should be merge for SDK testing.

build-android/jni/Android.mk
christophe-lunarg: Good catch, I removed it.

PR #1591, opened 2021-08-13
Update to 1.2.188 headers

Tony-LunarG:
Also remove troublesome Latex math code


scripts/api_dump_generator.py
```diff
@@ -2370,8 +2370,10 @@ def __init__(self, rootNode, constants, aliases, parentName):
         self.lengthMember = False
         lengthString = rootNode.get('len')
         if lengthString is not None:
-            if "ename:" in lengthString:
+            if "ename:" in lengthString or "latexmath:" in lengthString:
                 lengthString = rootNode.get('altlen')
+                if (lengthString == "(rasterizationSamples + 31) / 32"):
+                    lengthString = "rasterizationSamples / 32 + (31 / 32)"
```
Tony-LunarG, line 2376: On second thought, I don't think this works for integer math

Tony-LunarG: rasterizationSamples + 31 >> 5

charles-lunarg: Whoops, thats right. Integer math doesn't do fractions. The bitshift indeed does work here, even if its even less clean. 
Would be excellent if there was a comment detailing why this is special cased and why a bitshift by 5 is needed.

charles-lunarg (approved):
Looks good.
Hacks are hacks but I feel this is better than trying to redo the entire autogen arch for a single field.

PR #1592, opened 2021-08-20
[WIP] devsim: Add new gtest test for devsim

jeremyk-lunarg:
Modified CMakeList.txt files to build Devsim gtest files.
Built basic utility functions to build future devsim tests.
Added basic Devsim test to test file input and modification.


jeremyk-lunarg:
Still ironing out 1 or 2 kinks in the code.
Just wanted to get it up for people to look at and comment on.

lunarpapillo (commented):
FWIW... I'm not a primary reviewer, but this looks good to me.

(I could suggest Python improvements, but they would be fairly pervasive and probably not worth the effort.)

lunarpapillo:
> @lunarpapillo Do you know why it's failing cloud CI?

I don't have any particular insights into the reported errors, e.g.:
```
[  FAILED  ] test_physical_device_main_property_modification.single_file (2 ms)
```
as I didn't write those tests and so have no particular expertise with them.  If you need another set of eyes, I can work with you on them... it would be nice if the GitHub Actions provided better output, maybe via the `--output-on-failure` script.

lunarpapillo:
Hmmm.  GitHub Actions looks like it already passes `--output-on-failure`.  Not sure where the output is actually going...

lunarpapillo:
You probably already know this, but the error message:
```
[ RUN      ] test_physical_device_main_property_modification.single_file
/home/runner/work/VulkanTools/VulkanTools/tests/devsim_gtests/test_physical_device_main_property_modification.cpp:44: Failure
Expected equality of these values:
  err 
    Which is: -9
  VK_SUCCESS
    Which is: 0
[  FAILED  ] test_physical_device_main_property_modification.single_file (2 ms) 
```
is being provoked by an assertion failure in the test:
```
    err = inst_builder.makeInstance();
    ASSERT_EQ(err, VK_SUCCESS);
```
The test looks like it really needs to provide more helpful information in case of failure.  I don't think I can really help any more than that, but if you think of something let me know. :-)

jeremyk-lunarg:
I'm mostly confused because the failing test runs twice.  The first time it succeeds, but it fails on the second time with a return code of `VK_ERROR_INCOMPATIBLE_DRIVER` when it tries to create and instance.

Looks like the second time it's running with MockICD.

lunarpapillo:
It's cloud CI, so there's no "real" driver.  Is it not getting MockICD both times?

PR #1601, opened 2021-09-02
README: specify to have qt installed to run vkconfig from tarball

johnzupin:
README for vkconfig doesn't specify that qt must be installed. This commit updates to include a message that qt must be installed.

### DIFF
--- a/vkconfig/README.md
+++ b/vkconfig/README.md
@@ -39,7 +39,8 @@ Vulkan Configurator is a graphical user interface (GUI) that may be launched fro
 
 The tool is distributed differently, depending on the platform:
 - Ubuntu packages: Upon installing the `lunarg-vkconfig` package, the tools will be available from the command line as `vkconfig`.
-- Linux tarball: The vkconfig executable comes pre-built. Run vkconfig on the command line. 
+- Linux tarball: The vkconfig executable comes pre-built; be sure Qt is installed to run it. Run vkconfig on the command line.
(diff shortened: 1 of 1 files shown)


vkconfig/README.md
```diff
@@ -39,7 +39,7 @@ Vulkan Configurator is a graphical user interface (GUI) that may be launched fro
 
 The tool is distributed differently, depending on the platform:
 - Ubuntu packages: Upon installing the `lunarg-vkconfig` package, the tools will be available from the command line as `vkconfig`.
-- Linux tarball: The vkconfig executable comes pre-built. Run vkconfig on the command line. 
+- Linux tarball: The vkconfig executable comes pre-built, be sure qt is installed to run it. Run vkconfig on the command line.
```
lunarpapillo, line 42: , => ;
qt => Qt

It would be nice to provide instructions or a link to where Qt can be downloaded and installed, but maybe that's outside the scope of this document.

lunarpapillo (changes requested):

vkconfig/README.md
Tony-LunarG: I'm too lazy to want to go look up the package name.  I'd want qt5-default here but if we advance to qt6 we'd have to change the docs. How much of a problem is that?

lunarpapillo (approved):

christophe-lunarg (approved):

PR #1606, opened 2021-09-28
 build: Update update_deps.py 

charles-lunarg:
CMake heavily prefers forward slashes, this commit makes the helper.cmake
file print forward slashes instead of backwards slashes on windows. Also updates Update_deps.py
to match VVL.


Tony-LunarG (approved):
LGTM

PR #1623, opened 2021-11-17
VkConfig: Separate application name from path.

MarkY-LunarG:
Create two separate fields for application name and executable path so
that we can store different calls to the same application with different arguments.
This is especially useful for calling Steam applications which are all triggered
by running the "steam" executable.


MarkY-LunarG:
@richard-lunarg, I can't test this on Apple.  Can you verify this works?

MarkY-LunarG:
Old dialog:
vkconfig_executables_old

New Dialog:
vkconfig_executables_new

MarkY-LunarG:
Changed "Application Name" to "Application" since you might put details there like "vkcube with count" or something like that. 
 Also, noticed that the main "Vulkan Application Launcher" window wasn't big enough and was missing the output log info.  I bumped up the size (since I added "Application") there and it's showing again.

MarkY-LunarG:
I also added some documentation about the application launch usage.

MarkY-LunarG:
Updated because I found a bug where I would toggle the applications in the main dialog and the executable path didnt update.  Fixed and re-running.

christophe-lunarg:
I am trying to pull the branch and test it on my machine but... I need to download Qt and strangely, it takes forever to download. (16 hours~) 

Started yesterday, 5 hours left...

christophe-lunarg (approved):
I pulled the branch on Windows, it seems to work just fine !

PR #1625, opened 2021-12-14
Implementation of vkconfig doc html|settings|markdown <layername>

davidlunarg:
This is an initial implementation if adding "doc html|settings <layername>" to the command line options of vkconfig.

May still want to make further changes to add this capability to the GUI, and may want to change the output so as to better integrate the generation of layer documentation files with documentation needed for SDK creation.

### DIFF
--- a/vkconfig_core/override.cpp
+++ b/vkconfig_core/override.cpp
@@ -139,9 +139,8 @@ bool WriteLayersOverride(const Environment& environment, const std::vector<Layer
 }
 
 // Create and write vk_layer_settings.txt file
-bool WriteSettingsOverride(const Environment& environment, const std::vector<Layer>& available_layers,
+bool WriteSettingsOverride(const std::vector<Layer>& available_layers,
                            const Configuration& configuration, const std::string& settings_path) {
-    (void)environment;
 
     assert(!settings_path.empty());
     assert(QFileInfo(settings_path.c_str()).absoluteDir().exists());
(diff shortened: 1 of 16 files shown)


vkconfig_core/override.cpp
```diff
@@ -182,19 +182,47 @@ bool WriteSettingsOverride(const Environment& environment, const std::vector<Lay
...
+                   }
+                   stream << " " << word.c_str();
+                   nchars += (word.size()+1);
+               }
+            }
+            stream << "\n";
+
+            // If feature has unmet dependency, output it but comment it out
+            if (!::CheckDependence(*meta, parameter.settings)) {
+                stream << "#";
```
christophe-lunarg, line 220: Nice!

christophe-lunarg (commented):
This is really great, I'll pull the branch to do further testing but your implementation is spendid.

christophe-lunarg:
```
$ ./vkconfig.exe doc
Missing 'doc' command argument...

Name
        'doc_html' - Command to create Vulkan layer doc files

Synopsis
        vkconfig doc html <layer_name>
        vkconfig doc settings <layer_name>

Description
        vkconfig doc html <layer_name>
                Create the html documentation file for the given layer.

        vkconfig doc settings <layer_name>
                Create the vk_layers_settings.txt file for the given layer.
```

davidlunarg:
Added commit 373b978.

Fixed doc_html in help text, fixed seg faults when not enough args to doc command, changed default location of output files to cwd, and added an optional 4th arg that specifies output dir.

The problem with:

`$ ./vkconfig.exe doc settings VK_LAYER_KHRONOS_validation
vkconfig: could not find layer VK_LAYER_KHRONOS_validation`

is that in order to create the settings file, the layer needs to be loaded. If vkconfig can't load the layer, it outputs an error message. I changed the error message to indicated that the layer could not be loaded.

davidlunarg:
Abandoning this pull request, will create a another.

PR #1631, opened 2022-01-11
api_dump: avoid creating ApiDumpSettings at ~ApiDumpInstance()

zhangyiwei:
`first_func_call_on_frame` can be false if no interested vulkan api call has been made yet. However, `settings()` would instantiate `dump_settings` if it was a `nullptr`.

### DIFF
--- a/layersvt/api_dump.h
+++ b/layersvt/api_dump.h
@@ -693,9 +693,11 @@ class ApiDumpInstance {
     }
 
     inline ~ApiDumpInstance() {
+        if (!dump_settings) return;
+
         if (!first_func_call_on_frame) settings().closeFrameOutput();
 
-        if (dump_settings != NULL) delete dump_settings;
+        delete dump_settings;
     }
 
     inline uint64_t frameCount() {


zhangyiwei:
@chrisforbes I was trying to leverage this layer but hit this.

charles-lunarg (commented):
This feels like `std::unique_ptr<>` should of been used instead. Its a bit silly to have this manual allocation/deletion when better tools exist. The rest of the file isn't afraid of using standard library types, no reason it can't be done here.

zhangyiwei:
> This feels like `std::unique_ptr<>` should of been used instead. Its a bit silly to have this manual allocation/deletion when better tools exist. The rest of the file isn't afraid of using standard library types, no reason it can't be done here.

That refactoring is a separate issue. The bug here is `settings()` will still instantiate a `dump_settings` even when `dump_settings` is already a `nullptr`, no matter `dump_settings` gets refactored to use unique_ptr or not.

charles-lunarg (approved):
Fair point. This PR does fix the issue. Not sure when or why the settings object is null, but best to not create one when it isn't needed.

PR #1636, opened 2022-01-28
Update to 1.3.204 headers

Tony-LunarG:


charles-lunarg (approved):
Changes look good (meaning the generated files I built locally)

Issue #1645, opened 2022-02-08
Screenshot layer not working when scaled format is selected


ziga-lunarg:
Taking a screenshot with the screenshot layer and selecting format `USCALED` or `SSCALED` there is no image created, even though the log in VkConfig says: `Screen Capture file is: C:\Users\nnn\Desktop/1.ppm`

But it works with all other formats.

This fails on nvidia and intel gpus for me, but works on amd.

PR #1659, opened 2022-03-03
Error out when screenshot layer can't create image in requested format

davidlunarg:
Print out the the name of the capture file iff the capture actually
succeeded.

Fixes https://github.com/LunarG/VulkanTools/issues/1645


Tony-LunarG (approved):
LGTM

PR #1662, opened 2022-03-04
[via] Support additive environment variables

MarkY-LunarG:
Support logging the additive environment variable information that
the loader now supports.


christophe-lunarg (approved):
I guess it looks ok but I quite don't know enough how via is working.

MarkY-LunarG:
@christophe-lunarg, can you or someone else run it on Windows, just to make sure there's no failures in normal usage?  @richard-lunarg, same thing on Mac?

PR #1664, opened 2022-03-10
Fix Android property for screenshot dir in screenshot_layer.md

davidlunarg:


lunarpapillo (approved):
Well, that was easy. :-)

Issue #1314, opened 2021-01-07
Screenshot Layer: 3DMark Wildlife replay issues on Android with screen capture


wbarnesSamsung:
This is a companion issue with https://github.com/LunarG/gfxreconstruct/issues/477.  When replaying Wildlife, if the screenshot layer is enabled, it crashes.  See link and below for further details.

From dustin-lunarg,

> This crash is happening in the screenshot layer.

********** Crash dump: **********
#00 0x000000000004f834 /data/local/debug/vulkan/libVkLayer_screenshot.so (screenshot::CreateDevice(VkPhysicalDevice_T*, VkDeviceCreateInfo const*, VkAllocationCallbacks const*, VkDevice_T**)+164) (BuildId: 7c1c1fc4050f44c5f1e1ed4feade5d8eb0318401)
                                                                          screenshot::CreateDevice(VkPhysicalDevice_T*, VkDeviceCreateInfo const*, VkAllocationCallbacks const*, VkDevice_T**)
                                                                          VulkanTools\build-android\jni/../../layersvt\screenshot.cpp:1115:47
#01 0x000000000001224c /system/lib64/libvulkan.so (vulkan::api::CreateDevice(VkPhysicalDevice_T*, VkDeviceCreateInfo const*, VkAllocationCallbacks const*, VkDevice_T**)+1392) (BuildId: e42b2cfc977d1600a2bc2e39acea3eb3)
#02 0x0000000000659244 /data/app/com.lunarg.gfxreconstruct.replay-UMe_gnycPWsemTBqqh21Pw==/base.apk!libgfxrecon-replay.so (offset 0x1380000) (gfxrecon::decode::VulkanReplayConsumerBase::OverrideCreateDevice(VkResult, gfxrecon::decode::PhysicalDeviceInfo*, gfxrecon::decode::StructPointerDecoder<gfxrecon::decode::Decoded_VkDeviceCreateInfo> const*, gfxrecon::decode::StructPointerDecoder<gfxrecon::decode::Decoded_VkAllocationCallbacks> const*, gfxrecon::decode::HandlePointerDecoder<VkDevice_T*>*)+1732) (BuildId: 258a9d61f0b58fcc9c1d2aaa14a087203b0067b2)
[...]

PR #1665, opened 2022-03-11
Add vkEnumeratePhysicalDeviceGroups to screenshot layer

davidlunarg:
The screenshot layer needs a list of all devices queried by the app.
It gets them by intercepting calls to VkGetEnumeratePhysicalDevices.
vkEnumeratePhysicalDeviceGroups was later added to the Vulkan API,
and it too allows apps to get a list of devices.  This commit adds
vkEnumeratePhysicalDeviceGroups to the screenshot
layer so the layer gets the list of devices queried using that API call.

Fixes Issue #1314

### DIFF
--- a/layersvt/screenshot.cpp
+++ b/layersvt/screenshot.cpp
@@ -1189,6 +1189,26 @@ VKAPI_ATTR VkResult VKAPI_CALL EnumeratePhysicalDevices(VkInstance instance, uin
     return result;
 }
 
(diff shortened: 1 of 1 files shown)


Tony-LunarG (approved):
LGTM - Did you try it with a modified cube?

davidlunarg:
Tested this with a modified cube.c that used vkEnumeratePhysicalDeviceGroups rather than vkEnumeratePhysicalDevices.

PR #1683, opened 2022-04-08
Update scripts to explicitly use python3

richard-lunarg:
Update scripts to explicitly call python3. This is needed for building on macOS version 12.3 or later, and is a better practice overall on other platforms.

### DIFF
--- a/update_external_sources.bat
+++ b/update_external_sources.bat
@@ -14,7 +14,7 @@ git submodule update --init --recursive
    echo.
    echo Building %JSONCPP_DIR%
    cd  "%JSONCPP_DIR%"
-   python amalgamate.py
+   python3 amalgamate.py
 
    if not exist %JSONCPP_DIR%\dist\json\json.h (
       echo.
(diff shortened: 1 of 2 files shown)


update_external_sources.bat
```diff
@@ -14,7 +14,7 @@ git submodule update --init --recursive
    echo.
    echo Building %JSONCPP_DIR%
    cd  "%JSONCPP_DIR%"
-   python amalgamate.py
+   python3 amalgamate.py
```
lunarpapillo, line 17: Windows doesn't have a `python3`.  There is a `py.exe` that takes `-2` or `-3` to select a version of Python.

CI just uses `python` (which has always been Python 3).

The approved generic call would be:
```
py -3 amalgamate.py
```
but it's feasible to just leave this `python`.

lunarpapillo (changes requested):

update_external_sources.bat
richard-lunarg: My windows machine has a python3...?
py or python will NOT work on macOS. Hence the problem.

PR #1707, opened 2022-06-02
Update to 1.3.216 headers

Tony-LunarG:


Tony-LunarG:
Obviously needs changes in dependent repos before CI will pass

mikes-lunarg (approved):

Issue #948, opened 2020-01-21
via: Misleading information about failed external tests


karl-lunarg:
Reference: https://vulkan.lunarg.com/issue/view/5e22e47886da8143aabe5854

VIA reports that it can't _find_ the executable, even when it _can_ find it and there is a failure on execution.

There should be separate error messages for not finding something and getting a failure when executing that something.

Or at least say that it can't find something OR had a problem running it.

davidlunarg:
This was fixed in pull request https://github.com/LunarG/VulkanTools/pull/1723

PR #1723, opened 2022-08-10
via: Report if failed test is due failure or if test is missing

davidlunarg:
Addresses issue https://github.com/LunarG/VulkanTools/issues/948

### DIFF
--- a/via/via_system.cpp
+++ b/via/via_system.cpp
@@ -2006,7 +2006,6 @@ ViaSystem::ViaResults ViaSystem::GenerateTestInfo(void) {
 
     BeginSection("External Tests");
     if (_found_sdk) {
-        bool found_exe = false;
         std::string cube_exe;
         std::string full_cmd;
         std::string path = "";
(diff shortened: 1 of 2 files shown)


via/via_system_linux.cpp
```diff
@@ -89,6 +90,8 @@ int ViaSystemLinux::RunTestInDirectory(std::string path, std::string test, std::
             err = chdir(orig_dir);
         }
     }
+    if (err_code && err_code != 1)
```
MarkY-LunarG, line 93: Nit: Wrap conditional code in braces.

via/via_system.cpp
```diff
@@ -2052,11 +2051,6 @@ ViaSystem::ViaResults ViaSystem::GenerateTestInfo(void) {
             full_cmd = cube_exe;
             full_cmd += " --c 100 --suppress_popups";
             int test_result = RunTestInDirectory(path, cube_exe, full_cmd);
-            if (test_result == 0) {
```
MarkY-LunarG, line 2055: Removing this will make it attempt both times to run cube, why not just re-add the if (test_result != 0) { continue; }?

MarkY-LunarG (changes requested):

via/via_system.cpp
davidlunarg: If the test failed, we want to output some indication of failure. The code in the loop after the continue does that.

Also, there's a PrintEndTable that is always executed after the loop is exited. If we do a continue here, the corresponding PrintBeginTable is not executed.

via/via_system_linux.cpp
davidlunarg: Commit 71c075d9e59071ceffffe55b6a56bf18c24a58ad addressed this.

MarkY-LunarG (approved):

Issue #829, opened 2019-06-12
Enhancement ideas for VIA on Windows


mikew-lunarg:
(found during Windows SDK 1.1.108 testing)
The "user experience" of VIA on Windows could be improved. 

What happens now:
1. a console window opens
1. vkcube windows will pop open and close
1. the console window closes.

Refinement/enhancement requests:
* Give some human-informative status messages in the console while vkcube is running.
* Keep the console window open, and report where the results file could be found.  Let the user close it.
* Open the .html file in a browser automatically.

These features could be implemented with a .bat script around the executable.

mikew-lunarg:
From https://gitlab.khronos.org/vulkan/Vulkan-SDK-Packaging/issues/59

mikew-lunarg:
Have notice that on different runs: 0, 1, or 2 vkcube windows will pop open and close.
Is the variable number of vkcube windows just a race condition, or indicating failures?
The vkcube runs (and possible error?) could be reported in the console.

MarkY-LunarG:
The number of VkCube windows is based on whether or not VkCube is found.  This is reported in the console window as well as any success/failure messages.

- If it fails to find VkCube, no window appears, and a warning message is written into the generated HTML file.
- If it finds VkCube, it will at least run once, ideally twice.
  - First run just attempts to run VkCube to make sure it executes
  - Second run attempts to run with Validation layers enabled.
  - Again, a failure at any point will be output to the console as well as written to the generated HTML file.

This is all documented in the VIA readme that is also included with every SDK: https://github.com/LunarG/VulkanTools/blob/master/via/README.md#running.

So, per your requests:

1. Give some human-informative status messages in the console while vkcube is running.
  - This is done in several ways as mentioned above.
2. Keep the console window open, and report where the results file could be found. Let the user close it.
  - This was deemed by us early on as too intrusive.  The generated HTML file was believed to be sufficient.
3. Open the .html file in a browser automatically.
  - Since this can be run remotely, we didn't believe this was an option.  Perhaps we can revisit this.

mikew-lunarg:
I think much of this can be addressed adding informative clues to a persistent command window:
- attempt and outcome of each vkcube run
- the location of the .html file.

Tested on my machine, the console is on-screen less than 7 seconds and much of that time under the vkcube windows.  Hard to tell if any useful info was there.

[...]

PR #1728, opened 2022-08-12
via: Display console on Windows until user presses enter

davidlunarg:
Make text in vkvia console window easier to read:

  - Changed vkcube window size from 500x500 to 250x250.

  - Windows only: Added a sleep before displaying the vkvia.html
    file so that user might see the output for a while before the
    browser with the vkvia.html file covers it up. Added a sleep
    after displaying the vkvia.html file so the user might be able
    to bring the console to the front and see the console output.

This fix addresses https://github.com/LunarG/VulkanTools/issues/829

Change-Id: I396aa61f66e3751cd4915bed838913d0ad7cc2dd

### DIFF
--- a/via/via_system.cpp
+++ b/via/via_system.cpp
@@ -253,7 +253,24 @@ ViaSystem::~ViaSystem() {
 #ifdef VIA_WINDOWS_TARGET
     if (_out_file_format == VIA_HTML_FORMAT) {
         // Open the html file in a browser
+        // Sleep 1 second before bringing up the browser so console window is visible for a little
+        // while before browser comes up and covers it.
+        Sleep(1000);
         ShellExecute(NULL, "open", _full_out_file.c_str(), NULL, NULL, SW_SHOWNORMAL);
+
+        // If we are running in a console, wait for user to to press enter before
+        // exitting. We do this so that the user can read the output.
+        // Note that running vkvia from a Windows CMD prompt or from a Cygwin shell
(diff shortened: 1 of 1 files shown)


via/via_system.cpp
```diff
@@ -253,7 +253,12 @@ ViaSystem::~ViaSystem() {
 #ifdef VIA_WINDOWS_TARGET
     if (_out_file_format == VIA_HTML_FORMAT) {
         // Open the html file in a browser
+        // Sleep 3 seconds before bringing up the browser so console window is visible for a little while
+        Sleep(3000);
         ShellExecute(NULL, "open", _full_out_file.c_str(), NULL, NULL, SW_SHOWNORMAL);
+        // Sleep another 7 seconds after bringing up the browser. The console probably won't be visible,
+        // it will be covered by browser. But it's available on the screen for a while for viewing if needed.
+        Sleep(7000);
```
charles-lunarg, line 261: I'm conflicted about the delays, as while they do make the console a bit more visible, expecting people to be able to read it in the allotted time is expecting too much. 
If we want to make sure its visible, why close the console at all? Or put another way, is the delayed closing meant to allow people to *read* whats going on, or simply see that its happening?

davidlunarg: Made changes to address this. Removed the sleep, and now via asks user to press enter before closing the console window.

charles-lunarg (approved):

Issue #1495, opened 2021-05-06
vkvia (portability). Provokes validation errors on mac


KarenGhavam-lunarG:
I accidentally ran `vkvia` on Mac with validation enabled:
```
$ vkvia
UNASSIGNED-khronos-validation-createinstance-status-message(INFO / SPEC): msgNum: -671457468 - Validation Information: [ UNASSIGNED-khronos-validation-createinstance-status-message ] Object 0: handle = 0x7fdf95815c00, type = VK_OBJECT_TYPE_INSTANCE; | MessageID = 0xd7fa5f44 | Khronos Validation Layer Active:
    Settings File: Found at /Users/papillo/.local/share/vulkan/settings.d/vk_layer_settings.txt specified by VkConfig application override.
    Current Enables: None.
    Current Disables: VK_VALIDATION_FEATURE_DISABLE_THREAD_SAFETY_EXT.

    Objects: 1
        [0] 0x7fdf95815c00, type: 1, name: NULL
UNASSIGNED-khronos-validation-createinstance-status-message(INFO / SPEC): msgNum: -671457468 - Validation Information: [ UNASSIGNED-khronos-validation-createinstance-status-message ] Object 0: handle = 0x7fdf9583a000, type = VK_OBJECT_TYPE_INSTANCE; | MessageID = 0xd7fa5f44 | Khronos Validation Layer Active:
    Settings File: Found at /Users/papillo/.local/share/vulkan/settings.d/vk_layer_settings.txt specified by VkConfig application override.
    Current Enables: None.
    Current Disables: VK_VALIDATION_FEATURE_DISABLE_THREAD_SAFETY_EXT.

    Objects: 1
        [0] 0x7fdf9583a000, type: 1, name: NULL
[...]

PR #1735, opened 2022-08-19
via: Address validation errors on macOS

davidlunarg:
Added portability extensions to vkCreateDevice and vkCreateInstance
calls so that when via is run from vkconfig, validation errors
are not generated.

Addresses issue https://github.com/LunarG/VulkanTools/issues/1495

### DIFF
--- a/via/via_system.cpp
+++ b/via/via_system.cpp
@@ -459,13 +459,28 @@ ViaSystem::ViaResults ViaSystem::GenerateInstanceInfo(void) {
         }
     }
 
-    const char* portability_instance_extension_list[] = {VK_KHR_PORTABILITY_ENUMERATION_EXTENSION_NAME};
+    // Need to specify VK_KHR_portability_enumeration extension on Mac in vkCreateInstance. This
+    // is because the Vulkan on Metal driver is not fully Vulkan conformant, and VK_KHR_portability
+    // needs to be enabled in order for the Vulkan Loader to report the Vulkan device.
+
+    // Also need to specify the VK_KHR_get_physical_device_properties2 extension because we will be
+    // be using the VK_KHR_portability_subset extension in vkCreateDevice.
+
+    std::vector<const char *> portability_instance_extension_list;
+    inst_info.flags = 0;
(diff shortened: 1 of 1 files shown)


via/via_system.cpp
```diff
@@ -459,13 +459,21 @@ ViaSystem::ViaResults ViaSystem::GenerateInstanceInfo(void) {
...
         if (strcmp(extension.extensionName, VK_KHR_PORTABILITY_ENUMERATION_EXTENSION_NAME) == 0) {
+            portability_instance_extension_list.push_back(VK_KHR_PORTABILITY_ENUMERATION_EXTENSION_NAME);
+            inst_info.ppEnabledExtensionNames = portability_instance_extension_list.data();
             inst_info.flags = VK_INSTANCE_CREATE_ENUMERATE_PORTABILITY_BIT_KHR;
-            inst_info.enabledExtensionCount = 1;
-            inst_info.ppEnabledExtensionNames = portability_instance_extension_list;
-            break;
+            inst_info.enabledExtensionCount++;
+        }
+        if (strcmp(extension.extensionName, VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2_EXTENSION_NAME) == 0) {
```
MarkY-LunarG, line 473: Why did you add device properties 2 here?  I thought this change focused on portability?

davidlunarg: Maybe a comment might be in order here explaining why...

On Mac, if you specify VK_KHR_PORTABILITY_ENUMERATION_EXTENSION_NAME without specifying VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2_EXTENSION_NAME, validation produces an error.

charles-lunarg: Not `VK_KHR_portability_enumeration` specifically, but `VK_KHR_portability_subset`  (as the latter is a device extension and needs `vkGetPhysicalDeviceProperties2KHR`).

davidlunarg: Force-pushed a change adding comments.

charles-lunarg (approved):

Issue #1737, opened 2022-08-25
Validation error in screenshot layer


davidlunarg:
If vkcube is run with validation and screenshot enabled, and the output format is set to UINT, the following validation error message is displayed:

VUID-vkCmdBlitImage-srcImage-00230(ERROR / SPEC): msgNum: 569695033 - Validation Error: [ VUID-vkCmdBlitImage-srcImage-00230 ] Object 0: handle = 0x1fa202e0380, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0x21f4db39 | vkCmdBlitImage: If one of srcImage and dstImage images has unsigned integer format, the other one must also have unsigned integer format.  Source format is VK_FORMAT_B8G8R8A8_UNORM Destination format is VK_FORMAT_R8G8B8A8_UINT. The Vulkan spec states: If either of srcImage or dstImage was created with an unsigned integer VkFormat, the other must also have been created with an unsigned integer VkFormat (https://vulkan.lunarg.com/doc/view/1.3.224.0/windows/1.3-extensions/vkspec.html#VUID-vkCmdBlitImage-srcImage-00230)

This issue was originally submitted as https://gitlab.khronos.org/vulkan/Vulkan-SDK-Packaging/-/issues/1112 by @ziga-lunarg

juan-lunarg:
Closed by @davidlunarg in https://github.com/LunarG/VulkanTools/pull/1738/files

juan-lunarg:
Verified this is fixed by @davidlunarg 's PR by running validation with screenshot layer:
```
Selected GPU 0: Apple M2 Max, type: IntegratedGpu

screenshot: Incompatible output format requested, changing output format to VK_FORMAT_B8G8R8A8_UNORM

screenshot: Capture file is: /Users/juanramos/VulkanSDK/1.ppm 

Process terminated
```

juan-lunarg:
I updated the SDK screenshot layer testing regardless though to ensure this stays fixed.

PR #1738, opened 2022-08-25
Address validation errors in screenshot layer.

davidlunarg:
Validation error would occur when VK_SCREENSHOT_FORMAT was set
to SINT/UINT. Added code to override the output format for this
case and set it to the same as the swapchain image format.

Also changed wording in several error messages.

Fixes issue https://github.com/LunarG/VulkanTools/issues/1737

### DIFF
--- a/layersvt/screenshot.cpp
+++ b/layersvt/screenshot.cpp
@@ -265,7 +265,7 @@ void readScreenShotFormatENV(void) {
                                 vk_screenshot_format);
 #else
             fprintf(stderr,
-                    "Selected format:%s\nIs NOT in the list:\nUNORM, SNORM, USCALED, SSCALED, UINT, SINT, SRGB\n"
+                    "screenshot: Selected format:%s\nIs NOT in the list:\nUNORM, SNORM, USCALED, SSCALED, UINT, SINT, SRGB\n"
                     "Swapchain Colorspace will be used instead\n",
                     vk_screenshot_format);
 #endif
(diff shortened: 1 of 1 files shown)


layersvt/screenshot.cpp
```diff
@@ -683,6 +683,25 @@ static bool writePPM(const char *filename, VkImage image1) {
...
 
+    // From vulkan spec:
+    //   VUID-vkCmdBlitImage-srcImage-00229
+    //     If either of srcImage or dstImage was created with a signed integer VkFormat,
+    //     the other must also have been created with a signed integer VkFormat
+    //   VUID-vkCmdBlitImage-srcImage-00230
+    //     If either of srcImage or dstImage was created with an unsigned integer VkFormat,
+    //     the other must also have been created with an unsigned integer VkFormat
+    // If the destination format is not compatible, set destintation format to source format and print a warning.
+    if (FormatIsSINT(format) || FormatIsSINT(destformat) || FormatIsUINT(format) || FormatIsUINT(destformat)) {
```
charles-lunarg, line 694: How does this make sure that the src/dst signed-ness is the same? What I see is that if the format is signed or unsigned or if the destFormat is signed or unsigned, make the destFormat the same as the format.

davidlunarg: Went through all the possible case with Charles, and yes, setting destFormat to format is correct and insures consistent signed-ness. But his comments made me realize there is one problem here: The warning message is printed even when the destFormat is not changed. I'll need to fix that.

charles-lunarg (approved):
LGTM

ziga-lunarg (approved):

Tony-LunarG (approved):
LGTM

Issue #1740, opened 2022-09-07
GCC 12: cannot compile due to Werror=maybe-uninitialized in jsoncpp


MayeulC:
I am trying to build LunarG's Vulkan SDK, and encountering the same issue as that one: https://vulkan.lunarg.com/issue/view/6309e0835df1129602e803eb

The compiler warns that a variable may be uninitialized, and then fails the compilation (`[-Werror=maybe-uninitialized]`) as all warnings are treated as errors.

I'm compilling on vanilla Arch Linux, GCC 12.2.0.

<details><summary>Compilation log></summary>

```
[  6%] Building CXX object via/CMakeFiles/vkvia.dir/__/submodules/jsoncpp/dist/jsoncpp.cpp.o
In file included from /usr/include/c++/12.2.0/bits/new_allocator.h:36,
                 from /usr/include/c++/12.2.0/x86_64-pc-linux-gnu/bits/c++allocator.h:33,
                 from /usr/include/c++/12.2.0/bits/allocator.h:46,
                 from /usr/include/c++/12.2.0/string:41,
                 from /vulkansdk/1.3.224.1/source/VulkanTools/submodules/jsoncpp/dist/json/json.h:127,
                 from /vulkansdk/1.3.224.1/source/VulkanTools/submodules/jsoncpp/dist/jsoncpp.cpp:76:
In function 'std::_Require<std::__not_<std::__is_tuple_like<_Tp> >, std::is_move_constructible<_Tp>, std::is_move_assignable<_Tp> > std::swap(_Tp&, _Tp&) [with _Tp = Json::Value::ValueHolder]',
    inlined from 'void Json::Value::swapPayload(Json::Value&)' at /vulkansdk/1.3.224.1/source/VulkanTools/submodules/jsoncpp/dist/jsoncpp.cpp:2967:12,
    inlined from 'bool Json::Reader::readValue()' at /vulkansdk/1.3.224.1/source/VulkanTools/submodules/jsoncpp/dist/jsoncpp.cpp:433:31:
/usr/include/c++/12.2.0/bits/move.h:205:7: error: 'v.Json::Value::value_' may be used uninitialized [-Werror=maybe-uninitialized]
  205 |       __a = _GLIBCXX_MOVE(__b);
      |       ^~~
/vulkansdk/1.3.224.1/source/VulkanTools/submodules/jsoncpp/dist/jsoncpp.cpp: In member function 'bool Json::Reader::readValue()':
[...]

PR #1741, opened 2022-09-07
build: Disable -Werror for vkvia and vkconfig_core

mikes-lunarg:
GCC 12 emits -Werror=maybe-uninitialized when compiling the jsoncpp dependency with -Werror. The least invasive solution is to just disable -Werror for vkvia and vkconfig_core which use jsoncpp.cpp.

### DIFF
--- a/via/CMakeLists.txt
+++ b/via/CMakeLists.txt
@@ -19,6 +19,8 @@ if(WIN32)
 
 else()
     if (CMAKE_COMPILER_IS_GNUCC OR CMAKE_C_COMPILER_ID MATCHES "Clang")
+        # Disable -Werror because of warnings with jsoncpp and GCC 12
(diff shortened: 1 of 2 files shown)


mikes-lunarg:
Fixes #1740

christophe-lunarg:
I am personnally not very entousiastic about this fix. Maybe we should report the issue to JSONCPP instead and update our JSONCPP dependences to a fixed build.

christophe-lunarg:
I created an alternative fix : https://github.com/LunarG/VulkanTools/pull/1742

christophe-lunarg:
@mikes-lunarg, so I am guessing you can close this issue.

mikes-lunarg:
Fixed in #1742

PR #1742, opened 2022-09-12
build: Disable maybe-uninitialized GCC warning for vkvia and vkconfig core

christophe-lunarg:
GCC 12 emits -Werror=maybe-uninitialized when compiling the jsoncpp dependency with -Werror. The least invasive solution is to just disable -Wmaybe-uninitialized for vkvia and vkconfig_core which use jsoncpp.cpp.

This is an alternative fix of :
https://github.com/LunarG/VulkanTools/pull/1741


mikes-lunarg:
clang doesn't recognized `-Wno-maybe-uninitialized` so it either needs to go in a gcc-only section or we need to add `-Wno-unknown-warning-option` along with it.

christophe-lunarg:
@mikes-lunarg, I updated my commit to disable `-Wno-unknown-warning-option` on Clang.

Let me know what you think about this.

mikes-lunarg:
> Let me know what you think about this.

Sorry for the confusion, my suggestion was one OR the other. In other words, you only need to pass `-Wno-unknown-warning-option` to clang if you are also giving it `-Wno-maybe-uninitialized` (a GCC-only flag). Since you put `-Wno-maybe-uninitialized` in the GCC-only path, you don't need `-Wno-unknown-warning-option`

christophe-lunarg:
Ohhhhh, I see!

christophe-lunarg:
@mikes-lunarg updated !

mikes-lunarg (approved):
LGTM

Issue #1744, opened 2022-09-12
vkconfig clang warning


mikes-lunarg:
I was compiling vkconfig with clang and got what looks like a legitimate warning:
```
/home/mike/Projects/VulkanTools/vkconfig_core/parameter.cpp:85:33: error: use of logical '&&' with constant operand [-Werror,-Wconstant-logical-operand]
        if (parameters[i].state && PARAMETER_RANK_EXCLUDED) continue;
                                ^  ~~~~~~~~~~~~~~~~~~~~~~~
/home/mike/Projects/VulkanTools/vkconfig_core/parameter.cpp:85:33: note: use '&' for a bitwise operation
        if (parameters[i].state && PARAMETER_RANK_EXCLUDED) continue;
                                ^~
                                &
/home/mike/Projects/VulkanTools/vkconfig_core/parameter.cpp:85:33: note: remove constant to silence this warning
        if (parameters[i].state && PARAMETER_RANK_EXCLUDED) continue;
                               ~^~~~~~~~~~~~~~~~~~~~~~~~~~
/home/mike/Projects/VulkanTools/vkconfig_core/parameter.cpp:86:33: error: use of logical '&&' with constant operand [-Werror,-Wconstant-logical-operand]
        if (parameters[i].state && PARAMETER_RANK_MISSING) continue;
                                ^  ~~~~~~~~~~~~~~~~~~~~~~
/home/mike/Projects/VulkanTools/vkconfig_core/parameter.cpp:86:33: note: use '&' for a bitwise operation
        if (parameters[i].state && PARAMETER_RANK_MISSING) continue;
                                ^~
                                &
/home/mike/Projects/VulkanTools/vkconfig_core/parameter.cpp:86:33: note: remove constant to silence this warning
        if (parameters[i].state && PARAMETER_RANK_MISSING) continue;
                               ~^~~~~~~~~~~~~~~~~~~~~~~~~
```
```
$ clang --version
clang version 14.0.6
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
```

The logical && really doesn't make any sense there, I wonder if the following might be the intended behavior?
```diff
diff --git a/vkconfig_core/parameter.cpp b/vkconfig_core/parameter.cpp
index 8900495ce..577046f60 100644
--- a/vkconfig_core/parameter.cpp
+++ b/vkconfig_core/parameter.cpp
[...]

PR #1745, opened 2022-09-13
Fix clang CI

mikes-lunarg:
Two minor and related issues fixed here:
- #1744 
- The GitHub Actions clang build was using gcc instead of clang, so the first issue wasn't detected in CI.


christophe-lunarg:
The `ComputeMinApiVersion` fix is not quite right but that's a good catch. I added unit tests to ensure the implementation is right as part of this PR:
https://github.com/LunarG/VulkanTools/pull/1743

I also included your C.I. fix so that you don't have to modify your PR.

PR #1748, opened 2022-10-03
screenshot: Use std::mutex

charles-lunarg:
Previously, screenshot was using the mutexes from an accidentally exposed header file provided by the Validation Layers. Since they aren't installing that header, the definitions were missing when building an SDK.

The fix is to replace the use of platform specific mutex wrapper with std::mutex.


richard-lunarg (approved):
Pretty straightforward, looks good to me.

PR #1749, opened 2022-10-04
monitor: Remove vk_loader_platform.h header

charles-lunarg:
Unused header so it should be removed.

### DIFF
--- a/layersvt/monitor.cpp
+++ b/layersvt/monitor.cpp
@@ -1,9 +1,9 @@
 /*
  * Vulkan
  *
- * Copyright (C) 2016 Valve Corporation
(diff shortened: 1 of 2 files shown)


richard-lunarg (approved):
Thanks for burning some midnight oil on this!

charles-lunarg:
I found the source of the 'weird CI error'
https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/a4fc6748d90dd623c03e278476e9676c93f420dc/layers/generated/lvt_function_pointers.h#L28

The gist is the CreateEvent and CreateSemaphore are names in the winbase headers that clash with vulkan. The vk_loader_platform.h had protection against this and so by removing that header, the protections were gone.

PR #1752, opened 2022-10-11
layer_factory: Guard against win32 defines

charles-lunarg:
CreateEvent and CreateSemaphore are macros defined in some windows header file, and the Validation layers moved this guard in such a way that the layer factory no longer is protected from it. The simplest fix is to manually include vulkan, which will include various windows headers, undefine the offending names, then include the layer_dispatch_table (which contains the macro defined names).


Tony-LunarG (approved):
LGTM

PR #1762, opened 2022-11-15
scripts: use python3 if python is not available

lunarpapillo:
Some modern Linux distributions come with the "python3" executable but not the "python" executable.  This change allows whichever is present on the invoking machine to be used.

### DIFF
--- a/update_external_sources.sh
+++ b/update_external_sources.sh
@@ -1,12 +1,14 @@
 #!/bin/bash
 
+PYTHON=$(which python python3 | head -1)
+
 set -e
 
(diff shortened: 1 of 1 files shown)


lunarpapillo:
This provides clean builds with **Ubuntu 22.04** and allows the CI migration (replacing **Ubuntu 16.04** machines with **Ubuntu 22.04** machines) to continue.

richard-lunarg (commented):
This is really great, we have the same problem on macOS now as well.

christophe-lunarg (approved):

PR #1769, opened 2022-11-28
apidump: Fix whitespace between names & types

charles-lunarg:
setw() will reset back to 0 in many cases, one of which includes << This meant that call to `stream << setw();` does not apply the proper whitespace required, leading to not adding any space between the name and type. `setw() << "";` forces the correct width to be printed.

### DIFF
--- a/layersvt/api_dump.h
+++ b/layersvt/api_dump.h
@@ -441,6 +441,11 @@ class ApiDumpSettings {
             stream() << "[\n";
         }
 
+        if (use_spaces)
+            stream() << std::setfill(' ');
+        else
+            stream() << std::setfill('\t');
+
         if (isFrameInRange(0)) {
             setupInterFrameOutputFormatting(0);
         }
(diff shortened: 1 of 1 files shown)


christophe-lunarg (approved):
Looks good to me.

layersvt/api_dump.h
```diff
@@ -519,18 +524,20 @@ class ApiDumpSettings {
 
     std::ostream &formatNameType(std::ostream &stream, int indents, const char *name, const char *type) const {
         stream << indentation(indents) << name << ": ";
-
         if (use_spaces)
-            stream << std::setfill(' ') << std::setw(name_size - (int)strlen(name) - 2);
+            stream << std::setw(name_size - (int)strlen(name) - 2) << "";
```
lunarpapillo, line 528: Your commit message explains this construction, but it's likely a future engineer won't review the commit messages to figure out why this "needless" construction is done.  Add a comment here explaining why the unusual construction is used.

layersvt/api_dump.h
```diff
@@ -441,6 +441,11 @@ class ApiDumpSettings {
             stream() << "[\n";
         }
 
+        if (use_spaces)
+            stream() << std::setfill(' ');
+        else
+            stream() << std::setfill('\t');
+
```
lunarpapillo, line 448: This seems unrelated to the commit comment.  It's also clearly a behavioral change; its correctness is unclear from context.  Please explain.

lunarpapillo (changes requested):

layersvt/api_dump.h
charles-lunarg: Check the api dump refactor PR - this commit is in that and I address that issue there (with a comment)

layersvt/api_dump.h
charles-lunarg: Also added comment about that in the other PR.

charles-lunarg:
Closing this because #1771 has this commit and fixes some of the mistakes with it.

PR #1770, opened 2022-11-29
android: limit parallel builds

lunarpapillo:
The Android build is quite aggressive, attempting to consume all available cores; large builds (particularly the api_dump layer) cause all the processes involved to consume 3+G of memory, causing physical and swap memory to be exhausted, which causes sporadic build failures as the OOM killer kills the compiler processes.

These changes allow the use of an environment variable $VT_BUILD_CORES to limit how many cores are in use; it also prevents the build from using additional parallel jobs if the load average is too high (on one 8-core system, the load average climbed to over 26 before the OOM killer started killing processes).

### DIFF
--- a/build-android/build_all.sh
+++ b/build-android/build_all.sh
@@ -29,6 +29,7 @@ if [[ $(uname) == "Linux" ]]; then
 elif [[ $(uname) == "Darwin" ]]; then
     cores=$(sysctl -n hw.ncpu) || echo 4
 fi
+use_cores=${VT_BUILD_CORES:-$cores}
 
 function findtool() {
     if [[ ! $(type -t $1) ]]; then
@@ -66,7 +67,7 @@ function create_APK() {
 #
 ./update_external_sources_android.sh --no-build
 ./android-generate.sh
-ndk-build -j $cores
+ndk-build -j $use_cores -l $use_cores
 
 #
 # build VulkanLayerValidationTests APK


lunarpapillo:
The build mentioned above succeeded without overloading the system; I monitored the build system, and the load average never exceeded 9 (on an 8-core system): http://erusea:8080/job/VulkanTools/2395/BITS=64,BUILD_MODE=Release,label=Linux-Android/console
```
About to execute './build_all.sh' in directory /home/lunarg/.jenkins/vt/Release64/VulkanTools/build-android
duration: 15.68 minutes (940.6286418437958 seconds)
Archiving artifacts
...
Finished: SUCCESS
```

Tony-LunarG (approved):
LGTM

charles-lunarg (commented):
What happens if `VT_BUILD_CORES` isn't defined?
Would that break the build_all.sh script?

lunarpapillo:
> What happens if `VT_BUILD_CORES` isn't defined? Would that break the build_all.sh script?

You must seek harmony with your `bash`-*fu*, grasshopper... ;-)

The `${VAR:-XXX}` syntax evaluates to `XXX` if `$VAR` is unset or null.

charles-lunarg:
> The `${VAR:-XXX}` syntax evaluates to `XXX` if `$VAR` is unset or null.

TIL! Thanks for explaining that bit of wacky syntax to me :D

charles-lunarg (approved):

PR #1771, opened 2022-11-29
Refactor api dump

charles-lunarg:
Big refactor to the codebase which cleans up many aspects of the code, to reduce the code generation time and compile time.

Changes include:
* Removing variadic templates, replacing them with explicit storage in the ApiDumpInstance class. There are only two places in the entire API where this is needed, making everything require a variadic template to handle these two cases is a bit much.
* Removing redundant `inline` statements for class member functions
* Removing the return type for many of the api_dump.h functions. Returning the stream operator was used in one place and one place only, and didn't require returning the ostream type to function.
* Factor out the dump_head functions. The actual values printed are easily replaced with a single function that takes the function name & return type as parameters, removing large amounts of generated code.
* Remove runtime JSON comma printing - modify the codegen to put commas where necessary instead of keeping a boolean value to keep track of things.
* Fix unknown structures in the pNext chain being incorrectly quoted in JSON output
* Make the type parsing handle the case where the name of the function parameter/struct member is a substring of the type, causing incorrect printing of the type.
* Drastically improve the performance of the codegen by using dictionaries instead of sets. This is because the codegen previously would linearly search through the entire list when looking for a single element. Now lookups are just a O(1) check into a dictionary.

### DIFF
--- a/layersvt/api_dump.h
+++ b/layersvt/api_dump.h
@@ -1,6 +1,6 @@
-/* Copyright (c) 2015-2019 The Khronos Group Inc.
- * Copyright (c) 2015-2019 Valve Corporation
- * Copyright (c) 2015-2019 LunarG, Inc.
+/* Copyright (c) 2015-2023 The Khronos Group Inc.
+ * Copyright (c) 2015-2023 Valve Corporation
+ * Copyright (c) 2015-2023 LunarG, Inc.
  * Copyright (C) 2015-2016 Google Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -234,7 +234,7 @@ class ConditionalFrameOutput {
 
 class ApiDumpSettings {
    public:
-    ApiDumpSettings() {
+    ApiDumpSettings() : output_stream(std::cout.rdbuf()) {
         std::string filename_string = "";
         // If the layer settings file has a flag indicating to output to a file,
         // do so, to the appropriate filename.
@@ -259,15 +259,8 @@ class ApiDumpSettings {
         }
         // If one of the above has set a filename, open the file as an output stream.
         if (!filename_string.empty()) {
-            use_cout = false;
-            output_stream.open(filename_string, std::ofstream::out | std::ostream::trunc);
-            size_t last_slash_idx = filename_string.find_last_of("\\/");
-            if (std::string::npos != last_slash_idx) {
-                output_dir = filename_string.substr(0, last_slash_idx + 1);
-            }
-        } else {
-            // Otherwise, fallback to cout only
-            use_cout = true;
+            output_file_stream.open(filename_string, std::ofstream::out | std::ostream::trunc);
+            output_stream.rdbuf(output_file_stream.rdbuf());
         }
 
         // Get the remaining settings (some we also want to provide the ability to override
@@ -310,6 +303,7 @@ class ApiDumpSettings {
         }
 
         indent_size = std::max(readIntOption("lunarg_api_dump.indent_size", 4), 0);
+        tab_size = indent_size;
         show_type = readBoolOption("lunarg_api_dump.show_types", true);
         name_size = std::max(readIntOption("lunarg_api_dump.name_size", 32), 0);
         type_size = std::max(readIntOption("lunarg_api_dump.type_size", 0), 0);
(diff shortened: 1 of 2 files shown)


lunarpapillo:
Failure #2440 seems to be a system issue.  I'll fix it and then restart.

richard-lunarg (approved):
Massive cleanup. Good work.

layersvt/api_dump.h
```diff
@@ -833,389 +851,359 @@ class ApiDumpInstance {
...
-template <typename T, typename... Args>
-inline void dump_text_value(const T object, const ApiDumpSettings &settings, const char *type_string, const char *name, int indents,
-                            std::ostream &(*dump)(const T, const ApiDumpSettings &, int, Args... args), Args... args) {
+template <typename T>
+void dump_text_value(const T object, const ApiDumpSettings &settings, const char *type_string, const char *name, int indents,
+                     void (*dump)(const T, const ApiDumpSettings &, int)) {
     settings.formatNameType(settings.stream(), indents, name, type_string);
-    dump(object, settings, indents, args...) << "\n";
+    dump(object, settings, indents);
+    settings.stream() << "\n";
```
lunarpapillo, line 973: not really your code, but here and elsewhere, why `"\n"` instead of `std::endl`?

layersvt/api_dump.h
```diff
@@ -833,389 +851,359 @@ class ApiDumpInstance {
...
+    bool is_dynamic_scissor;
+    bool is_dynamic_viewport;
 };
 
 // Utility to output an address.
 // If the quotes arg is true, the address is encloded in quotes.
 // Used for text, html, and json output.
-void OutputAddress(const ApiDumpSettings &settings, const void *addr, bool quotes) {
-    if (quotes) settings.stream() << "\"";
+void OutputAddress(const ApiDumpSettings &settings, const void *addr) {
```
lunarpapillo, line 875: Really not your code... but (almost?) everywhere I look, callers of `OutputAddress` first do a special case to check the address against `NULL`, and then typically short-circuit.

It's confusing to have both callers checking the address against `NULL` and also checking it here.  Some refactoring of this would be welcome.

layersvt/api_dump.h
```diff
@@ -833,389 +851,359 @@ class ApiDumpInstance {
...
+
+    if (!dump_inst.firstFunctionCallOnFrame()) settings.stream() << ",\n";
+
+    // Display api call name
+    settings.stream() << settings.indentation(2) << "{\n";
+    settings.stream() << settings.indentation(3) << "\"name\" : \"" << funcName << "\",\n";
+
+    // Display thread info
+    if (settings.showThreadAndFrame()) {
+        { settings.stream() << settings.indentation(3) << "\"thread\" : \"Thread " << dump_inst.threadID() << "\",\n"; }
```
lunarpapillo, line 1181: does this also emit the frame?

lunarpapillo (approved):
Lots of joy here...

Don't forget to change the copyright year. :-)

I have some questions about the code, but found no issues suggesting that the code needed changes.  I'm marking **Approve**, but you should make sure you have a developer review too.

layersvt/api_dump.h
```diff
@@ -515,22 +520,23 @@ class ApiDumpSettings {
...
-    inline ApiDumpFormat format() const { return output_format; }
+    ApiDumpFormat format() const { return output_format; }
 
-    std::ostream &formatNameType(std::ostream &stream, int indents, const char *name, const char *type) const {
+    void formatNameType(std::ostream &stream, int indents, const char *name, const char *type) const {
         stream << indentation(indents) << name << ": ";
-
         if (use_spaces)
-            stream << std::setfill(' ') << std::setw(name_size - (int)strlen(name) - 2);
+            stream << std::setw(name_size - (int)strlen(name) - 2) << "";
```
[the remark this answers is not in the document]
charles-lunarg: Its a case of `std::setw()` not actually 'applying' when the semicolon is reached. 
This should have a comment because I realize how weird that is to see.

layersvt/api_dump.h
charles-lunarg: Looking at all the places OutputAddress is called, 7 places are guaranteed to be non-null, 4 are null, and 5 are unknowable at the call site. I see what you mean that they appear to be unnecessary null checks, but they are indeed necessary (in some places that is).

layersvt/api_dump.h
charles-lunarg: std::endl does _two_ things, add a new line and flush the output. I don't want to flush output unless explicitly asked for by the user, which is a separate option. Using "\n" allows for better performance in the general case cause it causes iostreams to only flush when necessary, and not on every line.

layersvt/api_dump.h
charles-lunarg: The showThreadAndFrame refers to the current thread and current frame number. Both HTML and JSON output encode the 'frame' number in the parent objects, whereas what this is printing is inside the child object.

layersvt/api_dump.h
```diff
@@ -441,6 +441,11 @@ class ApiDumpSettings {
             stream() << "[\n";
         }
 
+        if (use_spaces)
```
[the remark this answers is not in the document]
charles-lunarg: This was intentional because setfill stays active for basically the duration of the stream, and isn't invalidated. This prevents the need from calling setfill every time a fill is required. 

However, it would be prudent to determine if that behavior is 1. guaranteed (I read a bit of cppreference but have forgotten the specifics) and 2. whether the output of api_dump maintains space/tab consistency.

PR #1772, opened 2022-11-29
Update to 1.3.235 headers (and C++ 17)

Tony-LunarG:
VVL components upstream have gone to C++ 17 so compile layers that use them accordingly
Disable VK_EXT_descriptor_buffer until VVL upstream fixes codegen


juan-lunarg (approved):
Change looks good to me. The CMake can be cleaned up but that should be done in another PR.

charles-lunarg (commented):
~~Where is the corresponding update to known_good.json?~~

Ignore that, I was being blind.

charles-lunarg (approved):
Sorry for the confusion, these changes look good.

PR #1785, opened 2023-01-17
github: Rename master to main

juan-lunarg:


Tony-LunarG:
I thought we were only doing this in Khronos repos

juan-lunarg:
> I thought we were only doing this in Khronos repos

Agreed there isn't a need to do it here. It's just about consistency. Most of the people who contribute here work on Khronos repos where the default will be main.

juan-lunarg:
To be clear I don't think any other LunarG repo needs to change the default branch. Just this one.

charles-lunarg (commented):
I don't see any mention of the master->main switch in the readme or build. (I remember seeing a mention of that in the Vulkan-Tools repo)

christophe-lunarg (approved):
Consistency sounds good to me

Tony-LunarG:
Ok

juan-lunarg:
> I don't see any mention of the master->main switch in the readme or build. (I remember seeing a mention of that in the Vulkan-Tools repo)

Updated README to include that

> Consistency sounds good to me

Currently there are only 2 active PUBLIC repos in LunarG

image

`Vulkan-Tools` and `gfxreconstruct`

reconstruct uses the `dev` convention which is fine. But seeing as `Vulkan-Tools` devs often work on `Khronos` repos, it ultimately makes more sense to use `main` for consistency.

charles-lunarg (approved):
Thanks!

vkconfig/CHANGELOG.md
```diff
@@ -107,7 +107,7 @@
 - Add build date in the title bar
 
 ### Improvements:
-- Add link to [Vulkan Guide layers](https://github.com/KhronosGroup/Vulkan-Guide/blob/master/chapters/development_tools.md#vulkan-layers) list within the help menu
+- Add link to [Vulkan Guide layers](https://github.com/KhronosGroup/Vulkan-Guide/blob/main/chapters/development_tools.md#vulkan-layers) list within the help menu
```
juan-lunarg, line 110: Also using `main` consistently helps with minor issues like this in the documentation.

PR #1786, opened 2023-01-20
Update to 1.3.239 headers

Tony-LunarG:


charles-lunarg (approved):
Updating headers should be fine, assuming all the builds succeed.

PR #1790, opened 2023-02-10
API Dump support video extensions

charles-lunarg:
NOTE: This uses the currently in progress #1771 PR as a base, and therefore is not ready for primetime. Expect a rebase on main when the refactor PR is complete. Until then, this PR contains commits of both

This PR adds support for the video extensions in API Dump. The video extensions were not supported initially because they required definitions for the various external types that were defined in video.xml but because these definitions live in a different xml file, they weren't generated. This PR fixes it by running the generator script over the video.xml files to produce text, html, and json header files that are then used by the existing API Dump code.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -213,15 +213,21 @@ if (WIN32)
                          IMPORTED_LOCATION_DEBUG "${JSONCPP_DLIB}")
 endif()
 
-# Define macro used for building vkxml generated files
-macro(run_vulkantools_vk_xml_generate dependency output)
+# Define macro used for building vk.xml generated files
+function(run_vulkantools_vk_xml_generate dependency output)
     add_custom_command(OUTPUT ${output}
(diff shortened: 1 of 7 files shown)


charles-lunarg:
I have updated the code to fix the issues causing android & tests to fail.

Therefore the code changes are fit to be reviewed. However, the extended tests written in the internal CI scripts have not been finalized so this PR shouldn't be merged until those have a chance to run & check out.

Tony-LunarG:
The int8_t variables in vkCreateVideoSessionParametersKHR are not printing or are printing a strange character.  Other than that, LGTM

charles-lunarg:
> The int8_t variables in vkCreateVideoSessionParametersKHR are not printing or are printing a strange character. Other than that, LGTM

I thought I fixed that already. Maybe I just forgot to push it up, which is very possible! Anyhow, thanks for the review.

Tony-LunarG (approved):
LGTM

charles-lunarg:
Turns out I was confusing the `uint8_t` fix for the `int8_t` which explains why I thought I fixed it but really hadn't. I pushed the fix up now, which is simple because its the same solution for uint8_t & int8_t. Once it passes CI I will merge.

PR #1792, opened 2023-02-20
apidump: Don't lock for functions that may block

### LINKED ISSUE #1718: apidump causes deadlock by holding mutex on down-chain calls
Running https://github.com/KhronosGroup/Vulkan-Samples/tree/master/samples/extensions/timeline_semaphore and I get a deadlock when trying to use `VK_LAYER_LUNARG_api_dump`. Appears to be caused by the layer holding `outputMutex` during the call to the driver. The sample app waits on a timeline semaphore in one thread while submitting a signal operation in another--but the signal never gets submitted because the wait is holding the lock.

I'm not sure what the correct thing to do here is, since this behavior looks like it was introduced in response to other issues: #791 #932

charles-lunarg:
Originally API dump had no issues with blocking API calls, because it would only take the lock for printing after the function calls returned. But since the change to print the function name before calling down the chain (in case the function crashes) required printing, that meant the lock was taken before calling down the chain. If that function blocks then this stalls EVERY other API call from progressing, which may cause a dead lock.

The solution is to manually specify the functions which may block and have those functions revert to the original behavior (not print before calling down). This prevents deadlocks by not holding the lock during blocking Vulkan function calls.

To do this change and not just copy paste all of the code, the 'dump_function_head' function was added to api_dump.h which internally has the switch table between different output types. This makes the generated code much easier to read (the switch is inside of a function instead of appearing twice in the api_dump_generator.py) and also makes the generated code ~5000 lines shorter. The `update_object_name_map` is also added for similar reasons - why generate the same code for each function when it can be done once.

Fixes #1718

### DIFF
--- a/layersvt/api_dump.h
+++ b/layersvt/api_dump.h
@@ -826,6 +826,19 @@ class ApiDumpInstance {
         return vk_instance_map.at(phys_dev);
     }
 
+    void update_object_name_map(const VkDebugMarkerObjectNameInfoEXT *pNameInfo) {
+        if (pNameInfo->pObjectName)
+            object_name_map.emplace(pNameInfo->object, pNameInfo->pObjectName);
+        else
+            object_name_map.erase(pNameInfo->object);
+    }
+    void update_object_name_map(const VkDebugUtilsObjectNameInfoEXT *pNameInfo) {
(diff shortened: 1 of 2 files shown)


mikes-lunarg (approved):
LGTM!

PR #1794, opened 2023-02-21
Remove usage of VK_LAYER_EXPORT

### LINKED ISSUE #1793: VK_LAYER_EXPORT has been removed from Vulkan-Headers
Api dump, screenshot, monitor and anything using layer factory will need to stop using VK_LAYER_EXPORT

charles-lunarg:
This PR fixes API Dump, Monitor, Screenshot, and the vk_layer_settings utility to no longer use the VK_LAYER_EXPORT macro which was defined in vk_layer.h. The macro was recently removed thus all the projects using it must be updated.

In addition to defining the macro locally to each file, this commit removes function exporting for functions which do not need it. For Example, it was being added to each and every function in API Dump when only ~7 functions in the entire Vulkan interface need to be exported. Additionally, it was entirely removed from the vk_layer_settings utility due to the utility being a static library thus having no effect.

Fixes #1793

### DIFF
--- a/layersvt/monitor.cpp
+++ b/layersvt/monitor.cpp
@@ -84,7 +84,7 @@ static std::unordered_map<void *, monitor_layer_data *> layer_data_map;
 template monitor_layer_data *GetLayerDataPtr<monitor_layer_data>(void *data_key,
                                                                  std::unordered_map<void *, monitor_layer_data *> &data_map);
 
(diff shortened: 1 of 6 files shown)


Tony-LunarG (approved):
LGTM - Thanks

juan-lunarg (approved):

PR #1803, opened 2023-03-24
Update paths from VVL refactor

ncesario-lunarg:


build-android/build_all.sh
```diff
@@ -41,7 +41,7 @@ function findtool() {
 # Check for dependencies
 findtool aapt
 findtool zipalign
-findtool jarsigner
```
ncesario-lunarg, line 44: I think this is eventually necessary, and it was easier to do this than install the JDK locally for me.

build-android/build_all.sh
```diff
@@ -51,10 +51,10 @@ echo LAYER_BUILD_DIR="${LAYER_BUILD_DIR}"
 echo CUBE_BUILD_DIR="${CUBE_BUILD_DIR}"
 
 function create_APK() {
-    aapt package -f -M AndroidManifest.xml -I "$ANDROID_SDK_HOME/platforms/android-23/android.jar" -S res -F bin/$1-unaligned.apk bin/libs
```
ncesario-lunarg, line 54: I updated this to match what is in `Application.mk` so that only platform-26 is needed to build (not 26 _and_ 23).

scripts/api_dump_generator.py
```diff
@@ -1500,6 +1500,9 @@
 {{
     char *tm = (char*)t;
     size_t tmlen;
+#ifdef __ANDROID__
+    (void)tmlen; // Appears to be unused on Android
```
ncesario-lunarg, line 1504: @charles-lunarg I'm not familiar with code generation here, but--on Android at least--the `@foreach union` block below appears to not produce anything, causing an "unused variable warning" here.

charles-lunarg: oops!
I haven't touched the android stuff in api_dump in forever. Feel free to make it work as you see fit.

build-android/vulkan-validationlayers_url_android
```diff
@@ -1 +1 @@
-https://github.com/KhronosGroup/Vulkan-ValidationLayers.git
\ No newline at end of file
+https://github.com/ncesario-lunarg/Vulkan-ValidationLayers.git
```
ncesario-lunarg, line 1: This change depends on https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/5494

scripts/api_dump_generator.py
charles-lunarg: After further review & discussion, this is a bug. The unions used should be from the Vulkan API, not the Video API. Thus, I will make a PR to fix it shortly.

charles-lunarg: This has been fixed, please rebase on main to see the changes if you haven't already.

Tony-LunarG (approved):
LGTM - Thanks!

scripts/known_good.json
```diff
@@ -60,7 +60,7 @@
       "sub_dir" : "Vulkan-ValidationLayers",
       "build_dir" : "Vulkan-ValidationLayers/build",
       "install_dir" : "Vulkan-ValidationLayers/build/install",
-      "commit": "v1.3.243",
+      "commit": "main",
```
juan-lunarg, line 63: main?

PR #1805, opened 2023-03-31
apidump: Generate is_struct/is_union logic

charles-lunarg:
Move the logic for dertermining if something is a struct or union to be a hardcoded value instead of having to perform a runtime lookup. This is a possibility since whether a type is a struct or union is apart of the Vulkan API.

This fixes an issue introduced in the PR to support the video extensions where there is_struct & is_union functions only included the video API types, not the entire API.

### DIFF
--- a/layersvt/api_dump.h
+++ b/layersvt/api_dump.h
@@ -1222,7 +1222,8 @@ void dump_json_function_head(ApiDumpInstance &dump_inst, const char *funcName, c
 
 template <typename T>
 void dump_json_array(const T *array, size_t len, const ApiDumpSettings &settings, const char *type_string, const char *child_type,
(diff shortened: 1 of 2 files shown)


Tony-LunarG (approved):
LGTM - Any chance of adding a test to keep from reverting?

charles-lunarg:
> LGTM - Any chance of adding a test to keep from reverting?

Unfortunately the tests as they currently are assume 'correct' output, and haven't been enabled. 
The issue arises from having the same codegen run twice (once for vulkan, once for video) and not taking care what functions get multiple definitions. Assuming something like video doesn't happen often, I don't see this as being a common occurrence.
At least this change removes a function with hardcoded values (since its all statically defined) removing that vector for bugs.

That said, the video extensions did make it into the latest SDK which makes turning tests on infeasible since the current SDK output is used in the tests, and currently are incorrect.

PR #1816, opened 2023-04-21
Make update_object_name_map() update on name re-assignment

jmacnak:
The emplace() function will not update the value in the map if it is already present.

This is useful if you are labelling objects which get recycled. For examples, you might label a fence "Frame 1 Complete Fence" and then again "Frame 4 Complete Fence".

### DIFF
--- a/layersvt/api_dump.h
+++ b/layersvt/api_dump.h
@@ -837,13 +837,13 @@ class ApiDumpInstance {
 
     void update_object_name_map(const VkDebugMarkerObjectNameInfoEXT *pNameInfo) {
         if (pNameInfo->pObjectName)
-            object_name_map.emplace(pNameInfo->object, pNameInfo->pObjectName);
+            object_name_map[pNameInfo->object] = pNameInfo->pObjectName;
         else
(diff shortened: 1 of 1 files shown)


charles-lunarg (approved):
I did not know that about the emplace function. And looking at the blame, it appears it used to use `insert` which seems to have the same issue.

Assuming it passes CI, I think this is a good change.

charles-lunarg:
Seems that android does not like this change. This will need to be rectified before it can be merged.

juan-lunarg:
> Seems that android does not like this change. This will need to be rectified before it can be merged.

I don't think it's this change. `kDebugBit` used to be defined by the validation layers until https://github.com/KhronosGroup/Vulkan-ValidationLayers/commit/4272e43f7e48404eef2c90cef286791742a00d7e

charles-lunarg:
I see. I created a PR to fix it here: #1817

charles-lunarg:
@jmacnak can you rebase on main? The issue causing android CI to fail should be resolved with it.

PR #1817, opened 2023-04-22
layer_factory: Update debug bit to verbose

charles-lunarg:
The ValidationLayer repo updated the definition of the debug bit to be 'verbose', so VulkanTools needs to update to reflect that change.


juan-lunarg (approved):
How did this happen? I thought known_good.json pointed to a specific commit?

charles-lunarg:
Seems that it happened because the android build uses tip of main rather than a specific commit (like from known_good.json)

https://github.com/LunarG/VulkanTools/blob/main/build-android/vulkan-validationlayers_revision_android

charles-lunarg:
And *of course* this is going to fail every other build that uses known_good for its validation layer version 🤦‍♂️

charles-lunarg:
This PR will still need to happen, but is not useful right now since this repo is still on 1.3.246. Until this repo updates to 1.3.250 (or thereabouts) it should be ignored.

charles-lunarg:
Closing in favor of PR #1821

PR #1818, opened 2023-04-22
android ci: Use specific validation layer commit

charles-lunarg:
During various fixes to this repo, the android CI version was moved to main, which fixed the issues that we were seeing at the time. Now that the dust has settled, its time to move it back to a stable version.


juan-lunarg (approved):

charles-lunarg:
Welp. That didn't fix it. I'm going to look into this another day. Seems there is more going on than meets the eye - at least as far as updating versions go. Best option may be to perform a whole header update for this repo.

build-android/vulkan-validationlayers_revision_android
```diff
@@ -1 +1 @@
-main
+c935d1e214006b40743647572144c2ce3882e4aa
```
ncesario-lunarg, line 1: Is there something special about this commit, or could this point to the v1.3.247 tag?

ncesario-lunarg (approved):
One question, but LGTM!

build-android/vulkan-validationlayers_revision_android
charles-lunarg: I picked this at semi random. I didn’t go for the 247 tag because I feared that it would need newer headers but that may be a non issue. Then again having 246 and 247 tags is mighty confusing.

PR #1820, opened 2023-04-27
Update version.h

neurotok:
Fix for multiple compilation errors on Fedora 38 g++ (GCC) 13.0.1 due to the lack of cstdint include

### DIFF
--- a/vkconfig_core/version.h
+++ b/vkconfig_core/version.h
@@ -22,6 +22,7 @@
 #pragma once
 
 #include <cstddef>
+#include <cstdint>
 #include <string>
 
 #define SUPPORT_LAYER_CONFIG_2_2_0 1


juan-lunarg (approved):

juan-lunarg:
> Ok, I just ubmitted CLA

The CLA check is still showing as pending. (Sorry about this I know it's not ideal).

juan-lunarg:
Thanks for the fix @neurotok 

Apologies for the delay. The CLA tool Khrono uses is buggy sadly.

neurotok:
I work for Khronos @juan-lunarg, and we are doing our best ;) I'm also in touch with @adam-sawicki-a, the creator of VMA. The next release will have a fix, and the whole SDK shouldn't cause any problems by then.

This whole GCC 13 confusion took everyone by surprise, but I guess it's an important move towards future module-based compilation units architecture of C++

adam-sawicki-a:
I recommend to use latest version of VMA from "master" branch, where this missing include is fixed. Official releases were not updated for many months now.

juan-lunarg:
> I recommend to use latest version of VMA from "master" branch, where this missing include is fixed. Official releases were not updated for many months now.

We recently updated the version of VMA in the validation layers.

https://github.com/KhronosGroup/Vulkan-ValidationLayers/commit/c0c44a6452c9594ed1d36f28b7d1f7dfbda4ca03

But looking at our SDK scripts we still use `v3.0.1` I'll make sure our upcoming SDK release uses tip of master (https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/commit/0e89587db3ebee4d463f191bd296374c5fafc8ea)

KarenGhavam-lunarG:
>  I'll make sure our upcoming SDK release uses tip of master

We use the official releases from the memory allocator and rely upon @adam-sawicki-a judgment as to when that release should be updated. I do see that 3.0.1 is about a year old. @adam-sawicki-a should it be updated?

adam-sawicki-a:
I recommend taking latest version from “master” branch, not the Release, as we didn’t create a new Release for long time now, while the code on “master” branch is always kept in a good state and there were many bug fixes since then.

PR #1823, opened 2023-05-09
Fix JSON output on Mac M1

MarkY-LunarG:
The MacOS on the M1 produced a product version that seems to have a tab in it which messes up VkVia.  Attempt to fix this.


richard-lunarg (approved):
Looks good, and already tested on macOS.

PR #1835, opened 2023-06-10
apidump: Use logcat instead of stdout for Android

06393993:
* Implement the AndroidLogcatBuf class that derives from std::streambuf which writes to the Android logcat on a per-line basis.
* Modify ApiDumpSettings to hold an instance of AndroidLogcatBuf, and make the output_stream to use that streambuf instead of the iostream one at initialization.

### DIFF
--- a/layersvt/api_dump.h
+++ b/layersvt/api_dump.h
@@ -61,6 +61,8 @@
 #include <utility>
 
 #ifdef ANDROID
+#include <memory>
+#include <string_view>
 
 #include <android/log.h>
 #include <sys/system_properties.h>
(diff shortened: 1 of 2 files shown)


layersvt/api_dump.h
```diff
@@ -241,9 +242,69 @@ class ConditionalFrameOutput {
     }
 };
 
+#ifdef __ANDROID__
+
+class AndroidLogcatBuf : public std::streambuf {
```
06393993, line 247: We can make this class more generic, so that it can accept `wchar_t`, `char8_t`, `char16_t`, `char32_t`.

layersvt/api_dump.h
```diff
@@ -241,9 +242,69 @@ class ConditionalFrameOutput {
...
+        buffer[bufsize - 1] = traits_type::to_char_type(c);
+        flush_rest();
+        buffer[bufsize - 1] = '\0';
+        return traits_type::not_eof(c);
+    }
+
+    virtual int sync() override {
+        flush_rest();
+        if (!previous_rest.empty()) {
+            __android_log_print(4, "VALIDATION", "%s", previous_rest.c_str());
```
06393993, line 266: * We probably don't want to use the magic number 4
* We may want to use the tag name defined as a macro somewhere in the repo.

06393993:
One can find the example of output at https://gist.github.com/06393993/95beab8c36f975e07c1e22c6ff6ce2f6. Note that when dumping the return result, `:` will be on a separate line, because we flush after dumping the function head. This will trigger another Android log call, and result in a separate line.

We can implement to only call the Android log function at `\n`, but we then will risk that the log won't be written even when `std::flush` is used.

I am happy to add unit tests to `AndroidLogcatBuf`, but I need instructions on where to place the code. Thanks!

charles-lunarg (commented):
Not super familiar with the process of extending basic_streambuf, nor how android logs work, so I can't comment in great detail.

What I can say is that the purpose of printing only the function head then flushing is to make sure that if a crash occurs, at least the name of the most recently called function is present. A solution to the issue of formatting the arguments and return value is to modify how text output is formatted - just so it doesn't look as confusing.

layersvt/api_dump.h
```diff
@@ -241,9 +243,96 @@ class ConditionalFrameOutput {
...
+        if (c == traits::eof()) {
+            return traits::not_eof(c);
+        }
+        buffer_[bufsize_ - 1] = traits::to_char_type(c);
+        flushPending();
+        buffer_[bufsize_ - 1] = traits::to_char_type(0);
+        return traits::not_eof(c);
+    }
+
+    virtual int sync() override {
```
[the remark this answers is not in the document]
06393993: Done.

layersvt/api_dump.h
```diff
@@ -241,9 +243,96 @@ class ConditionalFrameOutput {
...
+        : buffer_(std::make_unique<char_type[]>(bufsize)), bufsize_(bufsize), log_writer_(std::move(log_writer)) {
+        // We use the last position of buffer_ to store the overflow character. We use 0 as a sentinel element to indicate if this
+        // overflow slot is in use.
+        this->setp(buffer_.get(), buffer_.get() + bufsize_ - 1);
+    }
+    ~AndroidLogcatBuf() = default;
+
+   private:
+    using int_type = typename traits::int_type;
+    virtual int_type overflow(int_type c) override {
```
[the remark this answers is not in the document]
06393993: Done.

06393993:
> just so it doesn't look as confusing

I personally don't think it's confusing and just prefer as is for simplicity. See the example output:

```
06-10 19:32:53.116  2490  2517 I api_dump: vkWaitForFences(device, fenceCount, pFences, waitAll, timeout) returns VkResult
06-10 19:32:53.116  2490  2517 I api_dump:  VK_SUCCESS (0):
06-10 19:32:53.116  2490  2517 I api_dump:     device:                         VkDevice = 0x7a3d604869f0
```

```
06-10 19:32:53.132  2490  2517 I api_dump: vkUnmapMemory(device, memory) returns void
06-10 19:32:53.132  2490  2517 I api_dump: :
06-10 19:32:53.132  2490  2517 I api_dump:     device:                         VkDevice = 0x7a3d604869f0
```

I agree with the current behavior that we flush after printing the function head in case of crash - that's the exact case where I am currently working.

lunarpapillo:
The LunarG CI Android failure appears to be systemic, but may be masking an Android issue.  I'm working on resolving it.

06393993:
Friendly ping, I am wondering if anyone could rerun the CI for this PR? The CI build fails because it fails to install QT, which I think is irrelevant to this PR. Thanks!

layersvt/api_dump.h
```diff
@@ -241,9 +243,96 @@ class ConditionalFrameOutput {
...
+template <class char_type = char, class traits = std::char_traits<char_type>>
+class AndroidLogcatBuf final : public std::basic_streambuf<char_type, traits> {
+   public:
+    class LogWriter {
+       public:
+        virtual void write(const std::basic_string<char_type, traits> &data) = 0;
+        virtual ~LogWriter() {}
+    };
+
+    AndroidLogcatBuf(std::unique_ptr<LogWriter> log_writer, size_t bufsize = 0x400)
```
06393993, line 256: I need to change the default `bufsize` to a smaller value like `0x200`, to prevent `__android_log_print` from truncating the message, we can easily hit this issue if we ask the api dump layer to output the shader source.

`__android_log_print` will truncate any input longer than 1024 because [`__android_log_print`](https://cs.android.com/android/platform/superproject/main/+/e60a2cd5ccf434f66761ab1bcd22169ab6372be2:system/logging/liblog/logger_write.cpp;l=410) [uses  `vsnprintf`](https://cs.android.com/android/platform/superproject/main/+/e60a2cd5ccf434f66761ab1bcd22169ab6372be2:system/logging/liblog/logger_write.cpp;l=421) with `LOG_BUF_SIZE` as the length parameter to format the input string. And `LOG_BUF_SIZE` is currently [defined as 1024](https://cs.android.com/android/platform/superproject/main/+/e60a2cd5ccf434f66761ab1bcd22169ab6372be2:system/logging/liblog/logger_write.cpp;l=65).

06393993:
Hi maintainers,

Can I ask for a code review for my PR? We really need this CL for our development, and do want to avoid fork as much as possible. Thanks!

spencer-lunarg:
sorry, I was going to review this, but as normal things got crazy around here... will take a look today/tomorrow

layersvt/api_dump.h
06393993: Done.

layersvt/api_dump.h
```diff
@@ -45,7 +45,9 @@
 #include <algorithm>
 #include <chrono>
 #include <fstream>
+#include <string_view>
```
spencer-lunarg, line 48: nit: can we wrap these includes with `#ifdef __ANDROID__` 

I plan to got through this file and remove a lot of these that are not being used (after realizing it)

so

```c++
#ifdef ANDROID
#include <memory>
#include <string_view>

#include <android/log.h>
#include <sys/system_properties.h>
```

spencer-lunarg (approved):
Looked good on my end locally

Just a nit about the `include` and happy to get this merged

spencer-lunarg:
(seems there is some issues with the CI unrelated, I will take care of those)

layersvt/api_dump.h
06393993: Done.

PR #1837, opened 2023-06-14
apidump: Use correct aliased types

### LINKED ISSUE #977: api_dump_generator.py codegen should handle typedef stucts automaticallly
The typedef section in scripts/api_dump_generator.py (about line 540) should be removed and code for these typedef structures should be autogen.

charles-lunarg:
VkAccelerationStructureTypeNV and VkBuildAccelerationStructureFlagsNV are aliased types which are explicitly used by the VK_NV_ray_tracing extension. Because apidump didn't know how to keep track of aliases and then fixup uses of aliased types into the real type, those two types had hand written typedefs. This commit is able to remove the hardcoding as well are removes both redundant code (self.trackedTypes) and dead code (str_VkStructureTypeToEnum).

Fixes #977

### DIFF
--- a/scripts/api_dump_generator.py
+++ b/scripts/api_dump_generator.py
@@ -431,23 +431,6 @@ void dump_text_pNext_trampoline(const void* object, const ApiDumpSettings& setti
 void dump_text_{unName}(const {unName}& object, const ApiDumpSettings& settings, int indents);
 @end union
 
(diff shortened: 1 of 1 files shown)


charles-lunarg:
Just tested it locally, but this fixes the issue causing VK_NV_cooperative_matrix to not generate correctly.

PR #1838, opened 2023-06-28
Fix msvc runtime library

charles-lunarg:


juan-lunarg (changes requested):
👎🏾 

See linker errors.

christophe-lunarg (approved):

PR #1848, opened 2023-08-15
api_dump: Prevent crash in VkDescriptorDataEXT union

charles-lunarg:
The VK_EXT_descriptor_buffer extension adds a union with pointer elements. Because API Dumps union logic is naive, it prints the value of every element in the union regardless of whichever the 'active' element is, as it is not always possible to easily get the active element. VkDescriptorDataEXT contains pointers which are dereferenced resulting in a crash. Fixing it is straightforward because the parent structure VkDescriptorGetInfoEXT contains the type to use, so it is just a matter using the PARAMETER_STATE and VALIDITY_CHECKS functionality to set & get the type when printing out the contents of the union.


charles-lunarg:
I reran the CI job, and it has succeeded. Going to go ahead and merge.

PR #1849, opened 2023-08-17
Remove unused CMake build

juan-lunarg:


build-android/cmake/README.md
```diff
@@ -1,30 +0,0 @@
-Build Validation Layers with Android CMake Plugin
```
juan-lunarg, line 1: ...

charles-lunarg (approved):
Always good to remove unused stuff that is very likely not tested or unknown in whether it works at all.

PR #1853, opened 2023-08-18
ci: Cleanup GitHub Actions CI

juan-lunarg:
Fixup performance issues
Fixup CI warnings

### DIFF
--- a/CMakeUserPresets.json
+++ b/CMakeUserPresets.json
@@ -1,0 +1,24 @@
+{
+    "version": 3,
+    "cmakeMinimumRequired": {
+        "major": 3,
+        "minor": 26,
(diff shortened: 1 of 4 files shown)


CMakeUserPresets.json
```diff
@@ -0,0 +1,24 @@
+{
+    "version": 3,
+    "cmakeMinimumRequired": {
+        "major": 3,
+        "minor": 26,
```
charles-lunarg, line 5: Is this the CMake version needed for the android platform specified?

juan-lunarg: I forgot to rebase... CMakeUserPresets.json shouldn't be commited into a repo.

PR #1855, opened 2023-08-18
apidump: Query device functions during GetInstanceProcAddr

charles-lunarg:
vkGetInstanceProcAddr is able to query device functions, which was not being done by api_dump. This caused debug utils functions to not be intercepted due to changes in how the loader queries those functions.

### DIFF
--- a/scripts/api_dump_generator.py
+++ b/scripts/api_dump_generator.py
@@ -366,12 +366,34 @@ VKAPI_ATTR {funcReturn} VKAPI_CALL {funcName}({funcTypedParams})
 }}
 @end function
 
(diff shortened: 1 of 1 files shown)


scripts/api_dump_generator.py
```diff
@@ -1519,7 +1539,7 @@
 @end function
 """
 
-POINTER_TYPES = ['void', 'xcb_connection_t', 'Display', 'SECURITY_ATTRIBUTES', 'ANativeWindow', 'AHardwareBuffer']
+POINTER_TYPES = ['void', 'xcb_connection_t', 'Display', 'SECURITY_ATTRIBUTES', 'ANativeWindow', 'AHardwareBuffer', 'wl_display']
```
juan-lunarg, line 1542: ah I see

juan-lunarg (approved):

PR #1861, opened 2023-08-22
cmake: Remove cmake directory

juan-lunarg:
We don't use these scripts


charles-lunarg (commented):
I definitely approve of this, even if it has already merged.

PR #1864, opened 2023-08-23
cmake: Remove excess cmake_minimum_required / project usage

juan-lunarg:


vkconfig/CMakeLists.txt
```diff
@@ -1,6 +1,3 @@
-cmake_minimum_required(VERSION 3.10.2)
```
juan-lunarg, line 1: Note currently min required is 3.17.2

but calling cmake_minimum_required like this changes to older policies. This can be very confusing and cause annoying CMake issues. Due to behavior changes between files.

PR #1878, opened 2023-08-31
layer: Use layer settings library

christophe-lunarg:


juan-lunarg (approved):

juan-lunarg:
Looks good overall. The only thing I saw as problematic was this workaround for the screenshot layer:

https://github.com/KhronosGroup/Vulkan-Utility-Libraries/commit/87801a6c479399889eea37069f83437d0ea9458e

christophe-lunarg:
> Looks good overall. The only thing I saw as problematic was this workaround for the screenshot layer:
> 
> [KhronosGroup/Vulkan-Utility-Libraries@87801a6](https://github.com/KhronosGroup/Vulkan-Utility-Libraries/commit/87801a6c479399889eea37069f83437d0ea9458e)

Yes, that's because the variable didn't have any name itself so it couldn't be identified by the library. It's not great, I created a name for the setting but didn't want to break backward compatibility.

juan-lunarg:
> Yes, that's because the variable didn't have any name itself so it couldn't be identified by the library. It's not great, I created a name for the setting but didn't want to break backward compatibility.

Does VUL have a mechanism for deprecating layer settings? That way we could provide a warning about the back compat breakage?

PR #1881, opened 2023-08-31
Add case for iOS builds with cmake

richard-lunarg:
A minor tweak to allow for building with iOS as a target.
Note I left out the case for VK_USE_PLATFORM_MACOS_MVK on purpose as it is deprecated. While there are probably people using this still for macOS, I thought it best not to encourage it moving forward, such as on iOS.


richard-lunarg:
Minor change, only affects iOS builds and passed CI. Merging without unnecessary review.

PR #1883, opened 2023-09-04
build: Remove VVL as a dependency

juan-lunarg:


scripts/api_dump_generator.py
```diff
@@ -176,11 +176,45 @@
     return result;
 }}
 
+VkResult util_GetExtensionProperties(const uint32_t count, const VkExtensionProperties *layer_extensions, uint32_t *pCount,
```
charles-lunarg, line 179: util_GetExtensionProperties is also defined inside of screenshot or monitor (dont remember which). Rather than being hardcoded in api_dump, it should live in vk_layer_table.h so that all 3 layers can use the same implementation.

juan-lunarg: 👍🏾

layersvt/api_dump.h
```diff
@@ -336,6 +336,9 @@ class ApiDumpSettings {
         std::string filename_string = "";
         // If the layer settings file has a flag indicating to output to a file,
         // do so, to the appropriate filename.
+
+        // getLayerOption is defined inside VVL via vk_layer_config.cpp / vk_layer_config.h
+        // How should this be updated?
         const char *file_option = getLayerOption("lunarg_api_dump.file");
```
juan-lunarg, line 342: @christophe-lunarg on how to procede.

I thought about copying vk_layer_config.cpp from VVL but that solution didn't seem correct while I was implementing it.

charles-lunarg (approved):
Looks great!

juan-lunarg:
This repo builds a LOT faster now. Same goes for the CI. VVL took up quite a bit of time.

PR #1884, opened 2023-09-05
layer: Use layer settings library in API dump

christophe-lunarg:
I added a layer test framework, based on the Profiles solution layer test framework, so that we can use it to debug, step in layer code, with Visual Studio

### DIFF
--- a/layersvt/test/layer_test_helper.cpp
+++ b/layersvt/test/layer_test_helper.cpp
@@ -1,0 +1,143 @@
+/*
+ * Copyright (c) 2023 The Khronos Group Inc.
+ * Copyright (c) 2023 Valve Corporation
+ * Copyright (c) 2023 LunarG, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Author: Christophe Riccio <christophe@lunarg.com>
+ */
+
+#include "layer_test_helper.h"
+
+#ifdef _WIN32
+#include <Windows.h>
+#elif __ANDROID__
+#include <android/log.h>
+#endif
+
+#include <sstream>
+#include <cstdlib>
+#include <cctype>
+
+// TODO: When the layer path issue is resolved with CI (set outside of the tests) remove this function
+void layer_test::SetEnvironmentSetting(std::string setting, const char* val) {
+#ifdef _WIN32
+    _putenv_s(setting.c_str(), val);
+#else
+    setenv(setting.c_str(), val, 1);
+#endif
+}
+
+void layer_test::UnsetEnvironmentSetting(std::string setting) {
+#ifdef _WIN32
+    _putenv_s(setting.c_str(), "");
+#else
+    unsetenv(setting.c_str());
+#endif
+}
+
+std::string layer_test::GetAbsolutePath(std::string filepath) {
+    std::string out;
+#ifdef _WIN32
+    char abs_path[_MAX_PATH];
+    _fullpath(abs_path, filepath.c_str(), _MAX_PATH);
+    out = abs_path;
+#else
+    char* abs_path = realpath(filepath.c_str(), nullptr);
+    out = abs_path;
+    delete[] abs_path;
+#endif
+    return out;
+}
+
+VkApplicationInfo layer_test::GetDefaultApplicationInfo() {
+    VkApplicationInfo out{VK_STRUCTURE_TYPE_APPLICATION_INFO};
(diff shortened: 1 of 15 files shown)


juan-lunarg (approved):

juan-lunarg (changes requested):
```
[ RUN      ] build/tests/apidump_test.sh
COUNTS: 498 1 0
[  FAILED  ] build/tests/apidump_test.sh
```

API dump tests are failing

layersvt/test/test_api_dump.cpp
```diff
@@ -0,0 +1,50 @@
...
+    const char* output_format = "html";
+
+    const std::vector<VkLayerSettingEXT> settings = {
+        {kLayerName, "file", VK_LAYER_SETTING_TYPE_BOOL32_EXT, 1, &file},
+        {kLayerName, "log_filename", VK_LAYER_SETTING_TYPE_STRING_EXT, 1, &filename_string},
+        {kLayerName, "output_format", VK_LAYER_SETTING_TYPE_STRING_EXT, 1, &output_format}};
+
+    layer_test::VulkanInstanceBuilder inst_builder;
+    VkResult err = inst_builder.Init(settings);
+    EXPECT_EQ(err, VK_SUCCESS);
```
charles-lunarg, line 49: Would it be too hard to have a check which loads the .html file and makes sure its not empty? That at least makes sure a file is written and the contents are *not* empty. This may require destroying the instance first.

layersvt/test/test_screenshot.cpp
```diff
@@ -0,0 +1,40 @@
...
+    static void SetUpTestSuite() {}
+    static void TearDownTestSuite(){};
+};
+
+TEST_F(ScreenshotTests, init_layer) {
+    TEST_DESCRIPTION("Test Creating a Vulkan Instance with a layer");
+
+    layer_test::VulkanInstanceBuilder inst_builder;
+    VkResult err = inst_builder.Init(kLayerName);
+    EXPECT_EQ(err, VK_SUCCESS);
```
charles-lunarg, line 39: Actually, it wouldn't be too hard to load the ppm file and validate it. Don't worry about adding it now, just figured I should say that out loud since ppm is a text file more or less.

charles-lunarg (approved):
Much better handling of settings? Test framework for basic tests? Sign me up!

layersvt/test/layer_test_helper.h
```diff
@@ -0,0 +1,77 @@
...
+#include <string>
+#include <vector>
+#include <array>
+#include <cmath>
+#include <cstdarg>
+
+namespace layer_test {
+
+inline std::string format(const char* message, ...) {
+    std::size_t const STRING_BUFFER(4096);
```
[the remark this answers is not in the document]
christophe-lunarg: Ahah, that's a code I have been copy pasting for decades XD

I renamed the variable into `max_string_length`.

layersvt/test/test_api_dump.cpp
```diff
@@ -0,0 +1,50 @@
...
+static const char* kLayerName = "VK_LAYER_LUNARG_api_dump";
+
+class ApiDumpTests : public VkTestFramework {
+   public:
+    ApiDumpTests(){};
+    ~ApiDumpTests(){};
+
+    static void SetUpTestSuite() {}
+    static void TearDownTestSuite(){};
+};
```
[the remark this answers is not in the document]
christophe-lunarg: I had no particularly desire to even keep the class but `VkTestFramework` is defined with `~VkTestFramework() = 0;` so we need a sub class and I don't want to break that behavior because that's how it behave on other repository. 

It's the same with SetUpTestSuite and TearDownTestSuite so the only think I could remove is the contructors (`ApiDumpTests()`) which I have removed.

layersvt/test/layer_test_helper.cpp
```diff
@@ -0,0 +1,148 @@
...
+#endif
+
+#include <sstream>
+#include <cstdlib>
+#include <cctype>
+
+// On Android, VK_MAKE_API_VERSION doesn't yet exist.
+#ifndef VK_MAKE_API_VERSION
+#define VK_MAKE_API_VERSION(variant, major, minor, patch) VK_MAKE_VERSION(major, minor, patch)
+#endif
```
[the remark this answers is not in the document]
christophe-lunarg: Another case of copy pasted probably upsolete, removing it.

layersvt/test/test_api_dump.cpp
christophe-lunarg: I added reading the beginning of generated file to the test. This was simple enough.

layersvt/test/test_screenshot.cpp
christophe-lunarg: This case is a lot more involved, implying writing enough code to render frame so I left it alone.

PR #1889, opened 2023-09-12
apidump: Fix regression in the NoAddr setting

charles-lunarg:
api_dump.h stores the no-address setting in a variable called show_address. This requires inverting the setting to match what the rest of the api_dump code expects it to mean. This fixes a regression introduced by the transition to using the Layer Settings library from Vulkan-Utility-Libraries.


Tony-LunarG (approved):
LGMT

PR #1891, opened 2023-09-14
apidump: Return NULL in Get*ProcAddr for disabled extensions

pixelcluster:
vkGetDeviceProcAddr specifies that only function pointers of enabled extensions should be exposed. If an extension is not enabled, vkGetDeviceProcAddr should return NULL. The same goes for vkGetInstanceProcAddr, except that the extension only needs to be available. Some users like winevulkan rely on this behaviour and break when used with the current API dump layer.

Assuming that drivers and other layers act accordingly and return NULL for non-enabled entrypoints, we can mirror this by simply returning NULL when the corresponding dispatch table entry is NULL.

### DIFF
--- a/scripts/api_dump_generator.py
+++ b/scripts/api_dump_generator.py
@@ -379,10 +379,10 @@ VKAPI_ATTR PFN_vkVoidFunction VKAPI_CALL api_dump_known_instance_functions(const
     return nullptr;
 }}
 
-VKAPI_ATTR PFN_vkVoidFunction VKAPI_CALL api_dump_known_device_functions(const char* pName)
+VKAPI_ATTR PFN_vkVoidFunction VKAPI_CALL api_dump_known_device_functions(VkDevice device, const char* pName)
 {{
     @foreach function where('{funcType}' == 'device')
-    if(strcmp(pName, "{funcName}") == 0)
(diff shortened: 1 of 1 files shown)


scripts/api_dump_generator.py
```diff
@@ -369,20 +369,25 @@
 }}
 @end function
 
-VKAPI_ATTR PFN_vkVoidFunction VKAPI_CALL api_dump_known_instance_functions(const char* pName)
+VKAPI_ATTR PFN_vkVoidFunction VKAPI_CALL api_dump_known_instance_functions(VkInstance instance, const char* pName)
 {{
     @foreach function where('{funcType}' in ['global', 'instance'] and '{funcName}' not in [ 'vkEnumerateDeviceExtensionProperties' ])
+    @if('${funcDispatchType}' == 'instance')
+    if(strcmp(pName, "{funcName}") == 0 && (!instance || instance_dispatch_table(instance)->{funcShortName}))
+    @end if
```
charles-lunarg, line 377: So this causes API Dump to return NULL if the *next* function in the chain is NULL? Seems like a pretty easy way to make sure to only return functions which will successfully pass through the chain (and not crash when called).

charles-lunarg: Just re-read the PR description, it does exactly what I surmised from the change.
And it makes sense, so I'll approve it.

charles-lunarg (approved):

PR #1892, opened 2023-09-15
layers: Update to new names in vk_dispatch_table.h

charles-lunarg:
The functions and structs in vk_dispatch_table.h have been updated to use a 'vku' prefix instead of 'vul'. Thus, this repo needs to update to it so the next header update goes smoothly.


charles-lunarg:
LunarG CI suffered a sporadic failure, rerunning it passed CI.

PR #1895, opened 2023-09-21
Update to latest Vulkan-Utility-Libraries

charles-lunarg:
Also fix regression in Environment Variable naming: 

The use of the layer settings library changed the established names for
the settings set through environment variables. This commit addresses
this by using the compatibility namespace feature to ensure existing
users aren't affected.

### DIFF
--- a/layersvt/api_dump.h
+++ b/layersvt/api_dump.h
@@ -461,20 +461,20 @@ class ApiDumpSettings {
     bool isFrameInRange(uint64_t frame) const { return condFrameOutput.isFrameInRange(frame); }
 
     void init(const VkInstanceCreateInfo *pCreateInfo, const VkAllocationCallbacks *pAllocator) {
-        VlLayerSettingSet layerSettingSet = VK_NULL_HANDLE;
-        vlCreateLayerSettingSet("VK_LAYER_LUNARG_api_dump", vlFindLayerSettingsCreateInfo(pCreateInfo), pAllocator, nullptr,
+        VkuLayerSettingSet layerSettingSet = VK_NULL_HANDLE;
+        vkuCreateLayerSettingSet("VK_LAYER_LUNARG_api_dump", vkuFindLayerSettingsCreateInfo(pCreateInfo), pAllocator, nullptr,
                                 &layerSettingSet);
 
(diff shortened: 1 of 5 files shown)


layersvt/api_dump.h
```diff
@@ -461,20 +469,22 @@ class ApiDumpSettings {
...
 
     void init(const VkInstanceCreateInfo *pCreateInfo, const VkAllocationCallbacks *pAllocator) {
-        VlLayerSettingSet layerSettingSet = VK_NULL_HANDLE;
-        vlCreateLayerSettingSet("VK_LAYER_LUNARG_api_dump", vlFindLayerSettingsCreateInfo(pCreateInfo), pAllocator, nullptr,
-                                &layerSettingSet);
+        VkuLayerSettingSet layerSettingSet = VK_NULL_HANDLE;
+        vkuCreateLayerSettingSet("VK_LAYER_LUNARG_api_dump", vkuFindLayerSettingsCreateInfo(pCreateInfo), pAllocator, nullptr,
+                                 &layerSettingSet);
+
+        vkuSetLayerSettingCompatibilityNamespace(layerSettingSet, GetDefaultPrefix());
```
juan-lunarg, line 476: @christophe-lunarg with `vkuSetLayerSettingCompatibilityNamespace` do we still need the workaround in VUL?

IE

```

static std::string GetEnvironment(const char *variable) {
#if defined(__ANDROID__)
    std::string result = GetAndroidProperty(variable);
    // Workaround for screenshot layer backward compatibility
    if (result.empty() && std::string(variable) == "debug.vulkan.screenshot.frames") {
        result = GetAndroidProperty("debug.vulkan.screenshot");
    }
    return result;
```

charles-lunarg:
Internal CI seems to have failed with a sporadic Qt build error.

layersvt/api_dump.h
christophe-lunarg: Unfortunately, yes we do. The issue is that the setting had no name. vkuSetLayerSettingCompatibilityNamespace only allow to set a compatibiilty namespace for the setting but the library still require names to identify setting.

christophe-lunarg (approved):

PR #1901, opened 2023-10-17
deps: Mark Vulkan-Tools as a test dependency

charles-lunarg:
The Vulkan-Tools repo is only used for testing, but known_good.json did not specify that.


juan-lunarg (approved):

juan-lunarg:
Fixup commit message typo

lunarpapillo (approved):
:pray:

PR #1903, opened 2023-10-17
screenshot: Always add queues

charles-lunarg:
Since queue handles are used to dispatch down the call chain, there is no reason to skip adding them in any case. This occurs if the screenshot setting is empty.


juan-lunarg (approved):

charles-lunarg:
LunarG CI failure is due to a sporadic failure of a single machine.

PR #1907, opened 2023-10-23
apidump: Append file extension is not present

charles-lunarg:
If the filename provided in the settings file does not end with the corresponding extension for the output type, api dump will append it.

### DIFF
--- a/layersvt/api_dump.h
+++ b/layersvt/api_dump.h
@@ -475,29 +475,7 @@ class ApiDumpSettings {
 
         vkuSetLayerSettingCompatibilityNamespace(layerSettingSet, GetDefaultPrefix());
 
-        // If the layer settings file has a flag indicating to output to a file,
-        // do so, to the appropriate filename.
-        std::string filename_string = "";
-        if (vkuHasLayerSetting(layerSettingSet, kSettingsKeyFile)) {
-            bool file = false;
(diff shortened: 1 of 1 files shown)


layersvt/api_dump.h
```diff
@@ -475,43 +475,69 @@ class ApiDumpSettings {
...
-                if (vkuHasLayerSetting(layerSettingSet, kSettingsKeyLogFilename)) {
-                    vkuGetLayerSettingValue(layerSettingSet, kSettingsKeyLogFilename, filename_string);
-                    if (filename_string.empty()) {
-                        filename_string = "vk_apidump.txt";
-                    }
-                }
+            if (output_format == ApiDumpFormat::Html) {
+                filename_string = "vk_apidump.txt";
+            } else if (output_format == ApiDumpFormat::Json) {
+                filename_string = "vk_apidump.txt";
```
Tony-LunarG, line 503: Should all three of these be the same?  Should html end with .html?

Tony-LunarG (commented):
Question about doing the same assign independent of output_format

layersvt/api_dump.h
charles-lunarg: Thats a copy paste error! Good catch.

Tony-LunarG (approved):
LGTM

PR #1913, opened 2023-10-30
CI Cleanup

juan-lunarg:
- MinGW fixes (via doesn't build, /bigobj only applies to MSVC compiler)
- Add MacOS build to CI
- Remove using clang on linux. clang is the default toolchain for mac/android
- Fix bug where qmake was relying on the incorrect version of vulkan-headers
- Remove check_code_format.sh (it has been doing nothing except fail silently for a while now)


vkconfig/vkconfig.pro
```diff
@@ -5,9 +5,8 @@ greaterThan(QT_MAJOR_VERSION, 4): QT += widgets network
 CONFIG += c++11
 CONFIG += sdk_no_version_check
 
-INCLUDEPATH += ../Build/Vulkan-Headers/include
+INCLUDEPATH += ../external/Vulkan-Headers/include
```
juan-lunarg, line 8: Note this fix isn't perfect. But will do for now.

charles-lunarg (approved):
Changes look appropriate, and represents good cleanup of the github actions CI.

juan-lunarg:
> CI VulkanTools build # 2769 failed.

CI is failing to what looks like a configuration issue with the 32 bit Windows machine. I started a run on main that doesn't include my changes.

http://tcubuser.lunarg.localdomain:8080/job/VulkanTools/2770/

If this one fails then something is wrong with Jenkins.

juan-lunarg:
> http://tcubuser.lunarg.localdomain:8080/job/VulkanTools/2770/
> 
> If this one fails then something is wrong with Jenkins.

It fails. Something is wrong with Jenkins.

juan-lunarg:
Re-ran jenkins on main multiple times and now it works...

PR #1914, opened 2023-10-30
dependabot / ubuntu back compat

juan-lunarg:
- Add dependabot
- Check that we don't break ubuntu backcompat


charles-lunarg (approved):
Very much like the other PR's that do the same thing in other repos

PR #1917, opened 2023-10-30
Use jsoncpp package config support

### LINKED ISSUE #1908: Can't rely on jsoncpp providing cmake configs
jsoncpp has deprecated cmake build system in favor of meson. Distros / packages are switching over, and so the cmake config cannot be relied anymore. This is a problem on Homebrew for instance.

The pkgconfig is generated in both build systems, and so we can switch over, or at least add it as a fallback, but preferably the former for simplicity.

Something like this works:
```
find_package(PkgConfig REQUIRED)
pkg_check_modules(jsoncpp REQUIRED IMPORTED_TARGET jsoncpp)
add_library(jsoncpp_static ALIAS PkgConfig::jsoncpp) # Or preferably don't create the alias and use the imported target directly.
```

juan-lunarg:
closes #1908

### DIFF
--- a/via/CMakeLists.txt
+++ b/via/CMakeLists.txt
@@ -19,8 +19,7 @@ else()
         set(COMMON_COMPILE_FLAGS "${COMMON_COMPILE_FLAGS} -fno-strict-aliasing -fno-builtin-memcmp")
         set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${COMMON_COMPILE_FLAGS} -fno-rtti")
     endif()
-    # jsoncpp has intentional fallthrough
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${COMMON_COMPILE_FLAGS} -std=c++11 -Wno-implicit-fallthrough")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${COMMON_COMPILE_FLAGS} -std=c++11")
 
     if(APPLE)
         add_definitions(-DVIA_MACOS_TARGET)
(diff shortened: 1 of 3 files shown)


juan-lunarg:
FYI @mizvekov

via/CMakeLists.txt
```diff
@@ -60,9 +59,18 @@ add_executable(vkvia
                   via_system_bsd.cpp)
 endif()
 
+# https://github.com/LunarG/VulkanTools/issues/1908
+find_package(PkgConfig)
+if (PKG_CONFIG_FOUND)
+    pkg_check_modules(jsoncpp IMPORTED_TARGET jsoncpp)
```
mizvekov, line 65:
```suggestion
    pkg_check_modules(jsoncpp REQUIRED IMPORTED_TARGET jsoncpp)
```
For clearer error in case the pkgconfig for jsoncpp can't be found.

mizvekov (commented):
Btw, did you ping jsoncpp developers about MR adding cmake config support?

juan-lunarg:
> CI VulkanTools build # 2779 failed.

Jenkins failing due to unrelated reason. I've seen this sporadic failure in the past.

```
C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\MSBuild\Microsoft\VC\v160\BuildCustomizations\masm.targets(70,5): error MSB4175: The task factory "XamlTaskFactory" could not be loaded from the assembly "Microsoft.Build.Tasks.v4.0, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a". Could not find file 'C:\Users\LunarG\AppData\Local\Temp\npj4s5tc.dll'. [D:\j\vt\Debug32\VulkanTools\external\Vulkan-Loader\build\loader\loader-unknown-chain.vcxproj]
```

juan-lunarg:
> Btw, did you ping jsoncpp developers about MR adding cmake config support?

Yes

charles-lunarg:
Was going to approve but failed to hit 'submit review' in time. (was working in parallel on vulkan-loader changes). 

The changes look good, if a bit sad due to needing to workaround the lack of pkg config support on some platforms (aka, making the build more involved.) But it is a necessity due to external factors, and so the code is needed.

PR #1919, opened 2023-11-01
cmake: Only require the CXX compiler

juan-lunarg:
Allows simplifying the rest of the CMake code

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 3.17.2)
 
-project(VULKAN_TOOLS)
+project(VULKAN_TOOLS LANGUAGES CXX)
 
 add_subdirectory(scripts)
 


CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 3.17.2)
 
-project(VULKAN_TOOLS)
+project(VULKAN_TOOLS LANGUAGES CXX)
```
juan-lunarg, line 3: By default for legacy reasons CMake enables both the C/C++ languages. This repo doesn't need that.

It makes things more complicated.

EX:

Needing to specify CMAKE_CXX_VISIBILITY_PRESET and CMAKE_C_VISIBILITY_PRESET and being more aware of the compiler flags being used.

charles-lunarg (approved):

PR #1920, opened 2023-11-01
cmake: Cleanup scripts

juan-lunarg:
- Remove unused variables
- Set CMAKE_CXX_STANDARD to 17
- Fix redundant code
- Fix usage of CMAKE_CXX_FLAGS with modern alternatives
- Remove usage of CMAKE_C_FLAGS (not needed)
- Simplify CMakeLists.txt for vkvia using modern target commands
- Use TARGET_NAME_IF_EXISTS for Vulkan::Loader


charles-lunarg (approved):
🧹

PR #1922, opened 2023-11-01
ci: Delete scripts/github_ci_win.py

juan-lunarg:
This script doesn't provide anything particularly useful.

Just the regular CMake build/test workflow.

Simpler to just run the commands via github actions with the Ninja generator.

Jenkins uses the Visual Studio generator so we don't lose any testing coverage.


scripts/github_ci_win.py
juan-lunarg, line 1: NOTE: github_ci_linux.py isn't as trivial. It also tests the qmake build and adi_dump testing which the windows script didn't.

.github/workflows/ci.yml
```diff
@@ -60,13 +60,13 @@ jobs:
...
-      - uses: actions/setup-python@v4
-        with:
-          python-version: '3.7'
-      - run: python3 scripts/github_ci_win.py --config Debug --arch x64
+      - uses: ilammy/msvc-dev-cmd@v1
+      - uses: jurplel/install-qt-action@v3
+      - run: cmake -S . -B build -D UPDATE_DEPS=ON -D BUILD_TESTS=ON -D CMAKE_BUILD_TYPE=Debug -G Ninja
+      - run: cmake --build build
+      - run: ctest --output-on-failure --test-dir build
+      - run: cmake --install build --prefix build/install
```
juan-lunarg, line 69: Furthermore I actually looked at the WIN32 installation and it appears to have issues:
```
Run cmake --install build --prefix build/install
  
-- Install configuration: "Debug"
-- Installing: D:/a/VulkanTools/VulkanTools/build/install/bin/vkvia.exe
CMake Warning (dev) at build/vkconfig/cmake_install.cmake:40:
  Syntax Warning in cmake code at column 108

  Argument not separated from preceding token by whitespace.
Call Stack (most recent call first):
-- Installing: D:/a/VulkanTools/VulkanTools/build/install/bin/vkconfig.exe
  build/cmake_install.cmake:62 (include)
This warning is for project developers.  Use -Wno-dev to suppress it.

"C:\Program" does not exist.
```

Special install code that doesn't seem to work properly. And NONE of the layers are installed.

charles-lunarg (approved):

PR #1926, opened 2023-11-02
cmake: Fix constant dirty builds

juan-lunarg:
Currently these custom targets are always dirty.

It's simpler to just use execute_process and utilize CMake's copy_if_different command during the configuration step.


charles-lunarg (approved):

juan-lunarg:
> CI VulkanTools build # 2799 failed.

Sporadic CI failure due to not finding Qt...

PR #1929, opened 2023-11-02
Add iOS build to CI

juan-lunarg:
Currently only API dump is being built/installed


charles-lunarg (approved):
Changes look fine but as not being an expert of apple, I may be unaware of the intricacies. 

I really like the check for macos/ios platform defines in api_dump.h!

PR #1938, opened 2023-11-13
Use pkg-config for jsoncpp if possible

juan-lunarg:
The main thing to note is we want to use the STATIC version of jsoncpp

By default pkg-config was pointing to the SHARED version

I explicitly turn this off in the known_good.json now so now the pkg-config path will work for us for the linux tarball.

### DIFF
--- a/scripts/known_good.json
+++ b/scripts/known_good.json
@@ -81,7 +81,10 @@
                 "-DJSONCPP_WITH_TESTS=OFF",
                 "-DJSONCPP_WITH_POST_BUILD_UNITTEST=OFF",
                 "-DJSONCPP_WITH_WARNING_AS_ERROR=OFF",
-                "-DJSONCPP_WITH_PKGCONFIG_SUPPORT=OFF"
+                "-DJSONCPP_WITH_PKGCONFIG_SUPPORT=ON",
+                "-DBUILD_SHARED_LIBS=OFF",
+                "-DBUILD_STATIC_LIBS=ON",
+                "-DBUILD_OBJECT_LIBS=OFF"
             ]
         },
         {
--- a/via/CMakeLists.txt
+++ b/via/CMakeLists.txt
@@ -66,21 +66,14 @@ elseif (MSVC)
     target_compile_definitions(vkvia PRIVATE _CRT_SECURE_NO_WARNINGS _USE_MATH_DEFINES)
 endif()
 
-find_package(jsoncpp CONFIG)
-
-if (TARGET jsoncpp_static)
-    # RPATH causes issues for the Linux SDK tarball due to absolute paths.
-    # This can be worked around with relative RPATHS but the tarball doesn't ship jsoncpp.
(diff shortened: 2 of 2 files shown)


via/CMakeLists.txt
```diff
@@ -68,17 +68,14 @@ endif()
...
-    # https://github.com/LunarG/VulkanTools/issues/1908
-    find_package(PkgConfig REQUIRED)
+# https://github.com/LunarG/VulkanTools/issues/1908
+find_package(PkgConfig)
+if (PKG_CONFIG_FOUND)
     pkg_check_modules(jsoncpp REQUIRED IMPORTED_TARGET jsoncpp)
     target_link_libraries(vkvia PRIVATE PkgConfig::jsoncpp)
+else()
+    find_package(jsoncpp CONFIG)
+    target_link_libraries(vkvia PRIVATE jsoncpp_static)
```
juan-lunarg, line 78: Note this part is important for internal Windows build that doesn't have pkg-config. We currently don't use pkg-config at all for our Windows builds and adding it is low priority.

scripts/known_good.json
```diff
@@ -81,7 +81,10 @@
                 "-DJSONCPP_WITH_TESTS=OFF",
                 "-DJSONCPP_WITH_POST_BUILD_UNITTEST=OFF",
                 "-DJSONCPP_WITH_WARNING_AS_ERROR=OFF",
-                "-DJSONCPP_WITH_PKGCONFIG_SUPPORT=OFF"
+                "-DJSONCPP_WITH_PKGCONFIG_SUPPORT=ON",
+                "-DBUILD_SHARED_LIBS=OFF",
+                "-DBUILD_STATIC_LIBS=ON",
+                "-DBUILD_OBJECT_LIBS=OFF"
```
juan-lunarg, line 87: I'll update the SDK scripts otherwise we will have issues.

charles-lunarg (approved):
Static linking is the right choice here I believe.

via/CMakeLists.txt
```diff
@@ -66,21 +66,14 @@ elseif (MSVC)
...
-    find_package(PkgConfig REQUIRED)
+# https://github.com/LunarG/VulkanTools/issues/1908
+find_package(PkgConfig)
+if (PKG_CONFIG_FOUND)
     pkg_check_modules(jsoncpp REQUIRED IMPORTED_TARGET jsoncpp)
     target_link_libraries(vkvia PRIVATE PkgConfig::jsoncpp)
 else()
-    message(FATAL_ERROR "Unable to find jsoncpp!")
+    find_package(jsoncpp CONFIG REQUIRED)
+    target_link_libraries(vkvia PRIVATE jsoncpp_static)
```
juan-lunarg, line 76: Internal Windows CI doesn't have package config. Furthermore most Windows developers don't tend to have this by default. So for now we support both paths. At least both paths are actively tested though which was the main issue before.

PR #1939, opened 2023-11-14
Cleanup layer CMake code

juan-lunarg:


layersvt/CMakeLists.txt
```diff
@@ -92,48 +76,31 @@ endif ()
...
-            CXX_STANDARD_REQUIRED ON
-            CXX_EXTENSIONS OFF
-        )
-        set_target_properties(copy-${target}-def-file PROPERTIES FOLDER "VkLayer_${target}")
+        add_library(VkLayer_${target} SHARED ${ARGN})
+        if (MSVC)
+            target_link_options(VkLayer_${target} PRIVATE /DEF:${CMAKE_CURRENT_SOURCE_DIR}/VkLayer_${target}.def)
+        elseif(MINGW)
+            target_sources(VkLayer_${target} PRIVATE /DEF:${CMAKE_CURRENT_SOURCE_DIR}/VkLayer_${target}.def)
+        endif()
```
charles-lunarg, line 84: Weird that this is different.

juan-lunarg: MINGW doesn't use the MSVC linker it uses a modified version of the GNU linker

juan-lunarg: Also see updated fix what I had there was wrong

layersvt/CMakeLists.txt
```diff
@@ -92,48 +76,31 @@ endif ()
...
-            CXX_STANDARD_REQUIRED ON
-            CXX_EXTENSIONS OFF
-        )
-        set_target_properties(copy-${target}-def-file PROPERTIES FOLDER "VkLayer_${target}")
+        add_library(VkLayer_${target} SHARED ${ARGN})
+        if (MSVC)
+            target_link_options(VkLayer_${target} PRIVATE /DEF:${CMAKE_CURRENT_SOURCE_DIR}/VkLayer_${target}.def)
+        elseif(MINGW)
+            target_sources(VkLayer_${target} PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/VkLayer_${target}.def)
+        endif()
```
juan-lunarg, line 84: @charles-lunarg

charles-lunarg: I see... Well, this code is necessary then.

juan-lunarg: Yeah this is what we do in VVL:
https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/f046acda58667e92e7e16e006f693d220139148f/layers/CMakeLists.txt#L282

charles-lunarg (approved):

PR #1942, opened 2023-11-15
Build screenshot layer for MacOS

juan-lunarg:
Tested it locally using vkconfig and it worked just like on Linux.

### DIFF
--- a/layersvt/screenshot.cpp
+++ b/layersvt/screenshot.cpp
@@ -70,7 +70,8 @@ static char *local_getenv(const char *name) {
 
 static void local_free_getenv(const char *val) { android_env_map.erase(std::string(val)); }
 
-#elif defined(__linux__) || defined(__FreeBSD__)
+#elif defined(__linux__) || defined(__FreeBSD__) || defined(__APPLE__)
+
 static inline char *local_getenv(const char *name) { return getenv(name); }
 
 static inline void local_free_getenv(const char *val) {}
--- a/layersvt/screenshot.cpp
+++ b/layersvt/screenshot.cpp
@@ -51,54 +51,6 @@ const char *kSettingsKeyFrames = "frames";
 const char *kSettingKeyFormat = "format";
 const char *kSettingKeyDir = "dir";
 
-#ifdef ANDROID
-
-static std::map<std::string, std::string> android_env_map;
-
-static char *local_getenv(const char *name) {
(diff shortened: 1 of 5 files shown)


juan-lunarg:
Pointed to my local version of vkcube and it worked without issue.

```

Launching Vulkan Application:
- Layers overridden by "Screenshot" configuration.
- Application: my_vkcube
- Executable: /Users/juanramos/projects/KHR-Vulkan-Tools/build/cube/vkcube.app/Contents/MacOS/vkcube
- Working Directory: /Users/juanramos/projects/KHR-Vulkan-Tools/build/cube/vkcube.app/Contents/MacOS
- Command-line Arguments: --suppress_popups
- Log file: /Users/juanramos/VulkanSDK/vkcube.txt

screenshot: Selected format:USE_SWAPCHAIN_COLORSPACE
Is NOT in the list:
UNORM, SNORM, USCALED, SSCALED, UINT, SINT, SRGB
Swapchain Colorspace will be used instead

Selected GPU 0: Apple M2 Max, type: IntegratedGpu

screenshot: Capture file is: /Users/juanramos/VulkanSDK/1.ppm 

Process terminated
```

charles-lunarg (approved):
Nice, just a 'make it compile' change with no really custom bits needed.

juan-lunarg:
> Nice, just a 'make it compile' change with no really custom bits needed.

Yeah when I was working on cleaning up the layer installation code. I got curious and was wondering what it would take to make this work.

Turns out almost nothing.

layersvt/screenshot.cpp
```diff
@@ -51,53 +51,6 @@ const char *kSettingsKeyFrames = "frames";
 const char *kSettingKeyFormat = "format";
 const char *kSettingKeyDir = "dir";
 
-#ifdef ANDROID
```
richard-lunarg, line 54: Why is all this being removed?

juan-lunarg: It does nothing. Should have removed when we integrated the utility libraries which handles this logic now.

richard-lunarg: Ah, great. It's the only thing that didn't make sense "from the outside".

richard-lunarg (approved):
Great work! Just building it for macOS is a testament to Vulkan's portability. Thanks also for adding the portability extension to the tests, this often get's missed.

PR #1945, opened 2023-12-12
apidump: Fix file setting being ignored

charles-lunarg:
If the 'file' option is set in the settings file, it will now be used to determine whether to write to a file output or not. Previously it was being ignored entirely, resulting in all output heading to a file.


richard-lunarg (approved):
Ship it!

richard-lunarg:
And please cherry pick to vulkan-tmp-1.3.273

charles-lunarg:
Done

PR #1947, opened 2023-12-14
Fix SDK regression

juan-lunarg:
Currently UNIX platforms like linux/mac default to using package config and don't use CMake config.

This broke our existing SDK process for jsoncpp

### DIFF
--- a/via/CMakeLists.txt
+++ b/via/CMakeLists.txt
@@ -66,14 +66,24 @@ elseif (MSVC)
     target_compile_definitions(vkvia PRIVATE _CRT_SECURE_NO_WARNINGS _USE_MATH_DEFINES)
 endif()
 
-# https://github.com/LunarG/VulkanTools/issues/1908
-find_package(PkgConfig)
-if (PKG_CONFIG_FOUND)
-    pkg_check_modules(jsoncpp REQUIRED IMPORTED_TARGET jsoncpp)
-    target_link_libraries(vkvia PRIVATE PkgConfig::jsoncpp)
-else()
-    find_package(jsoncpp CONFIG REQUIRED)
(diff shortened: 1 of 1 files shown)


via/CMakeLists.txt
```diff
@@ -66,16 +66,27 @@ elseif (MSVC)
...
-if (PKG_CONFIG_FOUND)
-    pkg_check_modules(jsoncpp REQUIRED IMPORTED_TARGET jsoncpp)
-    target_link_libraries(vkvia PRIVATE PkgConfig::jsoncpp)
-else()
-    find_package(jsoncpp CONFIG REQUIRED)
+# NOTE: jsoncpp hasn't had updates in a bit and isn't getting fixes.
+# may be worth using nlohmann/json in the future which gets more updates
+# and has better CMake support.
+find_package(jsoncpp CONFIG)
+if (TARGET jsoncpp_static)
```
juan-lunarg, line 73: Not because of a bug in jsoncpp I can't detect if the find_package worked like with other packages so I have to detect the target instead.

charles-lunarg (approved):

via/CMakeLists.txt
```diff
@@ -66,14 +66,24 @@ elseif (MSVC)
...
     target_link_libraries(vkvia PRIVATE jsoncpp_static)
+    
+# Support using jsoncpp.pc but only for UNIX platforms.
+# And only if UPDATE_DEPS is disabled.
+elseif (UNIX AND NOT UPDATE_DEPS)
```
juan-lunarg, line 78: Updated this to be an elseif otherwise it will still try and use pkg-config pointlessly

johnzupin (approved):

juan-lunarg:
Thanks for testing this @johnzupin

PR #1959, opened 2024-02-27
Update to 1.3.278

charles-lunarg:


richard-lunarg (commented):
Thanks for the quick fix!

PR #1980, opened 2024-04-16
vkconfig: Add system tray icon

christophe-lunarg:
New behavior:
image

Old behavior:
image

The system tray, it allows switching from `Vulkan Layers controlled by Vulkan Configurator` (red) and `Vulkan Layers controlled by Vulkan Applications` (Gray)
image
image

Here is the system tray menu, not much, I initially went crazy with option because realizing it was clunky:
image


vkconfig/resourcefiles/vkconfig-over.png
charles-lunarg, line 1: What is the purpose of this icon? Its a neat design, but I'm not seeing where its used.

charles-lunarg (approved):
Adding options to the tray is a small but really powerful change!

vkconfig_core/environment.h
```diff
@@ -34,8 +34,8 @@
 enum LayersMode {
     LAYERS_MODE_BY_APPLICATIONS = 0,
     LAYERS_MODE_BY_CONFIGURATOR_RUNNING,
-    LAYERS_MODE_BY_CONFIGURATOR_PERSISTENT,
-    LAYERS_MODE_BY_CONFIGURATOR_ALL_DISABLED,
+    // LAYERS_MODE_BY_CONFIGURATOR_PERSISTENT,
```
charles-lunarg, line 37: Should this be commented out?

vkconfig/resourcefiles/vkconfig-over.png
christophe-lunarg: I have been investigating multiple approaches and it's not use at the moment so I would remove it before I merge.

artem-lunarg:
What happen when you close the main window to the tray? Does it behave the same way as when it was opened or something changes?

christophe-lunarg:
> What happens when you close the main window to the tray? Does vkconfig work the same way it did when its main window was opened?

When "Keep in system tray" is checked, then Vulkan Configurator remains active in the same state it was when closing the main windows.

If "Keep in system tray" is unchecked, then Vulkan Configurator exist and the layers override files are removed so the layers are controlled by the Vulkan applications.

I can take a better label of the check box if you have a good idea !!! :D

artem-lunarg:
@christophe-lunarg thanks for the explanation. This is exactly the behavior that I expect from this functionality. Also really nice improvement for substituting "continue override" which felt for me too implicit, easy to forget and then wonder what's going on. 

I like the checkbox description, it's quite intuitive.

charles-lunarg:
I built this branch locally but could not get it to go to the tray. Rather, it exits when I close it (even though the popup does mention going to system tray). I believe its due to being on wayland, so there might be some quirkyness due to that.

artem-lunarg:
Also please ensure that vulkan configurator can be restored from system tray by double clicking (or single clicking, different apps behave differently, I always double click). I'm not sure if this functionality automatically provided by the windows or controlled by the application.

christophe-lunarg:
> Also please ensure that vulkan configurator can be restored from system tray by double clicking on the icon (or single clicking, different apps behave differently, I always double click). I'm not sure if this functionality is automatically provided by the windows or controlled by the application.

I tried this but it doesn't capture the difference between double click and single click. It means I need to choose between toggeling between active or not or restoring...

What do you prefer?

christophe-lunarg:
> I built this branch locally but could not get it to go to the tray. Rather, it exits when I close it (even though the popup does mention going to system tray). I believe its due to being on wayland, so there might be some quirkyness due to that.

I pushed some changed that might fix this as I reverted some toggles XD

artem-lunarg:
If any of this works that's great. Some apps I have respond for single click and some for double click. But because I always double click it makes no difference to me - if the app was designed to be opened with a single click it also reacts to a double click.

christophe-lunarg:
> > I tried this but it doesn't capture the difference between double click and single click. It means I need to choose between toggeling between active or not or restoring...
> > What do you prefer?
> 
> If any of this works that's great. Some apps I have respond for single click and some for double click. But because I always double click it makes no difference to me - if the app was designed to be opened with a single click it also reacts to a double click.

Currently, I am with enable/diable on click because I notice my other tray icons has basically in their menu Open and Exit only. So I assumed people are used to right click to restore.

charles-lunarg:
Suggestion: Make closing to tray optional. I have some apps which can go to the system tray but don't want that behavior, I want to make sure they are closed whenever I can't see them.

artem-lunarg:
From what I understand that is how it behaves currently, but +1 for Charles suggestion. Most apps that I can minimize to tray have  a configuration checkbox to disable this behavior and just close entire app.

artem-lunarg:
> So I assumed people are used to right click to restore.

Right click and then choose from menu? If yes, that's too slow. Personally I never do this. Use right click if need menu option, if there is any.

> Currently, I am with enable/diable on click because I notice my other tray icons has basically in their menu Open and Exit only. 

Could you clarify what does enable/disable mean in this context?

artem-lunarg:
Typically right click on tray opens context menu with popular actions, so you can access them without opening main window. There is also aspect of discoverability here. Context menu with actions has the name with each action and it's easy to learn. Clicking on tray icon is obviously not associated with some text description, that's one of the reasons most app default to the same behavior to just open a window.

christophe-lunarg:
> > So I assumed people are used to right click to restore.
> 
> Right click and then choose from the menu an option to restore app main window? If yes, that's too slow. Personally I never do this (just double click, it looks like all apps in my tray support this). I use right click if I need a menu option, if there is any.
> 
> > Currently, I am with enable/diable on click because I notice my other tray icons has basically in their menu Open and Exit only.
> 
> Could you clarify what does enable/disable mean in this context?

It means "Vulkan layers controlled by Vulkan Configurator" (enabled) and "Vulkan layers controlled by Vulkan Applications" (disabled)

christophe-lunarg:
> Suggestion: Make closing to tray optional. I have some apps which can go to the system tray but don't want that behavior, I want to make sure they are closed whenever I click the X button.

That's the purpose of the checkbox !

charles-lunarg:
> > Suggestion: Make closing to tray optional. I have some apps which can go to the system tray but don't want that behavior, I want to make sure they are closed whenever I click the X button.
> 
> That's the purpose of the checkbox !

Which checkbox? The one in the popup? I thought that was just "don't remind me about closing to a tray in the future". Or was there a different checkbox in the GUI that I missed?

christophe-lunarg:
> > > Suggestion: Make closing to tray optional. I have some apps which can go to the system tray but don't want that behavior, I want to make sure they are closed whenever I click the X button.
> > 
> > 
> > That's the purpose of the checkbox !
> 
> Which checkbox? The one in the popup? I thought that was just "don't remind me about closing to a tray in the future". Or was there a different checkbox in the GUI that I missed?

First screenshot in the screenshot, surrourded with green. XD

charles-lunarg:
OH I was running vkconfig3!

That would explain why the layout looked different...

christophe-lunarg:
> OH I was running vkconfig3!
> 
> That would explain why the layout looked different...

That is not working at all !

charles-lunarg:
I didn't test out vkconfig3 except for the tray closing features... I assumed wrongly that new features would be vkconfig3 exclusive.

PR #1983, opened 2024-04-17
vkconfig: Better handle implicit layer

christophe-lunarg:
1. All implicit layers must have disable_environment in their manifest
2. If disable_environment is set (to anything) in the environment, the layer will not be enabled
3. If enable_environment is in the layer's manifest, then the layer wont be enabled UNLESS the exact string in enable_environment is set in the environment.

image

### DIFF
--- a/vkconfig/CHANGELOG.md
+++ b/vkconfig/CHANGELOG.md
@@ -24,7 +24,7 @@
 - Better present implicit layers, showing "Implicitly On" or "Implicitly Off" depending on the environment variable
 
 ### Fixes:
-- Fix clear log when changing `Vulkan Kiader Messages` type when `Clear log on action` is unchecked
+- Fix clear log when changing `Vulkan Loader Messages` type when `Clear log on action` is unchecked
 - Fix spaces in application launcher command-line 'path' arguments
 - Fix 'Use of application list' check box not being saved
 - Clean up UI update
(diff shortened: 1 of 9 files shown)


charles-lunarg (commented):
"Implicitly Off" is confusing - "off is off implicitly" so people might wonder what this is doing.

I'm not sure of what a better way to describe the layer is however.

christophe-lunarg:
> "off is off implicitly"

No idea what that means ! XD

charles-lunarg:
> > "off is off implicitly"
> 
> No idea what that means ! XD

Apologies, that isn't very clear. "Implicitly Off" sounds like a double negative. So to me it isn't clear what it does. "Off" implies that it doesn't activate, but "implicitly off" doesn't add any clarity about when it is/isn't enabled. 

Expanding the description to `Implicitly Off (but can be enabled by env-vars)` is much clearer but is way too long.

Not to mention that I have a hard time distinguishing Off/On at a glance.

Maybe "Environment Variable controlled"? is a better name? Cause that is what it is.

christophe-lunarg:
> > > "off is off implicitly"
> > 
> > 
> > No idea what that means ! XD
> 
> Apologies, that isn't very clear. "Implicitly Off" sounds like a double negative. So to me it isn't clear what it does. "Off" implies that it doesn't activate, but "implicitly off" doesn't add any clarity about when it is/isn't enabled.
> 
> Expanding the description to `Implicitly Off (but can be enabled by env-vars)` is much clearer but is way too long.
> 
> Not to mention that I have a hard time distinguishing Off/On at a glance.
> 
> Maybe "Environment Variable controlled"? is a better name? Cause that is what it is.

"Implicit" has notice that implies on or off to me.

"Env Variable controlled On" "Env Variable controlled Off" and adding tooltips with the value of the env variables would work to me.

PR #2011, opened 2024-04-29
vkconfig: Only display setting are not change dynamically in context

christophe-lunarg:
I made and tweak for when the message is displayed based on your feedback in here:
https://gitlab.khronos.org/vulkan/Vulkan-SDK-Packaging/-/issues/1491

Now, the message is displyed only is the user is changing setting while an application is launched. The first time.


ziga-lunarg:
>while an application is launched

Which application? Only if launched through vkconfig?

christophe-lunarg:
> > while an application is launched
> 
> Which application? Only if launched through vkconfig?

Yes

PR #2042, opened 2024-05-23
vkconfig3: Implement loader settings file

christophe-lunarg:
- Replace the override layer manifest by the loader settings file
- Removed legacy layer manifest files
- Extract code from the GUI to the core to increase the test coverage
- Refactor `Path`
- Add tests


charles-lunarg (commented):
What will happen if the loader on the system is older and doesn't support the loader settings file? Warn the user and close? I ask as this seems to replace the old override layer stuff with the loader settings files, and so I'm asking if a new loader is a 'hard requirement' for vkconfig3 to function.

christophe-lunarg:
> What will happen if the loader on the system is older and doesn't support the loader settings file? Warn the user and close? I ask as this seems to replace the old override layer stuff with the loader settings files, and so I'm asking if a new loader is a 'hard requirement' for vkconfig3 to function.

Yes, after discussion with Karen, we decided it will be a hard requirement. I am going to add a message box saying that Vulkan developers can use Vulkan Configurator 2.

PR #2043, opened 2024-06-07
apidump: Add wayland support

charles-lunarg:
Just needed to fix the wl_surface not being considered a pointer type and adding the appropriate CMake build logic.


mark-lunarg:
Works on my end, LGTM.

PR #2045, opened 2024-06-12
layer: Update Android layer settings doc

christophe-lunarg:
Note that `VK_APIDUMP_LOG_FILENAME` is still supported for backward compatibility reason but `VK_API_DUMP_LOG_FILENAME` is the expect name.

Environment variable layer settings names are built this way:
`VK_[VENDOR_][LAYER_NAME_POSTFIX_]SETTING_NAME`

For Android, the name are build similarly:
`debug.vulkan.[VENDOR.][LAYER_NAME_POSTFIX.]setting_name`

The layer settings library will first search the full name :
`debug.vulkan.lunarg.api_dump.setting_name`

And then trim the vendor name:
`debug.vulkan.api_dump.setting_name`

And then trim the layer name:
`debug.vulkan.setting_name`

Enabling using the same setting for multiple layers and also as a trick to be more backward compatible with the previous layer setting system and each layer implementation that were quite inconsistent.

An inconsistency with API Dump is that in the environment name the `_` was remove... There is a workaround in API dump to support both `API_DUMP` and `APIDUMP`.

Unfortunately using a `_` in `apidump_log_filename` as a separator between the layer name and the setting name is a case that was not covered by the layer setting library... Should we add backward compatbility for that? That will be a pretty hacky thing to do, sadly not the only thing hacky in the layer settings library as the layer settings across layers were badly inconsistent with details like this...

### DIFF
--- a/layersvt/api_dump_layer.md
+++ b/layersvt/api_dump_layer.md
@@ -111,14 +111,14 @@ The easiest way to set a property is from the ADB shell:
 
 **For example:**
 
-To set the API Dump output log filename, which on desktop uses `VK_APIDUMP_LOG_FILENAME`
+To set the API Dump output log filename, which on desktop uses `VK_API_DUMP_LOG_FILENAME`
 set the following property:
 
-    debug.apidump_log_filename
+    debug.vulkan.api_dump.log_filename
 
 Which you can set in the following way:
 
-    adb shell "setprop debug.apidump_log_filename '/sdcard/Android/vk_apidump.txt'"
+    adb shell "setprop debug.vulkan.api_dump.log_filename '/sdcard/Android/vk_apidump.txt'"
 
 <br></br>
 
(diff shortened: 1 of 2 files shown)


mark-lunarg:
Hey @christophe-lunarg, this is great and is exactly what I had hoped for.  I think the doc change is appropriate, and I do not think it is worth trying to retroactively add support for the `apidump_setting_name` stuff in the old docs (obviously, these settings are not used very often).  So I think this is good to go as-is.  Note that there is also [documentation in the SDK](https://vulkan.lunarg.com/doc/sdk/1.3.283.0/windows/api_dump_layer.html) that should be updated to match also.  Thanks for this!  Oh, one more thing, I think the description you added for this PR is valuable enough to live permanently in the vulkan-utility-library docs someplace -- excellent!

mark-lunarg (approved):

Issue #2046, opened 2024-06-20
vkconfig New Vulkan Layers Management combo-box does not save configuration


dorian-apanel-intel:
The radio selection has been changed to a combo box:
![image](https://github.com/LunarG/VulkanTools/assets/72810961/38324ad8-0460-4ac8-b761-3ad83b11af29)

However, if you only change the value of the combo box without clicking elsewhere, you might assume that the configuration has been updated when, in fact, it has not. This issue was not present with the radio selection that was used previously, which worked correctly.

Example:
1. Open vkconfig. It starts with "Layers controlled by the Vulkan Configurator" and the Validation layer selected.
2. Choose "Layers Controlled by the Vulkan Applications" but do not click anything else.
3. Open the console and run the following commands:
`set VK_LOADER_DEBUG=all`
`%VULKAN_SDK%\Bin\vulkaninfoSDK.exe > my_file.txt 2>&1`
4. Review the file and notice that the Validation layer is loaded, although it should not be.
5. Execute `dir "%appdata%\..\Local\LunarG\vkconfig\override"` and note the presence of the `VkLayer_override.json` and `vk_layer_settings.txt` files.
6. Now, click only the Status button at the bottom of vkconfig.
7. Repeat steps 3, 4, and 5 - the results will be different.

It was working correctly in radio selection present before

Observed on VulkanSDK binary package v.1.3.283.0

npavlicek:
I am not familiar with Qt5 however I played with the code and got it to start working. Someone forgot to change the Qt signal from triggered to toggled for the combo box options. The signal is not emitted for actions that are check-able so as a result the Configurators configure function is never called.

I will make a pull request with my fix.

dorian-apanel-intel:
@npavlicek  Thank you very much for addressing the issue.

Initially, I was puzzled because the code modification seemed to pertain solely to the tray and not to the combo box. However, upon inspecting the code under a debugger, I noticed that `UpdateTray` is invoked within the `OnComboBoxModeChanged` method.

```
 	vkconfig.exe!MainWindow::UpdateTray() Line 212	C++
 	vkconfig.exe!MainWindow::UpdateUI() Line 347	C++
 	vkconfig.exe!MainWindow::OnComboBoxModeChanged(int index) Line 525	C++
```

PR #2057, opened 2024-07-09
VkConfig: Fix tray combo-box signal method

npavlicek:
The combo-box for selecting what layers are being controlled is not correctly signaling its corresponding slot. As a result the configurator never gets updated. Checkable actions use the "toggled" signal rather than "triggered". A simple if statement was also added so only the selected action updates the configurator.

#2046


christophe-lunarg:
Thanks for contributing!

PR #2096, opened 2024-08-15
apidump: Use regular mutex & lock guards

charles-lunarg:
Replaces the usage of recursive_mutex with the standard mutex. Replaces explicit calls of lock() & unlock() with lock_guard.


ziga-lunarg (approved):
LGTM

PR #2131, opened 2024-10-09
vkconfig3: Layer tabs UI improvements

christophe-lunarg:
show_view
add_layers_path
configuration_tips

### DIFF
--- a/vkconfig_core/layer.h
+++ b/vkconfig_core/layer.h
@@ -80,7 +80,7 @@ class Layer {
     std::string enable_env;
     bool disable_value;
     bool enable_value;
-    bool visible = true;
+    bool enabled = true;
 
     std::vector<SettingMeta*> settings;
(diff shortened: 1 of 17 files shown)


ziga-lunarg:
What counts as a user-defined path? Should the SDK path not be in there?

christophe-lunarg:
> What counts as a user-defined path? Should the SDK path not be in there?

The one added automatically thanks to $VULKAN_SDK or built-in locations.

ziga-lunarg:
The reason I asked for this option was because if a layer in the SDK has the same version as the layer from a custom path, I don't know which one vkconfig will use.

image

christophe-lunarg:
> The reason I asked for this option was because if a layer in the SDK has the same version as the layer from a custom path, I don't know which one vkconfig will use.
> 
> image

Ahhh, hence my comment in the Google doc.

PR #2140, opened 2024-10-18
ci: Update to macOS-13 in github actions

charles-lunarg:
macOS-12 is being removed from github actions.


charles-lunarg:
Internal CI seems to have failed a test spuriously, but isn't related to this change (so was already failing spuriously)... Going to go ahead and merge.

PR #2216, opened 2024-12-06
api_dump: Fix destroy instance dispatch table

antonio-lunarg:
Make sure to get the dispatch key before destroying the instance as it
might not be valid anymore after calling `DestroyInstance`.

### DIFF
--- a/layersvt/generated/api_dump.cpp
+++ b/layersvt/generated/api_dump.cpp
@@ -161,9 +161,10 @@ VKAPI_ATTR void VKAPI_CALL vkDestroyInstance(VkInstance instance, const VkAlloca
 {
     std::lock_guard<std::mutex> lg(ApiDumpInstance::current().outputMutex());
     dump_function_head(ApiDumpInstance::current(), "vkDestroyInstance", "instance, pAllocator", "void");
+    auto dispatch_key = get_dispatch_key(instance);
     instance_dispatch_table(instance)->DestroyInstance(instance, pAllocator);
     
-    destroy_instance_dispatch_table(get_dispatch_key(instance));
+    destroy_instance_dispatch_table(dispatch_key);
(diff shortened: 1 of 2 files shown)


ziga-lunarg (approved):

ziga-lunarg (approved):
LGTM

charles-lunarg (commented):
This doesn't make much sense to me. 
get_dispatch_key isn't doing anything fancy.
```c++
dispatch_key get_dispatch_key(const void *object) { return (dispatch_key) * (VkuDeviceDispatchTable **)object; }
```
And the key should be present until destroy_instance_dispatch_table() is called

```c++
void destroy_dispatch_table(device_table_map &map, dispatch_key key) {
    device_table_map::const_iterator it = map.find((void *)key);
    if (it != map.end()) {
        delete it->second;
        map.erase(it);
    }
}
```

How does calling down the chain make the 'VkInstance' handle no longer valid?

Put another way, I'm very curious what the crash/bug/issue you saw which this reportedly fixes.

charles-lunarg:
I will say "VkInstance is no longer valid after calling vkDestroyInstance" is a valid statement, but the validity applies to the Vulkan implementation below API Dump. The validity of its own implementation is something which API dump maintains.

I am not against this PR to be clear, I do like saving the handle before calling down the vkDestroyInstance chain, I am just confused why/how this solves any problem.

antonio-lunarg:
@charles-lunarg thanks for your review!

I believe the trick is in that tiny `*` in `get_dispatch_key()` which dereferences the handle (VkInstance) to get the pointer to the underlying dispatch table, I suppose. So, I imagine `DestroyInstance` clears out the memory pointed to by the VkInstance handle, hence `get_dispatch_key()` returns `0`, then there's no table in our map with that key, therefore we also leak them.

charles-lunarg (approved):

PR #2243, opened 2025-01-30
api_dump: Fix deadlock with blocking vkWaitSemaphoresKHR call

ashleysmithcol:
Fix deadlock with blocking vkWaitSemaphoresKHR call where api_dump is locking the outputMutex before the Wait call which results in a deadlock when another thread wants to signal the semaphore but can't because the outputMutex is locked.

vkWaitForFences is marked as blocking so the KHR version should be as-well.

Issue was found when running the CTS `dEQP-VK.synchronization.timeline_semaphore.device_host.write_copy_image_read_image_compute_indirect.image_128x128_r32g32b32a32_sfloat`


charles-lunarg (approved):
Apologies for taking so long to review & manage. Thank you for fixing this!

The change is fine, I've pushed it through internal CI to see if there are any issues there. (I doubt it)

Will merge when that passes.

PR #2297, opened 2025-03-05
cmake: Update jsoncpp to fix CMake 4.0 error

jpr42:
Compatibility with CMake < 3.5 has been removed from CMake 4.0

jsoncpp before version 1.9.6 will cause an error as a result

Link to release notes:
https://github.com/open-source-parsers/jsoncpp/releases/tag/1.9.6

### DIFF
--- a/via/CMakeLists.txt
+++ b/via/CMakeLists.txt
@@ -66,9 +66,6 @@ elseif (MSVC)
     target_compile_definitions(vkvia PRIVATE _CRT_SECURE_NO_WARNINGS _USE_MATH_DEFINES)
 endif()
 
-# NOTE: jsoncpp hasn't had updates in a bit and isn't getting fixes.
-# may be worth using nlohmann/json in the future which gets more updates
-# and has better CMake support.
 find_package(jsoncpp CONFIG)
 if (TARGET jsoncpp_static)
     target_link_libraries(vkvia PRIVATE jsoncpp_static)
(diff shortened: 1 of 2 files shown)


jpr42:
See release notes:
https://cmake.org/cmake/help/latest/release/4.0.html

via/CMakeLists.txt
```diff
@@ -66,31 +66,14 @@ elseif (MSVC)
...
-
-    # https://github.com/LunarG/VulkanTools/issues/1908
-    find_package(PkgConfig)
-    if (PKG_CONFIG_FOUND)
-        pkg_check_modules(jsoncpp REQUIRED IMPORTED_TARGET jsoncpp)
-        target_link_libraries(vkvia PRIVATE PkgConfig::jsoncpp)
-    endif()
-
-endif()
+find_package(jsoncpp 1.9.6 CONFIG)
```
jpr42, line 69: Even if the build uses meson it will still package a CMake configuration file. So now the CMake code can be cleaned up 💯

jpr42: While this code could potentially be cleaned up. I have reverted it since package managers could still use an older version of jsoncpp and this could cause an issue.

I'm rebased the PR to the bare minimum change needed to fix CMake 4.0 support

christophe-lunarg (approved):

christophe-lunarg:
Thanks for contributing @jpr42!

PR #2386, opened 2025-05-15
apidump: Pretty print apiVersions

charles-lunarg:


Eearslya:
I believe `VkLayerProperties.specVersion` also falls under this umbrella.

> specVersion is the Vulkan version the layer was written to, encoded as described in https://registry.khronos.org/vulkan/specs/latest/html/vkspec.html#extendingvulkan-coreversions-versionnumbers.

charles-lunarg:
@Eearslya You aren't wrong, but API Dump doesn't intercept vkEnumerateInstanceLayerProperties or vkEnumerateDeviceLayerProperties (well it does, but it doesn't write to the output. So there isn't anything to do.

PR #2416, opened 2025-06-02
Fix vkCmdDispatchTileQCOM API to match latest vulkan headers

HuYuxin:
The vkCmdDispatchTileQCOM API changed after this vulkan header updates:

https://github.com/KhronosGroup/Vulkan-Headers/commit/b11eecd68fb4b770f30fe2c9da522ff966f95b1e.

Change the API to match the latest.

### DIFF
--- a/layersvt/generated/api_dump.cpp
+++ b/layersvt/generated/api_dump.cpp
@@ -16811,7 +16811,7 @@ VKAPI_ATTR void VKAPI_CALL vkCmdCudaLaunchKernelNV(VkCommandBuffer commandBuffer
     }
 }
 #endif // VK_ENABLE_BETA_EXTENSIONS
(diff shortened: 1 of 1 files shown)


charles-lunarg (approved):
A full header update is whats needed for api dump printing, but this is good for the interim to unblock compilation & usage.

HuYuxin:
> Oh, right, all of the CI jobs fail because they are using the older definition of the function.

Can we update the CI to use the latest function definition in vulkan header? ANGLE cannot integrate the latest VulkanTools after it integrates the latest vulkan-headers: https://chromium-review.googlesource.com/c/angle/angle/+/6610818.

HuYuxin:
Thank you very much for the prompt help!

PR #2418, opened 2025-06-03
vkconfig: Add theme mode control in preference tab

christophe-lunarg:
image


artem-lunarg (approved):
Thanks for adding this.

christophe-lunarg:
> Thanks for adding this.

I found a way to make it works on Windows with Qt 6.8! \o/

PR #2430, opened 2025-06-18
vkvia: Maybe link to JsonCpp::JsonCpp

dg0yt:
In vcpkg, there is only one build type installed. JsonCpp::JsonCpp is always available from an installed jsoncpp.


christophe-lunarg:
Thanks for contributing!

christophe-lunarg:
> This appears to have been merged into the vulkan-sdk-1.4.313 branch, when it should have been merged into main. Typo?

Not really, because vkvia was removed on main.

Issue #1967, opened 2024-03-18
apidump: Print VkDeviceAddress in hex.


dorian-apanel-intel:
Currently VkDeviceAddress is printed in decimal format:
```
Thread 0, Frame 0:
vkGetBufferDeviceAddressKHR(device, pInfo) returns VkDeviceAddress 18446664944411934720:
    device:                         VkDevice = 00000215359B2600
    pInfo:                          const VkBufferDeviceAddressInfo* = 00000061570FC518:
        sType:                          VkStructureType = VK_STRUCTURE_TYPE_BUFFER_DEVICE_ADDRESS_INFO (1000244001)
        pNext:                          const void* = NULL
        buffer:                         VkBuffer = 00000216EC3AA820
```
I think it would be more useful to have it in hex, just like object handles.

Same issue I see with `VkDebugUtilsObjectNameInfoEXT.objectHandle`

PR #2448, opened 2025-06-30
dump: Use hex for VkDeviceAddress

spencer-lunarg:
for https://github.com/LunarG/VulkanTools/issues/1967 (likely others, so maybe not close it completely)

This was 

1. Something I wanted before
2. A quick test how easy @charles-lunarg refactoring made this change.... it was really easy to add, so good job Charles! :confetti_ball:  

-----
Before

Screenshot from 2025-06-30 12-15-35

After

Screenshot from 2025-06-30 12-15-57

### DIFF
--- a/scripts/generators/api_dump_generator.py
+++ b/scripts/generators/api_dump_generator.py
@@ -858,7 +858,6 @@ class ApiDumpGenerator(BaseGenerator):
         else:
             indent = 'indents + (Format == ApiDumpFormat::Json ? 2 : 1)'
 
-
         value = f'{object_access}{var.name}'
         if self.get_is_array(var):
             element_type = f'dump_type<Format, {custom_type}>'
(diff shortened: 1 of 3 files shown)


scripts/generators/api_dump_generator.py
```diff
@@ -448,7 +448,8 @@ def generate_dispatch_codegen(self):
 
             self.write('if (ApiDumpInstance::current().shouldDumpOutput()) {')
             if command.returnType != 'void':
-                if self.get_unaliased_type(command.returnType) in self.vulkan_defined_types:
+                return_type = self.get_unaliased_type(command.returnType)
+                if return_type in self.vulkan_defined_types or return_type == 'VkDeviceAddress':
```
charles-lunarg, line 452: This only affects return values, do you want VkDeviceAddress to be printed as hex in only that case? Or are there just not any places where it is currently used in a struct?

spencer-lunarg: aww good call, ya, we use them in `VK_EXT_device_generated_commands` (and many other spots)

will take a version 2 stab at it later today/tomorrow... regardless, was happy it was easy to do

spencer-lunarg: updated everywhere

before

image

after

image

charles-lunarg (approved):

PR #2449, opened 2025-07-01
api_dump: Print values as pointers, not their address

charles-lunarg:
This fixes a mistake which used the address of a value instead of the value when treating it as a pointer in the log output. Fixing this requires a bit more care with casting, as we are going from uint64_t to void*.


spencer-lunarg (approved):
image

PR #2452, opened 2025-07-01
scripts: Update known_good to 1.4.320

christophe-lunarg:


christophe-lunarg:
> Were there any new extensions in the headers? If so, API Dump can have the codegen run - although if there are issues feel free to leave it out and I can try and figure out what is wrong (since I did a big refactor after all, could have broken something).

I'll let you do a follow up PR, we are trying to resolve the SDK build

christophe-lunarg:
> > Were there any new extensions in the headers? If so, API Dump can have the codegen run - although if there are issues feel free to leave it out and I can try and figure out what is wrong (since I did a big refactor after all, could have broken something).
> 
> I'll let you do a follow up PR, we are trying to resolve the SDK build

@charles-lunarg, I changed my mind! If you could do the header update, it would be nice as the code generation is required and it's late for me and it looks like it should be done on Linux...

charles-lunarg (approved):

PR #2466, opened 2025-07-18
api_dump: Fix regression in double pointer array dumping

charles-lunarg:
Problem was dereferencing a possibly null pointer. Solution is to have a dedicated dump function which handles the array pointer being null and dereferencing the array before printing members.

### DIFF
--- a/layersvt/api_dump.h
+++ b/layersvt/api_dump.h
@@ -1506,6 +1506,31 @@ void dump_pointer_array(const T *array, size_t len, const ApiDumpSettings &setti
     dump_array_end<Format>(array, len, settings, indents);
 }
 
+template <ApiDumpFormat Format, typename T, typename DumpElement>
+void dump_double_pointer_array(const T *const *array, size_t len, const ApiDumpSettings &settings, const char *type_string,
+                               const char *name, const char *element_type, int indents, DumpElement dump_element) {
+    if (array == nullptr || len == 0) {
+        dump_nullptr<Format>(settings, type_string, name, indents);
+        return;
+    }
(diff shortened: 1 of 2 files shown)


layersvt/api_dump.h
```diff
@@ -1506,6 +1506,31 @@ void dump_pointer_array(const T *array, size_t len, const ApiDumpSettings &setti
     dump_array_end<Format>(array, len, settings, indents);
 }
 
+template <ApiDumpFormat Format, typename T, typename DumpElement>
+void dump_double_pointer_array(const T *const *array, size_t len, const ApiDumpSettings &settings, const char *type_string,
+                               const char *name, const char *element_type, int indents, DumpElement dump_element) {
+    if (array == nullptr || *array == nullptr || len == 0) {
```
artem-lunarg, line 1512: Lets assume we passed the first check (array != nullptr). Then we check the first pointer in the array is not null, but I guess other pointers in this array can be null. Not sure what should be validated here, but just want to point this in case it's an issue.

charles-lunarg: Okay this points to an incorrect assumption I made. 
This function prints the contents of a pointer to an array of pointers. Checking whether the first array element is not null isn't sufficient, since we call `dump_element()` on other elements without checking if they are NULL. So to be robust, this should instead call `dump_pointer()` on each array element instead, to perform the null check there.

artem-lunarg (approved):
This fixes the issue I had. One question.

artem-lunarg:
No issues with this update too.

PR #2481, opened 2025-08-16
build: Update to header 1.4.325

ziga-lunarg:


scripts/generators/api_dump_generator.py
```diff
@@ -610,6 +610,8 @@ def generate_implementation(self, video=False):
 
             self.write('')
             for member in struct.members:
+                if (struct.name == 'StdVideoH265HrdParameters' and (member.name == 'pSubLayerHrdParametersNal' or member.name == 'pSubLayerHrdParametersVcl')):
```
spencer-lunarg, line 613: this seems very hacky, we have any idea why this happened?

ziga-lunarg: Yes, the length of these arrays comes from another struct, I opened an issue here #2482

PR #2486, opened 2025-08-21
cmake: Fix layer solution folders having the same name as layer targets

christophe-lunarg:


ziga-lunarg (approved):
Fixes my issue, thanks!

PR #2489, opened 2025-08-25
Fix skipped std video h265 hrd parameters

charles-lunarg:
Adding the necessary special cases turned out to be a bit more work than I originally thought. This was because there was a bug in the get_validity_check generator logic, as well as the complex validity requirements that the previously skipped struct members had.


ziga-lunarg (approved):
LGTM

PR #2495, opened 2025-09-03
scripts: Remove unused custom changes to vulkan_object.py

charles-lunarg:
basetypes and platform types are no longer used by API Dump, so they can safely be removed from the local copy of vulkan_object.py & base_generator.py.


spencer-lunarg:
@charles-lunarg so they are not needed because which?

1. you got the information somehow else for API Dump
2. API Dump just was redone that the information is not needed

PR #2507, opened 2025-09-27
Use deviceName + driverName + driverVersion to distinguish drivers

charles-lunarg:
There may be multiple drivers that are for the same device, so add the driver name and driver version to the displayed name.

Add the driverUUID, driverVersion, and driverName fields to the loader settings json file, so that the loader picks out the correct VkPhysicalDevice.


christophe-lunarg:
This change break the additional drivers, investigating.

PR #2656, opened 2026-04-28
vkconfig: Fix highligher and app log streaming perf issue

christophe-lunarg:
And also add a control:
image


ziga-lunarg (approved):
I will retest with the next sdk build

PR #2696, opened 2026-07-08
vkconfig: optimize compiling of tests to fix link crashes

lunarpapillo:
We've seen situations where Linux and Mac test builds crash while linking tests.  The large number of Qt-based test executables (43) all trying to link at the same time can provoke out-of-memory conditions on Linux, triggering the OOM killer and build errors like:

    collect2: fatal error: ld terminated with signal 9 [Killed]

On Mac, the simultaneous Qt loads can provoke a linker crash with symptoms like:

    c++: c++: error: unable to execute command: Bus error: 10
    c++c++: error: linker command failed due to signal (use -v to see invocation)

Also, there appears to be a multiple load of GoogleTest, because the existing code references both GTest::gtest and GTest::gtest_main; provoking messages like:

    ld: warning: ignoring duplicate libraries: '/Users/lunarg/.jenkins/vt0/Debug64/VulkanTools/external/googletest/build/install/lib/libgtest.a'

Removing the reference to GTest::gtest removes the redundant reference.

### DIFF
--- a/vkconfig_core/test/CMakeLists.txt
+++ b/vkconfig_core/test/CMakeLists.txt
@@ -32,12 +32,11 @@ function(vkConfigTest NAME)
     file(GLOB FILES_TEXT ./*.txt)
     source_group("Test Files" FILES ${FILES_JSON} ${FILES_TEXT} ${FILES_LAYER_JSON})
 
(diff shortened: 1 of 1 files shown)


lunarpapillo:
There's something wrong with dependency management here.  Closing until I can figure it out.

PR #2715, opened 2026-08-19
vkconfig: Remove Windows ARM release notification workaround

### LINKED ISSUE #2679: vkconfig: Windows ARM SDK download is disabled
Due to a Qt 6.8.3 bug, we disabled SDK release detection and download was disabled. 

We are currently using Qt 6.9.3, so maybe we could check again

herlience:
Due to Qt update (>=6.9.3), the workaround for the Windows ARM release notification crash is no longer needed.

Fixes #2679

### DIFF
--- a/vkconfig_gui/tab_preferences.cpp
+++ b/vkconfig_gui/tab_preferences.cpp
@@ -95,23 +95,14 @@ TabPreferences::TabPreferences(MainWindow &window, std::shared_ptr<Ui::MainWindo
     this->ui->preferences_download->setText("Searching Latest Vulkan SDK...");
     this->ui->preferences_validate_layer->setChecked(configurator.layers.validate_manifests);
 
-#if WORKAROUND_WINARM_RELEASE_NOTIFICATION_BUG
-    // Windows ARM crash, it looks like a Qt bug in 6.8.2...
-    if (VKC_PLATFORM == PLATFORM_WINDOWS_ARM) {
-        this->ui->preferences_group_box_releases->setVisible(false);
-    } else {
-#endif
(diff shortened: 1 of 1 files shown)


christophe-lunarg:
We haven't verified that a newer version of Qt fix the issue on ARM. Did you ?

herlience:
I don't have a physical Windows ARM device to test on natively. What I did was build and run the code with Qt 6.11.2 on Windows x64, while forcing `VKC_PLATFORM == PLATFORM_WINDOWS_ARM` during testing to trigger the ARM code path and the `GetLatestReleaseSDK` network request. Under this setup, the network request completed successfully and fetched the latest release without any crashes.

christophe-lunarg:
I don't think that shows the issue is fixed on Windows ARM. The release notification had no problem on Windows x64.

herlience:
I see, you are right about the native ARM layer. Could you give it a try when you have the chance?

christophe-lunarg:
> I see, you are right about the native ARM layer. Could you give it a try when you have the chance?

Sadly, I don't have a Windows ARM machine either so I started a discussion with the SDK team to figure out whether this is something we could check during next SDK release cycle...

herlience:
Thanks for context and for following up with the SDK team! Sounds like a solid plan. 👍

richard-lunarg:
I can test this on a Windows ARM machine here, and will verify or update the Qt version (I may have already) fixes this.

herlience:
Awesome! Appreciate the support.

richard-lunarg:
This does appear to be fixed now with the newer Qt builds. Good to go.

PR #2721, opened 2026-09-02
vkconfig: Add layer manifest validation multithreading (#2659)

### LINKED ISSUE #2659: vkconfig: Add layer manifest validation multithreading

herlience:
Fixes #2659 

Splits manifest loading into a parallel parse phase (file read + JSON parse) and a sequential apply phase (validation + inserting into available_layers), since the latter touches shared state.

Added:
- ParseLayerFile(): reads a manifest file and parses its JSON. No shared state, so safe to run on multiple threads at once.
- ApplyParsedLayer(): takes already-parsed JSON, validates it, and inserts/updates it in available_layers. Kept sequential because it mutates shared data and depends on ordering.

Removed:
- LoadLayers(): did file read + parse + validate + insert all in one function, entirely sequentially. Replaced by the two functions above so the expensive I/O/parse part can run in parallel.

### DIFF
--- a/vkconfig_core/layer_manager.cpp
+++ b/vkconfig_core/layer_manager.cpp
@@ -28,6 +28,10 @@
 #include <QDesktopServices>
 #include <QFileDialog>
 
+#include <future>
+#include <thread>
+#include <atomic>
+
 static std::vector<Path> GetEnvVariablePaths(const char *variable_name, LayerType type) {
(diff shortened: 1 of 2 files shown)


herlience:
hmm lemme fix it
i understood why it gives errors

herlience:
Pushed a fix for the CI failures. Should pass now.

